PR #2, opened 2019-10-08
BBmap

johanneskoester:
@SilasK I have migrated your PR to the new Github home of snakemake-wrappers.


dlaehnemann:
Any plans for further work on this, or should this be closed for now?

SilasK:
Close it for now.  But yes I have some plans but little time.

Could you help me with some questions:

- Is there a way to use some python functions in a set of similar wrappers. E.g. there are many bb tools with similar CLI.

dlaehnemann:
Hi @SilasK . I am not sure about the complexities of the command line interface, but usually we try to handle as many parameters as possible through some generic argument like `params: args="--example-arg 5"` which are then included in the shell command in the wrapper via `{snakemake.params.args}`. Only args that don't fit into this scheme would then be handled separately. Does this reduce the complexity of the wrapper(s), or am I missing some greater CLI complexity here?

dlaehnemann:
Or maybe the complex tool can be broken down into simple use cases with defined input, outputs and a few arguments set in a specific way. Then, a simple wrapper could be created for each such use case.

I am not familiar with bbmap, so this might not be workable. Just throwing around ideas, here... ;)

SilasK:
My question is only if there is a simple way to use the same code (parser functions) for different wrappers or need I copy it.

dlaehnemann:
No, I'm afraid the general idea is to keep wrappers separate from each other. Hence my suggestions of working around the need for a big parsing function.

dlaehnemann:
@SilasK There is now `snakemake-wrapper-utils` for functionality that is to be shared between wrappers. Have a look here:
https://github.com/snakemake/snakemake-wrapper-utils/tree/master/snakemake_wrapper_utils

Maybe this can be helpful in reviving this effort. Also, there is another BBtool with a wrapper, that might have good ideas:
https://github.com/snakemake/snakemake-wrappers/tree/master/bio/bbtools/bbduk

johanneskoester:
Closing for now as requested by @SilasK.

PR #4, opened 2019-10-08
Bismark tools wrapper

iromeo:
Smakemake wrappers for Bismar tools for BS-Seq processing


iromeo:
> test.py::test_cairosvg FAILED   

1 test fails, and it isn't related to this PR

iromeo:
@johanneskoester Any chance to include it in next wrappers release? Several versions were released since this pull request.

iromeo:
@johanneskoester Tests are green now, SonarCloud code check doesn't understand that `snakemake` module is already available in `wrappers.py` and doesn't require an explicit import. Could you please review this code request and accept if everything is ok?

iromeo:
@johanneskoester thank you for your feedback. Wrappers already support logs, log section just isn't mentioned in `Snakefile` in tests. I'm going to fix tests, do you expect logs to be placed in a separate `logs` folder (as in some existing wrappers) or it is ok to put them in the same directories as output files? I prefer putting logs near output files in my pipelines, but for consistency with other examples seems better is to use separate `logs` folder. Right?

iromeo:
> May I ask you to fix the formatting of the example snakefiles? There shall be no whitespace around the = in input, output and params lists. Also please add log file support (see other wrappers).

Done, please review

PR #7, opened 2019-10-09
Wrapper for primerclip.

Smeds:


bio/primerclip/test/Snakefile
```diff
@@ -0,0 +1,12 @@
+rule primerclip:
+    input:
+        master_file="master_primer_file",
+        sam="mapped/{sample}.sam"
```
johanneskoester: Does it work with bam or cram as well? If not, the wrapper should probably automatically add samtools piping. Nobody should be forced to use sam if not needed ;-).

Smeds: Good suggestion! I'm able to add samtool conversion for the input. But I get a truncation error when I try to do it for the output, primerclip outputs a file with sam format.

johanneskoester: But can't you pipe the output into `samtools view -Sbh`?

Smeds: Manage to find out why I couldn't pipe the output successfully. Primerclip outputs some run information to stdout which mess up the alignment file. I solved it by removing the 3 last lines from the primerclip output, which contained that information.

Smeds: Fixed the formatting issue a few minutes ago.

Smeds: Updated Snakemake file

Smeds:
Don't really understand why SonarCloud fail=/

PR #18, opened 2019-11-06
add wrapper for PLASS protein assembler

bluegenes:
repo: https://github.com/soedinglab/plass

paper: [Steinegger M, Mirdita M and Soeding J. Protein-level assembly increases protein sequence recovery from metagenomic samples manyfold. Nature Methods, doi: doi.org/10.1038/s41592-019-0437-4 (2019).](https://www.nature.com/articles/s41592-019-0437-4)


bio/plass/test/Snakefile
```diff
@@ -0,0 +1,14 @@
+
+rule trinity:
```
johanneskoester: I guess the rulename is wrong?

bio/plass/wrapper.py
```diff
@@ -0,0 +1,51 @@
...
+__license__ = "MIT"
+
+from os import path
+from snakemake.shell import shell
+
+extra = snakemake.params.get("extra", "")
+
+# allow multiple input files for single assembly
+left = snakemake.input.get("left")
+single = snakemake.input.get("single")
```
johanneskoester: The single case is not documented in the example. Please have a second rule that illustrates and tests it (you can call it explicitly in the test case function in test.py).

PR #21, opened 2019-11-07
Add hmmer

bluegenes:
Add wrappers for hmmbuild, hmmpress, hmmscan


bluegenes:
code analysis failing bc hmmscan and hmmpress are very similar wrappers

PR #23, opened 2019-11-08
Add last aligner

bluegenes:
LAST aligner

software: http://last.cbrc.jp/

publication list: http://last.cbrc.jp/doc/last-papers.html


bluegenes:
ready for review

PR #24, opened 2019-11-08
Add transdecoder wrappers

bluegenes:
TransDecoder identifies candidate coding regions within transcript sequences, such as those generated by de novo RNA-Seq transcript assembly using Trinity, or constructed based on RNA-Seq alignments to the genome using Tophat and Cufflinks.

https://github.com/TransDecoder/TransDecoder


bluegenes:
ready for review

PR #29, opened 2019-11-17
diamond blast

bluegenes:


dlaehnemann:
@bluegenes, just checking in to see if you plan to further work on this PR? Anything where another pair of eyes might help?

dlaehnemann:
@bluegenes Would you like a review on this Pull Request? Or is it obsolete and we can close it?

bluegenes:
@dlaehnemann The basic diamond blast was working well, but I think the taxonlist functionality did not pass tests -- something about `ftputil` if I recall. 

I can take a look again later this week. A review would be helpful if you have time!

bio/diamond/blast/test/Snakefile
```diff
@@ -0,0 +1,46 @@
...
+        extra=""
+    threads: 4
+    wrapper:
+        "master/bio/diamond/blast"
+
+rule diamond_blastp_taxonlist:
+    input: 
+        pep_query="pep.fa",
+        db="pep_with_taxinfo.dmnd", # see diamond makedb wrapper for how to generate this file
+        #Comma-separated list of NCBI taxonomic IDs to filter the database by. Any taxonomic rank can be used, and only reference sequences matching one of the specified taxon ids will be searched against. Using this option requires setting the --taxonmap and --taxonnodes parameters for makedb.
```
dlaehnemann, line 35: I'm not sure what this comment refers to? Is it the `db=` specification above? Or the `output` right below? Or even the `taxonlist="3760"` even further below?

Also, this is a rather long comment to be part of the example rule for a wrapper. People will usually copy-paste this and this pastes a lot of extra stuff.

Maybe this could be reduced to one sentence plus some link -- either to the wrapper's docs (for the expected link syntax, simply check out the doc links of existing wrappers), or to the tool's docs. In the first case, more detailed infos could be provided in the wrapper description or the input / output specification.

And by convention I would put a comment above or at the end of a line that it refers to.

bio/diamond/blast/wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+extra = snakemake.params.get("extra", "")
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+pep_query = snakemake.input.get("pep_query")
+nucl_query = snakemake.input.get("nucl_query")
+# assert pep_query is not None or nucl_query is not None, "either pep_query or nucl_query are required as input"
+queryfiles = [pep_query, nucl_query]
+assert (
+    list(map(bool, queryfiles)).count(True) == 1
+), "one query is required: either a peptide query (pep_query) for diamond blastp OR a nucleotide query (nucl_query) for diamond blastx"
```
dlaehnemann, line 18: Maybe this is a bit shorter (and hopefully works, I haven't tested):
```suggestion
assert list(map(bool, [pep_query, nucl_query])).count(True) == 1,
    "one query is required: either a peptide query (pep_query) for diamond blastp OR a nucleotide query (nucl_query) for diamond blastx"
```

bio/diamond/blast/wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+        # this mode is a lot more sensitive that the default and generally recommended for aligning longer sequences.
+        s_cmd = " --sensitive "
+    elif sensitivity == "more-sensitive":
+        # more-sensitive: This mode provides some additional sensitivity compared to the sensitive mode.
+        s_cmd = " --more-sensitive "
+    else:
+        raise ValueError(
+            "alignment sensitivity must be either 'sensitive' or 'more-sensitive'"
+        )
+
```
dlaehnemann, line 61: Wouldn't it be easier to leave this kind of configuration up to the users? They can always specify any of these parameters via something like `extra="--taxonlist=3760"` and this way the wrapper does not need to change, even if command line arguments or their behaviour changes. It would keep maintenance down to a minimum, here. Or is there something I am overlooking, that requires an explicit handling of such options?

bio/diamond/makedb/test/Snakefile
```diff
@@ -0,0 +1,41 @@
...
+        """
+        mv {input} {params.zipfile}
+        unzip {params.zipfile}
+        """
+
+rule diamond_makedb_taxinfo:
+    input: 
+        db="pep.fa",
+        taxonmap="prot.accession2taxid.gz",
+        taxonnodes="nodes.dmp",
```
dlaehnemann, line 35: I like, that the two different versions of this rule already demonstrate, that these two inputs are optional. But just to make it as obvious as possible, maybe add an `# optional input` to each of them (and skip the trailing comma, it will probably cause a syntax error):

```suggestion
        taxonmap="prot.accession2taxid.gz", #optional input
        taxonnodes="nodes.dmp" #optional input
```

bio/diamond/makedb/test/Snakefile
```diff
@@ -0,0 +1,41 @@
...
+rule ftp_get_taxonnodes:
+    input: FTP.remote(f"ftp://ftp.ncbi.nlm.nih.gov/pub/taxonomy/taxdmp.zip", static=True, keep_local=True, immediate_close=True)
+    output: "nodes.dmp"
+    params:
+        zipfile= "taxdmp.zip"
+    shell:
+        """
+        mv {input} {params.zipfile}
+        unzip {params.zipfile}
+        """
```
dlaehnemann, line 29: I think this is functionality that a user should implement in their pipelines separately, i.e. which doesn't need to be part of the example Snakefile here. The optimal solution for the example rule below (`diamond_makedb_taxinfo`) would in my opinion be to find a `taxonmap` and a `taxonnodes` file that are really small dummy data but nevertheless work, and include them as testing data in the `/test/` subfolder. Could you for example subset the files you are downloading here to only use a couple of rows each and then gzip the result again for `taxonmap`? Or would `diamond makedb` complain about this?

dlaehnemann:
I guess this PR would take a bit more time than the other two I just ping'ed on, @bluegenes . Also, some diamond subcommands have been added in the meantime through PR #279.

But I think this PR still has stuff to add on top of that. Or should we rather close this one and move this to a new PR whenever you can find some time?

PR #31, opened 2019-11-21
Wrapper for tximport

tdayris-perso:
Hi,

I added a tximport wrapper. All arguments can be passed through snakemake params. I copied the salmon quant output in order to test this wrapper.

Readr and jsonlite are optional requirement for tximport, however: 

- No Kallisto/Salmon boostrap importation is possible without jsonlite
- readr package makes the process a lot faster

Thanks in advance for your review.


bio/tximport/test/Snakefile
```diff
@@ -0,0 +1,12 @@
...
+        quant = expand("quant/A/quant.sf")
+    output:
+        txi = "txi.RDS"
+    message:
+        "Testing tximport"
+    params:
+        type = "salmon",
+        txOut = True
+    wrapper:
+        "file:.."
```
johanneskoester: Please use "master/bio/tximport" as wrapper url. It is replaced with the right version tag when building the docs, and allows people to directly copy paste from this test case into their workflow.

bio/tximport/test/Snakefile
```diff
@@ -0,0 +1,10 @@
+rule tximport:
+    input:
+        quant = expand("quant/A/quant.sf")
```
johanneskoester: Please mention the optional tx2gene (e.g. commented out).

bio/tximport/test/Snakefile
```diff
@@ -0,0 +1,10 @@
...
+    output:
+        txi = "txi.RDS"
+    params:
+        type = "salmon",
+        txOut = True
```
johanneskoester: Are the matrixes generated by txOut included into the RDS?

tdayris-perso: In this precise case, given that: (1) no bootstraping steps are performed within the salmon quant wrapper tests and (2) no transcript to gene aggregation was possible on the two contigs used to test the salmon index, then there is not much to include in the RDS object besides abundaces, tpm, length and expected length. These fields are available with classic `txi$` syntax.

Heavier transcriptome file would be needed to test transcripts counts aggregation into genes counts (2), and larger fastq files would be needed to test inferential replicates importation (1).

However, these matrices would be included in the txi object returned by tximport if the quantification process would satisfy both (1) and (2). The function `saveRDS` applied on the root of the txi object saves all of its member variables and member functions. If you know any exception to this behaviour, please tell me. (I use RDS files to display results through local shiny appliances, if I'm doing things wrong... I'd prefer to know.)

bio/tximport/test/Snakefile
tdayris-perso: I shall be done in next update. Thanks for your review!

bio/tximport/test/Snakefile
tdayris-perso: I shall be added in next update. Thanks again!

bio/tximport/test/Snakefile
johanneskoester: That sounds reasonable. Can you give a more speaking name, also in snake_case (in order to adhere with Python and Snakemake style), to txOut?

tdayris-perso: `txOut`, `type`, and other possible arguments are given by tximport itself, and passed like keyword-named arguments into the `tximport::tximport()` function.

I can wrap all arguments into a python dictionnay named "`extra`", which will contain the tximport arguments and follow classic wrappers behavior.

Would that suit you?

johanneskoester: `type` is fine. I just want `txOut` to be self-explanatory named. For example `transcript_only`?

tdayris-perso: `txOut` is a tximport argument, its name could have been better, I agree. However, this choice belongs to the tximport development team.

I changed the wrapper, so now `tximport` arguments are passed through a string just like any other wrapper.

Many thanks for your reviews!

johanneskoester: This is indeed a good idea, since it also allows other args to be passed.

PR #33, opened 2019-11-25
add wrapper for downloading dbsnp

johanneskoester:


johanneskoester:
Not needed anymore, we have the possibility via ensembl now.

PR #34, opened 2019-11-25
wrappers for downloading sequence, variation and annotation from ensembl servers.

johanneskoester:


mobilegenome:
This came just in time, as I was working on something similar. I noticed a small issue: The test-Snakefile indicate that plain text-files are written (.fasta, .gtf, .gff3) however, the wrapper writes gzipped files. As most operation on the files require the unzipped files, it might be useful to add a unzipping routine in the wrapper.

PR #36, opened 2019-12-04
Add checksum verification for ENSEMBL sequence and annotation wrappers

mobilegenome:
- I've added a routine to compare checksums for downloaded files.  
- Output files in the test snakefiles are now ending with ".gz" to reflect the downloaded fileformat correctly. 

Ideally, the checksum() method I've added would be stored separately to avoid redundancy, but I believe this would break the ability of each wrapper to function stand-alone. If there is a way to import common libraries, please let me know.

### DIFF
--- a/bio/reference/ensembl-sequence/wrapper.py
+++ b/bio/reference/ensembl-sequence/wrapper.py
@@ -4,6 +4,9 @@ __email__ = "johannes.koester@uni-due.de"
 __license__ = "MIT"
 
 from ftplib import FTP
+from io import StringIO
+from subprocess import run
+from os.path import basename
 
 species = snakemake.params.species.lower()
 release = snakemake.params.release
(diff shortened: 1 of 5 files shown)


mobilegenome:
Before going further with this, I would like to discuss an issue I have discovered with the current version: Python's `ftplib` does not work behind a corporate proxy-server and I didn't find a good workaround for this. Now, I have a working version of these sets of wrappers that call `curl` via `subprocess.run`. If interested in this change, I am happy to renew this pull-request.

bio/reference/ensembl-sequence/wrapper.py
```diff
@@ -4,13 +4,37 @@
...
 species = snakemake.params.species.lower()
 release = snakemake.params.release
 build = snakemake.params.build
 
 suffixes = ""
 datatype = snakemake.params.get("datatype", "")
+
+
+def checksum():
+    lines = r.getvalue().strip().split("\n")
```
johanneskoester: Does this mean that the entire file is loaded into memory?

johanneskoester: As one piece?

mobilegenome: Yes, but this is only the checksum text file, a few KB maximum. Should not be a problem...

johanneskoester: Oh, right!

johanneskoester (changes requested):

PR #40, opened 2019-12-13
fastp wrapper updated: make 'trimmed' path optional

iromeo:
`fastp` tool trims FASTQ files optionally, e.g. you could use it just for QC as an alternative to FastQC, not only for trimming. Current `fastp` wrapper implementation force user to perform trimming in any case otherwise it fails with an error:
```
  File "/workdir_pilot_small/.snakemake/scripts/tmpj0s1jjw7.wrapper.py", line 25, in <module>
    trimmed = "--out1 {} --out2 {}".format(*snakemake.output.trimmed)
AttributeError: 'OutputFiles' object has no attribute 'trimmed'
```
This fix resolves the issue and allows using the wrapper both for trimming and just for QC check.


iromeo:
Thx, now I can stop using my fork as custom wrappers repository and switch back to the default one!

PR #44, opened 2019-12-18
Arriba wrapper

jafors:
Wrapper for the fusion detection tool arriba.


bio/arriba/test/Snakefile
```diff
@@ -0,0 +1,27 @@
...
+        fusions="fusions/{sample}.tsv",
+        # discarded gene fusions
+        discarded="fusions/{sample}.discarded.tsv" # optional
+    log:
+        "logs/arriba/{sample}.log"
+    params:
+        # path to reference genome
+        genome="genome.fasta",
+        # path to annotation gtf
+        annotation="annotation.gtf",
```
johanneskoester: Above two can now be input files. I plan to modify all wrappers accordingly soon.

test.py
```diff
@@ -106,6 +113,33 @@ def test_bedtools_slop():
...
+    )
+
+
+def test_bedtools_merge():
+    run("bio/bedtools/merge", ["snakemake", "A.merged.bed", "--use-conda", "-F"])
+
+
+def test_bedtools_slop():
+    run("bio/bedtools/slop", ["snakemake", "A.slop.bed", "--use-conda", "-F"])
+
```
johanneskoester: Where do all these tests come from? Have they been missing before?

bio/arriba/environment.yaml
```diff
@@ -0,0 +1,7 @@
+channels:
+  - conda-forge
+  - bioconda
+  - defaults
+dependencies:
+  - arriba ==1.1.0
+  - star ==2.7.1a
```
johanneskoester: Why does it need star as a dependency? Star is not called in the wrapper.

jafors: Correct, I probably had it in there for testing it together with star.

bio/arriba/test/Snakefile
jafors: True, makes more sense to have those as inputs.

test.py
jafors: They should have been included in my last PR actually. Maybe I messed up the commits.

PR #48, opened 2020-01-16
Update STAR to latest version.

tdido:


tdido:
Ah, that's great to know. Does this apply in general to bioconda, i.e. will the channel setup instructions be updated eventually?

johanneskoester:
No, in general, it is advantegeous to have a higher priority for conda-forge, because it ensures that you get the latest version for software that has been moved over. Here, however, versions are always specified explicitly. Hence it is better to have bioconda first, because it ensures that always the intended software is used instead of accidentally grabbing something with the same name from conda-forge (although conda-forge checks for name conflicts with bioconda, who knows what might happen in the future).

tdido:
Awesome, thank you for the explanation.

PR #62, opened 2020-03-02
fasterq-dump: Allow setting the temporary directory via params

tedil:
Since downloaded files may be rather large and the default temporary directory location may not be large enough to contain them, it may sometimes be useful to be able to set a custom tempdir.

### DIFF
--- a/bio/sra-tools/fasterq-dump/wrapper.py
+++ b/bio/sra-tools/fasterq-dump/wrapper.py
@@ -20,4 +20,3 @@ with tempfile.TemporaryDirectory(dir=tmpdir) as tmp:
     shell(
         "fasterq-dump --temp {tmp} {extra} {outdir} {snakemake.wildcards.accession} {log}"
     )
-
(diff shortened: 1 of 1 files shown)


johanneskoester:
Why not just rely on $TEMPDIR? This way, it is configurable outside of the workflow itself.

johanneskoester:
If TEMPDIR is too small on a system, it is the admin job to increase it accordingly I think.

tedil:
Ah, right, didn't think of that! However, the changes I proposed still default to the system $TEMPDIR and just add some more control over the tempdir by allowing to set it explicitly.

tedil:
... though that might not be idiomatic snakemake-wrappers style?

johanneskoester:
Well, the issue is that it kind of encourages people to add a system-specific setting into a workflow definition (e.g. an absolute path), rendering the entire workflow not portable. What we could think about instead is that snakemake always sets `TEMPDIR` to a folder in the working directory, e.g., `.snakemake/temp`. But this is dangerous as well, because there will be setups where admins want temp to be on a certain partition...

dlaehnemann:
Any more plans for this PR? Or should this PR simply be closed?

dlaehnemann:
@tedil If you want to pick this back up, feel free to reopen the PR or start a new one.

PR #76, opened 2020-03-18
freebayes: added placeholders for optional input files

daynefiler:
Included now are parameters for `--targets`, `--cnv-map`, & `--populations`.


bio/freebayes/test/Snakefile
```diff
@@ -9,6 +9,9 @@ rule freebayes:
         "logs/freebayes/{sample}.log"
     params:
         extra="",         # optional parameters
+        targets="targets.bed",
+        cnvMap="cnv-map.txt",
+        populations="",
```
johanneskoester: Please add these as optional input files instead. Otherwise, using this wrapper will fail when executing on distributed systems like the cloud.

bio/freebayes/test/Snakefile
```diff
@@ -1,15 +1,19 @@
 rule freebayes:
     input:
-        ref="genome.fasta",
+        ref = "genome.fasta",
```
johanneskoester: Please follow the python style guide: no spaces around keyword args (which essentially is what we have here in input/output/params).

daynefiler:
This can probably be closed -- it was really the regions option I wanted, and #85 added it. I guess users can add the other optional files into the 'extra' param?

johanneskoester:
Ok! Thanks anyway!

PR #77, opened 2020-03-19
Modified snpeff call

FelixMoelder:
This PR fixes an error where snpeff fails with the following input error: `Error:  Cannot read input file '/dev/fd/63'`.
This is resolved by adding a directed-pipe.


bio/snpeff/wrapper.py
```diff
@@ -38,6 +38,6 @@
 
 shell(
     "snpEff {data_dir} {stats_opt} {csvstats_opt} {extra} "
-    "{snakemake.params.reference} {incalls} "
+    "{snakemake.params.reference} < {incalls} "
```
johanneskoester: Incalls can be `<(bcftools view {})`. Hence, this could fail. Instead, the right place to fix this is an else in line 25, turning incalls into `< {}`.

PR #86, opened 2020-04-15
Update freebayes

FelixMoelder:
It shows that freebayes calling currently does not work when passing a bed region file.

The issue here is that `bedtools intersect` requires two tab-separated bed-files.
Currently `fasta_generate_regions.py {snakemake.input.ref}.fai {chunksize}` generates output which looks as following : `chromosome:start-end`
To fix this we can just use `sed` to find-replace `:`and `-` by a tab. (Could these characters appear in a chromosome name?!)

So, after fixing this issue, `bedtools intersect` creates a nice tab-separated bed output causing the next issue.

Freebayes itself needs the single regions in the previously changed `chromosome:start-end` format...not tab-separated.
So, how do we get back there?


johanneskoester:
Sounds like a job for awk.

johanneskoester:
Indeed, I think chromosome names can potentially contain hyphens. So, using a regular expression based approach might be safer. https://opensource.com/article/19/11/how-regular-expressions-awk

dlaehnemann:
Also, for a comprehensive awk manual, check out:
https://www.math.utah.edu/docs/info/gawk_toc.html

And specifically the part on regular expressions is here:
https://www.math.utah.edu/docs/info/gawk_5.html#SEC25

FelixMoelder:
Applying sed also worked when using capture groups.

First we replacing a pattern like `chr1:10-100` into `chr1\t10\100` by applying `sed 's/:\([0-9]*\)-\([0-9]*\)/'$'\t''\1'$'\t''\2/g'`
The idea is to split the term into chromosome, start- and endpoint. The term just searches for a pattern starting with a colon, followed by some digits, a hyphen and more digits. 

After applying bedtools intersect the output must be reverted into the original syntax separated by colon and hyphen.
This is done by applying: `sed 's/'$'\t''\([0-9]*\)'$'\t''\([0-9]*\)/:\1-\2/g'`
Instead of looking for colons and hyphens we now check for tabs.

This way I guess we should only replace the separators as there should be no chromosome-name containing the following pattern `:[0-9]*-[0-9]*`.

So, if I'm not missing something we should be good to go.

bio/freebayes/test/Snakefile_bed
```diff
@@ -2,7 +2,7 @@ rule freebayes:
     input:
         ref="genome.fasta",
         # you can have a list of samples here
-        samples="mapped/{sample}.bam"
+        samples="mapped/{sample}.bam",
         # optional bed-file defining chromosomal regions being mutual exclusive to chunksize defined in params
```
dlaehnemann, line 6: Not sure what "mutual exclusive" means, here? In my understanding, "mutually exclusive" would mean that you can only either specify one or the other.

But if I understand the code in `wrapper.py` correctly, you can actually specify both and they do different things:
* This region file restricts freebayes to call only on those regions.
* The chunksize is supposed to further chop these regions into smaller chunks for further parallelization.

So maybe a comment with the following text would be more accurate (feel free to improve further):
```
# optional BED file specifying chromosomal regions on which freebayes 
# should run, e.g. all regions that show coverage; for further parallelization,
# these regions can be split into smaller chunks by specifying a chunksize:
# params:
#     chunksize = 100000
```

dlaehnemann: Just edited this, realised that I was confusing what `fasta_generate_regions.py` generates and what is the bed input file.

bio/freebayes/wrapper.py
```diff
@@ -20,11 +20,11 @@
     freebayes = "freebayes"
 else:
     chunksize = snakemake.params.get("chunksize", 100000)
-    regions = (
-        "<(fasta_generate_regions.py {snakemake.input.ref}.fai {chunksize})"
-    ).format(snakemake=snakemake, chunksize=chunksize)
+    regions = "<(fasta_generate_regions.py {snakemake.input.ref}.fai {chunksize})".format(
+        snakemake=snakemake, chunksize=chunksize
+    )
```
dlaehnemann, line 25: Is this edit intentional, because it does not change the code but only the formatting? And isn't the original formatting more along the lines of Python style guides, e.g. enforced by `black`?

bio/freebayes/wrapper.py
```diff
@@ -20,11 +20,11 @@
...
     if snakemake.input.get("regions", ""):
-        regions = "<(bedtools intersect -a {regions} -b {snakemake.input.regions})".format(
+        regions = "<(sed 's/'$'\t''\([0-9]*\)'$'\t''\([0-9]*\)/:\1-\2/g' <(bedtools intersect -a =(sed 's/:\([0-9]*\)-\([0-9]*\)/'$'\t''\1'$'\t''\2/g' {regions}) -b {snakemake.input.regions}))".format(
```
dlaehnemann, line 27: Nice regex, a few points that might be improved:

* [ ] I think this might be a bit more readable as a pipe.
* [ ] Also, chromosome names can theoretically contain anything, and I have seen both `:` and `-`, so I would further anchor your regex at the end of the line, to ensure that it doesn't match in some weird chromosome names that do contain the full pattern (and I would also do this anchoring for converting back from BED to chr:start-end, because it's easily enforced, consistent and doesn't hurt).
* [ ] Also, I'm not sure what kind of escape the `'$'\t''` constructs are and think, that this should work with simply `\\t` instead. Generally, all `\` in a shell command that you want to keep in the eventual execution command have to be protected from the Python format minilanguage parsing by escaping them with another `\`. Check out the [respective entry in snakemake FAQ](https://snakemake.readthedocs.io/en/latest/project_info/faq.html#i-get-a-nameerror-with-my-shell-command-are-braces-unsupported). But maybe I am also wrong here, and this should simply work with a single `\` for all of them, maybe @johanneskoester can clarify.

Anyhoot, here's a version with my suggestions:

```
regions = 
    "<( bedtools intersect -a <( sed 's/:\\([0-9]*\\)-\\([0-9]*\\)$/\\t\\1\\t\\2/g' {regions}) -b {snakemake.input.regions} | "
    "    sed 's/\\t\\([0-9]*\\)\\t\\([0-9]*\\)$/:\\1-\\2/g' )"
```

bio/freebayes/test/Snakefile_bed
FelixMoelder: > Not sure what "mutual exclusive" means, here? In my understanding, "mutually exclusive" would mean that you can only either specify one or the other.
> 
> But if I understand the code in `wrapper.py` correctly, you can actually specify both and they do different things:
> 
> * This region file restricts freebayes to call only on those regions.
> * The chunksize is supposed to further chop these regions into smaller chunks for further parallelization.
> 
> So maybe a comment with the following text would be more accurate (feel free to improve further):
> 
> ```
> # optional BED file specifying chromosomal regions on which freebayes 
> # should run, e.g. all regions that show coverage; for further parallelization,
> # these regions can be split into smaller chunks by specifying a chunksize:
> # params:
> #     chunksize = 100000
> ```

Oh, your right. That comment is outdated, as in a previous version you either created a region file by the chunk size or alternatively by passing a bed-file.
Now you need to pass both to perform the intersection.

FelixMoelder: Just changed the comment in the way you suggested

bio/freebayes/wrapper.py
FelixMoelder: Good point anchoring it to the end of line.
The `'$'\t''` gave me a bit of a headache. The issue here is, that sed does not recognize `\t` or even escaping with `\\t` as a tab. This can be archived by prepending `$`.
BUT prepending it to the whole expression doesn't allow to use the capture groups invoked by `\1`and `\2` anymore.

But maybe I found a better way by replacing all this quotes by `$(printf '\t')`. This way we just need to escape the outer-quotes as it is interleaved into a python string.

bio/freebayes/wrapper.py
FelixMoelder: Actually, the code has changed a bit in the last PR.
Previously this was something like:
```
freebayes = (
        "freebayes-parallel <(fasta_generate_regions.py "
        "{snakemake.input.ref}.fai {chunksize}) "
        "{snakemake.threads}"
).format(snakemake=snakemake, chunksize=chunksize)
```
In the current version there is only one single line left that get's formatted. So does it really help to put it into brackets?

FelixMoelder:
There is one annoying issue left.
It seems like python or snakemake is interpreting the regular expression turning `\t` into an actual tab and the capture groups `\1` and `\2` just disappear.

While the command should look like
`"s/:\([0-9]*\)-\([0-9]*\)$/$(printf '\t')\1$(printf '\t')\2/g"` 
but snakemake instead calls
`"s/:\([0-9]*\)-\([0-9]*\)$/$(printf '	')$(printf '	')/g"`.
(Mind the missing `\t`, `\1`and `\2`)

I did not find a way to solve this. Escaping using a backslash does not help, neither using a raw-string.
Any suggestions?

bio/freebayes/wrapper.py
dlaehnemann: I think we might be referring to different places in the code, here. Just quickly look at it via the [`Files changes` tab](https://github.com/snakemake/snakemake-wrappers/pull/86/files) and maybe my comment becomes clearer.

bio/freebayes/wrapper.py
dlaehnemann: Generally, sed should accept `\t` per default, see e.g. the examples here (but I also quickly tested this in a bash terminal):
https://www.grymoire.com/Unix/Sed.html#uh-4

So I think the culprit here really is the Python format minilanguage, for which you just need to escape any `\` into `\\`. Have your tried my suggested code version, where all the `\` are double-escaped? I.e. `\\(`, `\\)`, `\\t`, `\\1`, and `\\2`?

FelixMoelder:
For some reason the test does not work although it calls a command that's running fine when called manually:
`(freebayes-parallel <(bedtools intersect -a =(sed "s/:\([0-9]*\)-\([0-9]*\)$/$(printf '\t')\1$(printf '\t')\2/g" <(fasta_generate_regions.py genome.fasta.fai 100000)) -b regions.bed | sed "s/$(printf '\t')\([0-9]*\)$(printf '\t')\([0-9]*\)$/:\1-\2/g") 2  -f genome.fasta mapped/a.bam | bcftools view -Ob - > calls/a.bcf)  2> logs/freebayes/a.log`

The log-file says: 
```
/bin/bash: process substitution: line 0: syntax error near unexpected token `('
Failed to read from standard input: unknown file type
```

Any idea whats going one?

dlaehnemann:
I think `=(` this in front of the first `sed` command should instead be `<(`, right? I corrected that in my code suggestion, but did not mention it earlier. Maybe that's the reason?

FelixMoelder:
> I think `=(` this in front of the first `sed` command should instead be `<(`, right? I corrected that in my code suggestion, but did not mention it earlier. Maybe that's the reason?

You are right. I previously had to change to `=(`as this writes to a temp file instead of `/dev/fd` as some piece of code was only working with temp files. But `=(` is limited to zsh which is why snakemake fails.

So, any further suggestions or are we good to go?

dlaehnemann:
Upon testing a bit in an interactive python3 shell and using print to produce what actually comes out, here's which `\` to escape, namely only the numbers, i.e. `\\1` and `\\2`. All the other ones should be respected by the `.format()`, so you should be able to simply put `\t`, `\(`, and `\)`. Also, I would quote the actual sed command `s///` with single quotes, to not have to escape the double quotes, so using `'s///'`. And I would remove the `g` option from the sed command, because that allows for multiple subsitutitions per line, but we only want one (or is the original file with multiple `chr:start-stop` entries per line?). So I would opt for the following code:

```
        regions = (
           "<(bedtools intersect -a "
            "<(sed 's/:\([0-9]*\)-\([0-9]*\)$/\t\\1\t\\2/' "
            "{regions}) -b {snakemake.input.regions} | "
            "sed 's/\t\([0-9]*\)\t\([0-9]*\)$/:\\1-\\2/\')"
        ).format(regions=regions, snakemake=snakemake)
```

This should do the job.

bio/freebayes/wrapper.py
dlaehnemann: Ah, OK. Upon looking at this again, I now got it. Sorry for the noise.

dlaehnemann:
@johanneskoester Could you add another review and possibly merge?

FelixMoelder:
> Upon testing a bit in an interactive python3 shell and using print to produce what actually comes out, here's which `\` to escape, namely only the numbers, i.e. `\\1` and `\\2`. All the other ones should be respected by the `.format()`, so you should be able to simply put `\t`, `\(`, and `\)`. Also, I would quote the actual sed command `s///` with single quotes, to not have to escape the double quotes, so using `'s///'`. And I would remove the `g` option from the sed command, because that allows for multiple subsitutitions per line, but we only want one (or is the original file with multiple `chr:start-stop` entries per line?). So I would opt for the following code:
> 
> ```
>         regions = (
>             "<(bedtools intersect -a "
>             "<(sed 's/:\([0-9]*\)-\([0-9]*\)$/\t\\1\t\\2/' "
>             "{regions}) -b {snakemake.input.regions} | "
>             "sed 's/\t\([0-9]*\)\t\([0-9]*\)$/:\\1-\\2/')"
>         ).format(regions=regions, snakemake=snakemake)
> ```
> 
> This should do the job.

Acctually this creates a command like 
`
freebayes-parallel <(bedtools intersect -a <(sed 's/:\([0-9]*\)-\([0-9]*\)$/     \1      \2/' <(fasta_generate_regions.py genome.fasta.fai 100000)) -b regions.bed | sed 's/ \([0-9]*\)      \([0-9]*\)$/:\1-\2/') 2  -f genome.fasta mapped/a.bam
`
which we had before.

If I only execute `bedtools intersect -a <(sed 's/:\([0-9]*\)-\([0-9]*\)$/     \1      \2/' <(fasta_generate_regions.py genome.fasta.fai 100000)) -b regions.bed | sed 's/ \([0-9]*\)      \([0-9]*\)$/:\1-\2/'` it fails with the following message:

```
Error: unable to open file or unable to determine types for file /dev/fd/17

- Please ensure that your file is TAB delimited (e.g., cat -t FILE).
- Also ensure that your file has integer chromosome coordinates in the
  expected columns (e.g., cols 2 and 3 for BED).
```

Looks like tabs are interpreted again.

dlaehnemann:
I'm now trying double-escaping the tabs, i.e. `\\t`.

But for reference: where do you check which command actually gets run? I don't see it in the tests. Do you have a local pipeline where you use the development version of the wrapper? So far, I was only test with minimal examples like `print( "sed 's/:\(test\)/\\t\\1/' {bed}".format(bed="BED") )` to see the substitution behaviour, and it didn't come up with the `\t` to whitespace substitution...

dlaehnemann:
Also: Why then does the `freebayes_bed` test run through without complaining? Maybe I'm misunderstanding what it is testing?

FelixMoelder:
> I'm now trying double-escaping the tabs, i.e. `\\t`.
> 
> But for reference: where do you check which command actually gets run? I don't see it in the tests. Do you have a local pipeline where you use the development version of the wrapper? So far, I was only test with minimal examples like `print( "sed 's/:\(test\)/\\t\\1/' {bed}".format(bed="BED") )` to see the substitution behaviour, and it didn't come up with the `\t` to whitespace substitution...

Actually the tab-thing was my biggest issue why I had to try a lot .
The best way was my `printf` variant which is describt here: https://unix.stackexchange.com/questions/145299/simple-sed-replacement-of-tabs-mysteriously-failing

This also requires double quotes.

FelixMoelder:
> Also: Why then does the `freebayes_bed` test run through without complaining? Maybe I'm misunderstanding what it is testing?

The tests are not really helpful as they only check for for `exit 0` as the files are minimalistic and create an empty bcf-file.
As the critical part was the regex and the correct interpretation by python I had to manually check the command being created and execute parts (mostly `bedtools intersect including sed`) it so see if the bed-regions piping creates the right output.

FelixMoelder:
> I'm now trying double-escaping the tabs, i.e. `\\t`.
> 
> But for reference: where do you check which command actually gets run? I don't see it in the tests. Do you have a local pipeline where you use the development version of the wrapper? So far, I was only test with minimal examples like `print( "sed 's/:\(test\)/\\t\\1/' {bed}".format(bed="BED") )` to see the substitution behaviour, and it didn't come up with the `\t` to whitespace substitution...

One thing we I just tried what we could do is replacing `printf '\\t'` by `printf '\t'` or `echo '\t'`.
Both is working, but python will replace `\t` by an actually tab. While this is working, I would prefer escaping here, as real tabs in a command line are not really readable and may complicate debugging.

dlaehnemann:
Of course it needs to work, but I still think that this looks pretty much unreadable and would prefer anything that's a bit cleaner. Also, when I test `print`ing out a string with a double-escaped `\\t` tab character in an interactive python session, it does not substitute it into an actual tab space. Or am I missing some step where this substitution happens? What steps do you do to test the strings?

For my testing, see e.g. this minimal example:
```
❯ python
Python 3.8.2 | packaged by conda-forge | (default, Apr 16 2020, 18:04:51) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> print( "sed 's/\\t:\(c\)/\\1\\t/' {b} {{b2}}".format(b="b").format(b2="b2") )
sed 's/\t:\(c\)/\1\t/' b b2
```

And if this still does not work, because for some reason the default sed that snakemake uses cannot interpret the `\t` tab escape, reading through the StackExchange question you posted earlier (<https://unix.stackexchange.com/a/145385>) suggests that a cleaner bash alternative of making this work would be to prepend the subsitution command string with a `$`, i.e.:
```
>>> print( "sed $'s/\\t:\(c\)/\\1\\t/' {b} {{b2}}".format(b="b").format(b2="b2") )
sed $'s/\t:\(c\)/\1\t/' b b2
```

Sorry for the hassle, I just want to make this as maintainable as possible...

FelixMoelder:
> ```
> >>> print( "sed $'s/\\t:\(c\)/\\1\\t/' {b} {{b2}}".format(b="b").format(b2="b2") )
> sed $'s/\t:\(c\)/\1\t/' b b2
> ```
> 
> Sorry for the hassle, I just want to make this as maintainable as possible...

Indeed, the parsing works which is not the issue. The issue here is, that the sed expression is not valid and returns a syntax error when executing and I was not able to parse a valid expression in a `$''` closure. 

I fully get, that the wrapper should be maintainable but it already cost me two days without a better result while the current release stays broken und a urgently need a working wrapper.
This is why I would suggest two merge this while it is working and open a refinement branch.

johanneskoester:
Yes, we merge now in order to fix the master branch. Then, please open an refinement PR in order to get this more readable. We can also write a Python function for this. The impact compared to the freebayes runtime should be minimal. Also note that there is `r""`, which is an unescaped raw string, where each "\" is interpreted as such.

dlaehnemann:
@FelixMoelder will you open the refinement PR, or should I? I don't currently have this repo as a local copy, but obviously that's only a `git clone` away. However, you seem to have the better testing strategy, for now.

PR #87, opened 2020-04-20
Refinement of sed call for freebayes

FelixMoelder:
As the current freebayes wrapper builds a not-so-nice sed command this PR aims for refactoring.

The main difficulty is not parsing the command by python but to replace the regex by a valid one being correctly interpreted by sed.

The current sed comment for replacing colons and hyphens by tabs looks as following (without separating the chromosome name):
`sed "s/:\([0-9]*\)-\([0-9]*\)$/$(printf '\t')\1$(printf '\t')\2/g"`

For testing I propose the minimal example:
`echo "MT0:1-AB:10-100" | sed "s/:\([0-9]*\)-\([0-9]*\)$/$(printf '\t')\1$(printf '\t')\2/g"`
The result should be `MT0:1-AB	10	100`

### DIFF
--- a/bio/freebayes/wrapper.py
+++ b/bio/freebayes/wrapper.py
@@ -27,9 +27,9 @@ else:
     if snakemake.input.get("regions", ""):
         regions = (
             "<(bedtools intersect -a "
-            "<(sed 's/:\([0-9]*\)-\([0-9]*\)$/\\t\\1\\t\\2/g' "
+            r"<(sed 's/:\([0-9]*\)-\([0-9]*\)$/\t\1\t\2/g' "
             "{regions}) -b {snakemake.input.regions} | "
-            'sed "s/\\t\([0-9]*\)\\t\([0-9]*\)$/:\\1-\\2/g")'
+            r"sed 's/\t\([0-9]*\)\t\([0-9]*\)$/:\1-\2/g')"
         ).format(regions=regions, snakemake=snakemake)
     freebayes = ("freebayes-parallel {regions} {snakemake.threads}").format(
         snakemake=snakemake, regions=regions
--- a/bio/freebayes/wrapper.py
+++ b/bio/freebayes/wrapper.py
@@ -23,7 +23,6 @@ else:
     regions = "<(fasta_generate_regions.py {snakemake.input.ref}.fai {chunksize})".format(
         snakemake=snakemake, chunksize=chunksize
     )
-    # sed command needs to be improved
     if snakemake.input.get("regions", ""):
         regions = (
             "<(bedtools intersect -a "
--- a/bio/freebayes/wrapper.py
+++ b/bio/freebayes/wrapper.py
@@ -27,9 +27,9 @@ else:
     if snakemake.input.get("regions", ""):
         regions = (
             "<(bedtools intersect -a "
-            r"<(sed 's/:\([0-9]*\)-\([0-9]*\)$/\t\1\t\2/g' "
+            r"<(sed 's/:\([0-9]*\)-\([0-9]*\)$/\t\1\t\2/' "
             "{regions}) -b {snakemake.input.regions} | "
-            r"sed 's/\t\([0-9]*\)\t\([0-9]*\)$/:\1-\2/g')"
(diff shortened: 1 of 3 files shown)


dlaehnemann:
Thanks for starting this PR and for the testing example you're using.

Funnily, for me, using `\t` works exactly as intended within that example, i.e. I get:

```
❯ echo "MT0:1-AB:10-100" | sed 's/:\([0-9]*\)-\([0-9]*\)$/\t\1\t\2/g'
MT0:1-AB	10	100
```

Maybe what you are seeing is due to a different version of `sed` used in OS X? I would hope, that the version used in Snakemake by default is fixed, as is the use of bash for shell command interpretation. So maybe you can test the following minimal `Snakefile` run through `snakemake` on your system, to double-check that Snakemake does not use your system sed (note the `r` before the `shell`-string -- this is needed to avoid having to double-escape the capture group calls for the Python string parsing that is happening, i.e. so that we don't have to write `\\1` and `\\2`):

```
rule all:
    output: "results.txt"
    shell:
        r"echo 'MT0:1-AB:10-100' | sed 's/:\([0-9]*\)-\([0-9]*\)$/\t\1\t\2/' | tee {output}"
```

And in any case, the safest way to ensure that this works on any system is probably to add the `sed` from `conda-forge` to the environment `yaml` for this wrapper and ensure that the syntax works with this version of sed. I.e. to test locally, my setup is:

```
conda create -n sed sed
conda activate sed
❯ echo "MT0:1-AB:10-100" | sed 's/:\([0-9]*\)-\([0-9]*\)$/\t\1\t\2/g'
MT0:1-AB	10	100
```

FelixMoelder:
> Thanks for starting this PR and for the testing example you're using.
> 
> Funnily, for me, using `\t` works exactly as intended within that example, i.e. I get:
> 
> ```
> ❯ echo "MT0:1-AB:10-100" | sed 's/:\([0-9]*\)-\([0-9]*\)$/\t\1\t\2/g'
> MT0:1-AB	10	100
> ```
> 
> Maybe what you are seeing is due to a different version of `sed` used in OS X? 
> ```

You are right! This seems to be an issue of of the freeBSD sed version used by macOS...
When installing sed via conda your command is working just fine.
So, adding sed to the environment should be the right way to ensure that sed is working right.

FelixMoelder:
So, I guess the last commit should solve this.

dlaehnemann:
Halleluja! :D

Two little, mostly cosmetic suggestions:

* [ ] drop the `g` at the end of the `sed 's///'` substitution commands (this indicates, that any number of occurrences per line should be subsituted, but we only expect one occurrence per line and anchor it correctly to the line end)
* [ ] make the use of `'` and `"` consistent across lines, i.e. by switching their use on the second `sed` line

FelixMoelder:
The inconsistencies were unintended and fixed now.

I also added a raw-string so we dont need `\\` in front of tabs and capture groups.
But as there are multi-line string this now looks a bit ugly:
```
        regions = (
            "<(bedtools intersect -a "
            r"<(sed 's/:\([0-9]*\)-\([0-9]*\)$/\t\1\t\2/' "
            "{regions}) -b {snakemake.input.regions} | "
            r"sed 's/\t\([0-9]*\)\t\([0-9]*\)$/:\1-\2/')"
        ).format(regions=regions, snakemake=snakemake)
```

bio/freebayes/wrapper.py
```diff
@@ -23,12 +23,13 @@
     regions = "<(fasta_generate_regions.py {snakemake.input.ref}.fai {chunksize})".format(
         snakemake=snakemake, chunksize=chunksize
     )
+    # sed command needs to be improved
```
dlaehnemann, line 26: Ah, and this line can be dropped now! :D

FelixMoelder:
Finally! I feel years older...and a little wiser now. :D

dlaehnemann:
I also prefer the syntax with the `r""` strings and don't mind the look of that in turn for dropping the double `\\`.

@johanneskoester, I think this is good to merge.

PR #88, opened 2020-04-21
add prosolo wrappers

dlaehnemann:


bio/prosolo/control-fdr/test/Snakefile
```diff
@@ -0,0 +1,19 @@
+rule prosolo_fdr_control:
+    input:
+         "variant_calling/{sc}.{bulk}.prosolo.bcf"
+    output:
+         "fdr_control/{sc}.{bulk}.prosolo.fdr.bcf"
+    message:
+        "Testing ProSolo"
```
johanneskoester, line 7:
```suggestion
```

bio/prosolo/control-fdr/test/Snakefile
```diff
@@ -0,0 +1,19 @@
...
+    threads:
+        1
+    params:
+        # comma-separated set of events for whose (joint)
+        # false discovery rate you want to control
+        events = "ADO_TO_REF,HET",
```
johanneskoester, line 13: Are you sure? In varlociraptor, this is space separated.

bio/prosolo/single-cell-bulk/wrapper.py
```diff
@@ -0,0 +1,21 @@
...
+__email__ = "david.laehnemann@uni-due.de"
+__license__ = "MIT"
+
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+shell(
+    "( prosolo single-cell-bulk "
+    "--omit-indels "
```
johanneskoester, line 14: Why does the wrapper enforce omitting of indels?

bio/prosolo/single-cell-bulk/test/Snakefile
```diff
@@ -0,0 +1,19 @@
...
+        single_cell_index = "data/mapped/{sc}.bam.bai",
+        bulk = "data/mapped/{bulk}.bam",
+        bulk_index = "data/mapped/{bulk}.bam.bai",
+        ref = "data/genome.fasta",
+        ref_idx = "data/genome.fasta.fai",
+        candidates = "data/{sc}.{bulk}.prosolo_candidates.bcf",
+    output:
+        "variant_calling/{sc}.{bulk}.prosolo.bcf"
+    message:
+        "Testing ProSolo"
```
johanneskoester, line 13:
```suggestion
```

bio/prosolo/single-cell-bulk/test/Snakefile
```diff
@@ -0,0 +1,19 @@
...
+    threads:
```
johanneskoester, line 14: Add params->extra, for passing addiitonal args.

bio/prosolo/control-fdr/test/Snakefile
dlaehnemann: Jep, this was my initial implementation that still holds for the last release of `prosolo` (`0.6.1`). But otherwise, this should be caught by the tests once I have created proper basic testing data.

bio/prosolo/single-cell-bulk/wrapper.py
dlaehnemann: Because that is what we have previously tested and compared with the other tools. So for now (`prosolo 0.6.1`), I would probably stick with this fixed. But once we update to a newer version of varlociraptor functionality, we can also remove this, as I think the amplification bias model (and the modelling of the bulk background sample) should apply to indels just as well.

bio/prosolo/single-cell-bulk/test/Snakefile
dlaehnemann: Done.

dlaehnemann:
@johanneskoester Two questions:

1. Any idea why the test for `bio/prosolo/control-fdr` complains that: `Conda env file does not exist: /tmp/tmprx6wbo6t/master/bio/prosolo/single-cell-bulk/environment.yaml`? It should not look for the `environment.yaml` of the other wrapper to start with -- but also, this other `environment.yaml` should be there. Maybe some sort of optimisation in sharing the environment plus a race condition, where `control-fdr` references an environment that `single-cell-bulk` has not yet created or already deleted?
2. I merged in the latest changes from master to get the GitHub Actions to run but it seems like I might have regressed some of the tests. I guess I'll have to double-cross-check with the current master branch to get this right?

dlaehnemann:
Okay, I cleaned up the tests mess. But the weird test failure of `test_prosolo_fdr` remains. @johanneskoester could you have a look:
https://github.com/snakemake/snakemake-wrappers/runs/652942636

PR #91, opened 2020-04-30
Split snpeff into download and annotate.

johanneskoester:


jafors:
I added a snpEff download wrapper in [PR81](https://github.com/snakemake/snakemake-wrappers/pull/81), but I feel your idea of putting it into the snpEff subdirectory makes more sense. But maybe we should delete my version once this is ready?

johanneskoester:
Oh, sorry, I forgot about this! Yes, I'd prefer this structure.

PR #98, opened 2020-05-08
add gdc-client download wrapper

dlaehnemann:


bio/gdc-client/download/test/Snakefile
```diff
@@ -0,0 +1,20 @@
+rule gdc_download:
+    output:
+        # * {UUID} is the column `id` from a GDC manifest file
+        # * {filename} is the column `filename` from a GDC manifest file
+        # * the `gdc/` folder is used to avoid ambiguity as filename can
+        #   contain any file extension and theoretically generate any type
+        #   of file
+        "gdc/{UUID}/{filename}"
```
johanneskoester, line 8: Can we move the UUID into params instead, and make this filename freely choosable?

johanneskoester: Should be possible by running gdc-client targeting a `TemporaryDirectory` and moving the file after completion.

dlaehnemann: Jep, we could of course do that.

My thought behind this was that this way, the raw input files for a workflow retain the exact names they have in GDC. They could then be renamed by using an input function in a downstream analysis step. But the unit -> UUID mapping via a column in the `unit.tsv` that is needed for that could of course also be used for a function in params that retrieves the UUID based on the filename. Then, we just have to include that mapping in the tests here directly...

bio/gdc-client/download/test/Snakefile
```diff
@@ -0,0 +1,20 @@
...
+    log:
+        "logs/gdc-client/download/{UUID}.{filename}.log"
+    params:
+        # a gdc_token is only required for controlled access samples,
+        # leave blank otherwise (`gdc_token=""`) or skip entirely
+        gdc_token="gdc/gdc-user-token.2020-05-07T10_00_00.555Z.txt",
+        # for valid extra command line arguments, check command line help or:
+        # https://docs.gdc.cancer.gov/Data_Transfer_Tool/Users_Guide/Data_Download_and_Upload/
+        extra = ""
```
johanneskoester, line 17:
```suggestion
        extra = "",
        uuid="..."
```

dlaehnemann:
@johanneskoester 
It's now as you suggested, and I added a second example usage of the wrapper in `test/Sankefile` to illustrate how to use it with UUIDs where multiple files are downloaded, using `bam` plus index, annotation and logs directory as the example. However, I have not included a test for this second example, as this would mean downloading larger amounts of data. Hope that's OK.

PR #102, opened 2020-05-14
Fix reference ensembl variation2

kilpert:


johanneskoester:
Thanks a lot. I have taken this as inspiration for a slightly simpler approach, that can be found in PR #104. The downside is restriction to release >=98, but I think that should be fine, since hg38 is still available there.

PR #105, opened 2020-05-18
samtools idxstats wrapper

AntonieV:
snakemake-wrapper for samtools idxstats


bio/samtools/idxstats/environment.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
+  - bioconda
+  - conda-forge
+dependencies:
+  - samtools ==1.9
```
dlaehnemann, line 5: `samtools 1.10` is there, and also [available via `bioconda`](https://bioconda.github.io/recipes/samtools/README.html?highlight=samtools#recipe-Recipe%20&#x27;samtools&#x27;), and while you were working on it, the other samtools wrappers have been updated. So maybe switch up to `- samtools ==1.10` before merging.

bio/samtools/idxstats/meta.yaml
```diff
@@ -0,0 +1,8 @@
+name: samtools idxstats
+description: Use samtools to retrieve and print stats form indexed bam, sam or cram files
+authors:
+  - Antonie Vietor
+input:
+  - indexed bam, sam or cram file (.bam, .sam, .cram)
```
dlaehnemann, line 6: My personal preference for the order these file extensions (it just doesn't sound right in any other order, don't ask why! :sweat_smile: ):
```
  - indexed sam, bam or cram file (.sam, .bam, .cram)
```

dlaehnemann: Also, as you suggested, I'd add the index file as a required input, as well.

bio/samtools/idxstats/test/Snakefile
```diff
@@ -0,0 +1,12 @@
+rule samtools_idxstats:
+    input:
+        # bam input files should be indexed by samtools index
+        # the index files should be in the same directory as the bam files
+        # without index files reverting to slow method
+        bam = "mapped/{sample}.bam"
```
dlaehnemann, line 6: As mentioned above and as you suggested, I'd add the `idx = ` equivalent as a second required input here, which also enables dropping the comment and makes the wrapper simpler and clearer to use.

test.py
```diff
@@ -770,6 +770,13 @@ def test_samtools_flagstat():
     )
 
 
+def test_samtools_idxstats():
```
dlaehnemann, line 773: I know the order of the tests is all over the place. But whenever I add one recently, I try to at least put somewhat into alphanumerical sort order. So in this case, this would be a bit further down among the other samtools tests. Sorry for another round of pedantic review... :sweat_smile:

dlaehnemann: Actually, nevermind. Had another look at the sort order of the samtools wrappers, and they are far from anything resembling an acutal sort order... So forget about it...

dlaehnemann:
Thanks again @AntonieV , this looks good!
@johanneskoester: This should be good to go, but please feel free to have another look!

bio/samtools/idxstats/test/Snakefile
```diff
@@ -0,0 +1,10 @@
+rule samtools_idxstats:
+    input:
+        bam = "mapped/{sample}.bam",
+        idx = "mapped/{sample}.bam.bai"
```
johanneskoester, line 4:
```suggestion
        bam="mapped/{sample}.bam",
        idx="mapped/{sample}.bam.bai"
```

PR #106, opened 2020-05-22
GDC BAM Slicing download wrapper

dlaehnemann:
This is a wrapper to use the GDC BAM Slicing API to download only specified regions of GDC BAM files with curl.

As this needs a valid GDC Access Token to work, the test has to work around the error this creates. Interestingly, the curl command exits with exit status `0`, even if an error in JSON format is returned by the API request. So the wrapper handles that case and will fail upon small files that contain the word `error`. And the provided test merely goes looking for the expected token error in the log file, at least ensuring that everything else works fine.


bio/gdc-api/bam-slicing/test/Snakefile
```diff
@@ -0,0 +1,20 @@
+rule gdc_api_bam_slice_download:
+    output:
+        bam = "raw/{sample}.bam",
```
johanneskoester, line 3:
```suggestion
        bam="raw/{sample}.bam",
```

bio/gdc-api/bam-slicing/test/Snakefile
```diff
@@ -0,0 +1,20 @@
...
+        # the column `id` in the GDC manifest files, which can be used to
+        # systematically construct sample sheets)
+        uuid="092c8a6d-aad5-41bf-b186-e68e613c0e89",
+        # a gdc_token is required for controlled access and all BAM files
+        # on GDC seem to be controlled access (adjust if this changes)
+        gdc_token="gdc/gdc-user-token.2020-05-07T10_00_00.555Z.txt",
+        # provide wanted `region=` or `gencode=` slices joined with `&`
+        slices="region=chr22&region=chr5:1000-2000&region=unmapped&gencode=BRCA2",
+        # extra command line arguments passed to curl
+        extra = ""
```
johanneskoester, line 18:
```suggestion
        extra=""
```

bio/gdc-api/bam-slicing/test/Snakefile
dlaehnemann: I'll never get used to this recommended style... :D

PR #108, opened 2020-05-24
fix gdc api token reading

dlaehnemann:
With not being able to use a genuine token, the test didn't catch this -- the token has to pasted verbatim into the API request being made.
But as far as I can see, this does not pose the risk of exposing the token via snakemake logs, as the exact command being run is hidden behind the temporary `wrapper.py` in the `.snakemake/log` files. But @johanneskoester, if you see any possibility of the token being put somewhere where users might (accidentally) make it publicly available, please do say! We definitely don't want this as a side-effect, also not in any future versions of snakemake (e.g. if there are any plans of putting the temporary wrapper.py contents in the log eventually).

### DIFF
--- a/bio/gdc-api/bam-slicing/wrapper.py
+++ b/bio/gdc-api/bam-slicing/wrapper.py
@@ -17,7 +17,7 @@ if token_file == "":
     raise ValueError(
         "You need to provide a GDC data access token file via the 'token' in 'params'."
     )
-token=""
+token = ""
 with open(token_file) as tf:
     token = tf.read()
 os.environ["CURL_HEADER_TOKEN"] = "'X-Auth-Token: {}'".format(token)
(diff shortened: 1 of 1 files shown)


bio/gdc-api/bam-slicing/wrapper.py
```diff
@@ -4,19 +4,23 @@
...
+token_file = snakemake.params.get("gdc_token", "")
+if token_file == "":
     raise ValueError(
-        "You need to provide a GDC data access token via the 'token' in 'params'."
+        "You need to provide a GDC data access token file via the 'token' in 'params'."
     )
+token=""
+with open(token_file) as tf:
+    token = tf.read()
+os.environ["CURL_HEADER_TOKEN"] = "'X-Auth-Token: {}'".format(token)
```
dlaehnemann, line 23: @johanneskoester Is this what you meant by putting the token into an environment variable and then using that (see below).

PR #109, opened 2020-05-26
include BAM indexes BAI as freebayes required input

dlaehnemann:
In my local pipeline, with BAMs covering only one chromosome, freebayes failed to recompute indexes on the fly. So it seems like the BAI files should be required input alongside the BAM files. As these are easily and quickly generated with another wrapper and aren't very big, it is in any case cleaner to have them generated once and then lying around next to the BAMs---this way, whenever they are needed with an index again (e.g. delly), it's there.

### DIFF
--- a/bio/freebayes/test/Snakefile
+++ b/bio/freebayes/test/Snakefile
@@ -2,7 +2,9 @@ rule freebayes:
     input:
         ref="genome.fasta",
         # you can have a list of samples here
-        samples="mapped/{sample}.bam"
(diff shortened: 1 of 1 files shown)


johanneskoester:
Definitely. I did not know that it tries to recompute them on the fly. This would even be dangerous because it could generate a race condition in case of parallelization (which might explain an issue that @jafors and @FelixMoelder once had with the varlociraptor pipeline). Good catch!

PR #110, opened 2020-05-27
bamtools filter and bamtools stats

AntonieV:
wrapper for bamtools filter via command line and without json-script, wrapper for bamtools filter via json-script and wrapper for bamtools stats


bio/bamtools/filter/test/Snakefile
```diff
@@ -0,0 +1,18 @@
+rule bamtools_filter:
+    input:
+        "{sample}.bam"
+    output:
+        "filtered/{sample}.bam"
+    params:
+        # optional parameters
+        tag = "NM:<4",    # string of an key:value pair
```
dlaehnemann, line 8: Here, users might want to filter by multiple tags, so maybe we should make this a list?
```
        tag = [ "NM:<4", "MQ:>=10" ],    # list of key:value pair strings
```
For a suggestion on how to parse this, see the comment in the wrapper.

bio/bamtools/filter/wrapper.py
```diff
@@ -0,0 +1,48 @@
...
+extra_limits = ""
+tag = snakemake.params.get("tag")
+min_size = snakemake.params.get("min_size")
+max_size = snakemake.params.get("max_size")
+min_length = snakemake.params.get("min_length")
+max_length = snakemake.params.get("max_length")
+additional_params = snakemake.params.get("additional_params")
+
+if tag and tag is not None:
+    params = params + ' -tag "' + tag + '"'
```
dlaehnemann, line 21: If we make tag a list of strings, this should become something like:
```
    params = params + ' -tag "' + ' -tag "'.join(tag) + '"'
```

bio/bamtools/filter/environment.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
+  - bioconda
+  - conda-forge
+dependencies:
+  - bamtools ==2.5.1
```
dlaehnemann, line 5: For some reason, GitHub is complaining about a missing newline at the end of the file, here? Maybe this has something to do with the Editor / IDE configuration you use and might be something worth checking on your end in general? I think the expectation is to just have a newline character at the end of every line, including the last.

bio/bamtools/stats/test/Snakefile
```diff
@@ -0,0 +1,11 @@
+rule bamtools_stats:
+    input:
+        "{sample}.bam"
+    output:
+        "{sample}.bamstats.txt"
```
dlaehnemann, line 5: Very minor thing: I think `"{sample}.bamstats"` without the `.txt.` suffix is just fine.

bio/bamtools/filter/wrapper.py
johanneskoester: More idiomatic:
```
" ".join(map("-tag {}".format, tags))
```

dlaehnemann: Thanks, that's quite a bit cleaner!

PR #111, opened 2020-06-02
Shovill wrapper

sk-sahu:
Wrapper for [Shoviil](https://github.com/tseemann/shovill) added. 

Note: By default, Shovill generates raw assembly with `assembler_name.fa`. To make that unified in all case the file renamed to `raw_assembly.fa`.

Now the output will have
* `raw_assembly.fa`
* `contigs.fa`


bio/shovill/test/Snakefile
```diff
@@ -0,0 +1,15 @@
+rule shovill:
+  input:
+    r1="reads/input_R1.fq.gz",
+    r2="reads/input_R2.fq.gz",
+  output:
+    raw_assembly="assembly/raw_assembly.fa",
+    contigs="assembly/contigs.fa"
```
dlaehnemann, line 7: Can we generalise input and output to use a `{sample}` wildcard, so that in theory you can assemble more than one sample within the same pipeline. I.e. I would suggest:
```
  input:
    r1="reads/{sample}_R1.fq.gz",
    r2="reads/{sample}_R2.fq.gz",
  output:
    raw_assembly="assembly/{sample}.assembly.fa",
    contigs="assembly/{sample}.contigs.fa"
```

This will require adjust the example input data file names to match whatever you request as output in the test.

dlaehnemann: Also, in case someone wants to assemble the same data with different assemblers within one pipeline, it might make sense to include the assembler as a wildcard instead of as a params entry. This way, the assembler used is documented in the outpu file name. I.e.:
```
  input:
    r1="reads/{sample}_R1.fq.gz",
    r2="reads/{sample}_R2.fq.gz",
  output:
    raw_assembly="assembly/{sample}.{assembler}.assembly.fa",
    contigs="assembly/{sample}.{assembler}.contigs.fa"
```

bio/shovill/test/Snakefile
```diff
@@ -0,0 +1,15 @@
...
+  params:
+    assembler="spades",
+    extra=""
+  log:
+    "logs/shovill/job.log"
```
dlaehnemann, line 12: `log` will need to use all the wildcards from output, i.e. something like:
```
     "logs/shovill/{sample}.{assembler}.log"
```

bio/shovill/wrapper.py
```diff
@@ -0,0 +1,28 @@
...
+    " --outdir assembly"
+    " --R1 {snakemake.input.r1}"
+    " --R2 {snakemake.input.r2}"
+    " --cpus {snakemake.threads}"
+    " {snakemake.params.extra}) {log}"
+)
+
+os.rename(
+    "assembly/" + snakemake.params.assembler + ".fasta", "assembly/raw_assembly.fa"
+)
```
dlaehnemann, line 28: When someone runs the same assembler on multiple input samples within a pipeline, we would get a race condition with the current setup. I would suggest putting initial results into a unique temporary directory and then moving them into the requested output location with the requested output file names at the end. For a suggestion of how such a setup can look, see e.g. the `fastqc` wrapper (you also need the `from tempfile import TemporaryDirectory` at the top):

https://github.com/snakemake/snakemake-wrappers/blob/5c11e7279b7fe91c2bc7d8cc6c71cb7c96ccfc61/bio/fastqc/wrapper.py#L28-L46

bio/shovill/wrapper.py
```diff
@@ -0,0 +1,28 @@
+"""Snakemake wrapper for shovill."""
+
+__author__ = "Sangram Keshari Sahu"
+__copyright__ = "Copyright 2020, Sangram Keshari Sahu"
+__email__ = "sangramsahu15@gmail.com"
+__license__ = "MIT"
+
+import os
```
dlaehnemann, line 8: This import is not needed if you do the suggested moving of files with the shell. But do feel free to do the moving with python's `os`.

bio/shovill/wrapper.py
```diff
@@ -0,0 +1,28 @@
...
+from snakemake.shell import shell
+
+# Placeholder for optional parameters
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+params = snakemake.params.get("extra", "")
+
+shell(
+    "(shovill"
+    " --assembler {snakemake.params.assembler}"
+    " --outdir assembly"
```
dlaehnemann, line 19: `--outdir {outdir}` should probably also be adjustable via the rule definition and simply parsed from `snakemake.output.raw_assembly` with `os.path`, i.e.:
```
outdir = path.dirname(snakemake.output.raw_assembly)
```

bio/shovill/test/Snakefile
```diff
@@ -0,0 +1,15 @@
+rule shovill:
+  input:
+    r1="reads/input_R1.fq.gz",
+    r2="reads/input_R2.fq.gz",
```
dlaehnemann, line 4: There is an extra comma at the end of this line. Maybe this is the culprit currently making the test fail.

sk-sahu:
Hi, @dlaehnemann Thanks for the valuable suggestions. The wildcard variable is a great point, Somehow I missed it. I'll work on these and do a commit soon.

sk-sahu:
@dlaehnemann I have updated Snakefile in the test directory.

But in the [Github-Actions test](https://github.com/snakemake/snakemake-wrappers/pull/111/checks?check_run_id=800205522#step:7:218) this not getting updated! I'm not sure why?

test.py
```diff
@@ -78,6 +78,20 @@ def run(wrapper, cmd, check_log=None):
...
+            "snakemake",
+            "assembly/raw_assembly.fa",
+            "assembly/contigs.fa",
+            "--cores", "1", 
+            "--use-conda", 
+            "-F", 
+        ],
+    )
+
+
```
jafors, line 94: It looks like the new error occurs here. Since you rewrote the output as `{sample}.{assembler}.assembly.fa` and `{sample}.{assembler}.contigs.fa`, you need to use this structure when using the Snakefile and fill the wildcards with the desired values. 

In your case, that would most probably be `input.spades.assembly.fa` and `input.spades.contigs.fa`, since you want to use spades and your test files are called `input_R*.fq.gz`.

bio/shovill/test/Snakefile
```diff
@@ -0,0 +1,15 @@
+rule shovill:
+  input:
+    r1="reads/{sample}_R1.fq.gz",
+    r2="reads/{sample}_R2.fq.gz"
+  output:
+    raw_assembly="assembly/{sample}.{assembler}.assembly.fa",
+    contigs="assembly/{sample}.{assembler}.contigs.fa"
+  params:
+    assembler="spades",
+    extra=""
```
jafors, line 10: I like the idea of writing the used assembler as a wildcard for the output files! This will also allow you to determine the `assembler`-parameter directly from the output files, so you don't need to give it again in the `params` field.

bio/shovill/wrapper.py
```diff
@@ -0,0 +1,32 @@
...
+with TemporaryDirectory() as tempdir:
+    shell(
+        "(shovill"
+        " --assembler {snakemake.params.assembler}"
+        " --outdir {tempdir}"
+        " --R1 {snakemake.input.r1}"
+        " --R2 {snakemake.input.r2}"
+        " --cpus {snakemake.threads}"
+        " {snakemake.params.extra}) {log}"
+    )
```
jafors, line 24: Here, you could just use `{params}` instead of `{snakemake.params.extra}` in the shell command since you already defined this variable with `params = snakemake.params.get("extra", "")`

You could also infer the assembler param directly from one of the output files, as a mentioned before. You could go for something like `assembler = path.basename(snakemake.output.raw_assembly).split('.')[1]` and then `--assembler {assembler}` but there is probably a more elegant solution to this.

sk-sahu:
Thanks @jafors for suggestions and fixes feedback. Taking assembler name from output file seems okay, a less param. For now, I have done the same.

bio/shovill/wrapper.py
```diff
@@ -0,0 +1,33 @@
...
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+params = snakemake.params.get("extra", "")
+
+# Determine the assembler from wildcard of output file name
+assembler = path.basename(snakemake.output.raw_assembly).split(".")[1]
+
+with TemporaryDirectory() as tempdir:
+    shell(
+        "(shovill"
+        " --assembler {assembler}"
```
dlaehnemann, line 22: Here, you can simply access the wildcard directly and drop the path parsing from above:

```
        " --assembler {snakemake.wildcards.assembler}"
```

The `wrapper.py` is like an external script and can access all the variables that snakemake provides by default, see here in the docs:
https://snakemake.readthedocs.io/en/latest/snakefiles/rules.html#external-scripts

bio/shovill/wrapper.py
```diff
@@ -0,0 +1,33 @@
...
+__copyright__ = "Copyright 2020, Sangram Keshari Sahu"
+__email__ = "sangramsahu15@gmail.com"
+__license__ = "MIT"
+
+from os import path
+from snakemake.shell import shell
+from tempfile import TemporaryDirectory
+
+# Placeholder for optional parameters
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
```
dlaehnemann, line 13: Did you intend for `stdout` not showing up in the log file? Because if I understand the CLI logic correctly, all output is explicitly written to files, so stdout might contain additional logging infos?

sk-sahu: No, It's better to have the `stdout` in a log file. Maybe, in this case, I need to change it  `stdout=False`

bio/shovill/wrapper.py
```diff
@@ -0,0 +1,33 @@
...
+from os import path
+from snakemake.shell import shell
+from tempfile import TemporaryDirectory
+
+# Placeholder for optional parameters
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+params = snakemake.params.get("extra", "")
+
+# Determine the assembler from wildcard of output file name
+assembler = path.basename(snakemake.output.raw_assembly).split(".")[1]
```
dlaehnemann, line 17: I think this is not needed any more.

bio/shovill/wrapper.py
```diff
@@ -0,0 +1,33 @@
...
+        " --assembler {snakemake.wildcards.assembler}"
+        " --outdir {tempdir} --force"
+        " --R1 {snakemake.input.r1}"
+        " --R2 {snakemake.input.r2}"
+        " --cpus {snakemake.threads}"
+        " {params}) {log}"
+    )
+
+    shell(
+        "mv {tempdir}/{assembler}.fasta {snakemake.output.raw_assembly}"
```
dlaehnemann, line 31: Instead of `{assembler}`, I'd also go for directly using `{snakemake.wildcards.assembler}`, here.

sk-sahu:
Thanks @dlaehnemann I have learned during this process :)

bio/shovill/meta.yaml
```diff
@@ -0,0 +1,5 @@
+name: shovill
+description: Assemble bacterial isolate genomes from Illumina paired-end reads.
+homepage: https://github.com/tseemann/shovill
+authors:
+  - Torsten Seemann
```
dlaehnemann, line 5: The authors field here is actually meant for the wrapper authors, so it's OK to put you name here. With the homepage link, it's clear who created the tool itself and that attribution should be clear.

dlaehnemann: Jep, just revert to having your name in there and commit.

dlaehnemann:
Thanks again, @sk-sahu , also for merging in the master---this saved me one step! :)

PR #112, opened 2020-06-03
Deepvariant wrapper

tetsuro90:
[Deepvariant](https://github.com/google/deepvariant) wrapper for calling SVs added.  
This wrapper uses biconda deepvariant package.  bioconda package is a littele bit different from the source.  (ex. bioconda package cannot output gvcf.)


bio/deepvariant/wrapper.py
```diff
@@ -0,0 +1,39 @@
...
+    "--cores {snakemake.threads} "
+    "--outfile {output_dir}/{basename}.tmp "
+    "--sample {basename} "
+    "--examples {output_dir}/{basename} "
+    "--model {snakemake.params.model} \n"
+    "dv_postprocess_variants.py "
+    "--ref {snakemake.input.ref} "
+    "--infile {output_dir}/{basename}.tmp "
+    "--outfile {snakemake.output.vcf} \n"
+    "rm -rf {output_dir}/{basename}.tmp ) {log}"
```
johanneskoester, line 38: Would it make sense to have these as three wrappers instead? Perhaps they even have different resource usages?

tetsuro90: @johanneskoester 
Thank you for your reply!

The resource implements the `run_deepvariant` command, which runs three commands sequentially: `make_example`, `call_variants` and `postprocess_variants` (conda package doesn't have `run_deepvariant` ). 
This is the same process as my wrapper. And, in the resorce, the following is mentioned
```
DeepVariant is composed of three programs: make_examples, call_variants, and postprocess_variants.
```
https://github.com/google/deepvariant/blob/r0.10/docs/deepvariant-details.md

bio/deepvariant/wrapper.py
```diff
@@ -0,0 +1,40 @@
...
+__license__ = "MIT"
+
+import os
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+extra = snakemake.params.get("extra", "")
+
+log_dir = os.path.dirname(str(snakemake.log))
+output_dir = os.path.dirname(str(snakemake.output))
```
johanneskoester, line 13: You can omit the str() call.

bio/deepvariant/wrapper.py
```diff
@@ -0,0 +1,40 @@
...
+
+# sample basename
+basename = os.path.splitext(os.path.basename(str(snakemake.input.bam)))[0]
```
johanneskoester, line 16: This sounds like we could have name conflicts. in the calls below. Instead, create a temporary directory with Python, and store the intermediate files in there.

tetsuro90: Thanks for the advice. I used Python's `tempfile.TemporaryDirectory()` to create temporary directory.

bio/deepvariant/wrapper.py
tetsuro90: I omitted the `str()`. But, like this `os.path.dirname(snakemake.output[0])`, I have to add index to execute successfully.

johanneskoester:
Excellent, thanks a lot!

PR #119, opened 2020-06-21
Wrapper for preseq lc_extrap to predict the yield for future experime…

AntonieV:
…nts with tests for .bam- and .bed-files.


bio/preseq/lc_extrap/test/Snakefile
```diff
@@ -0,0 +1,11 @@
+rule preseq_lc_extrap:
+    input:
+        "samples/{sample}.sorted.bam"
```
johanneskoester, line 3: Please mention in a comment that input can also be a bed.

bio/preseq/lc_extrap/wrapper.py
```diff
@@ -0,0 +1,22 @@
...
+__license__ = "MIT"
+
+import os
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+
+params = ""
+sm_input = str(snakemake.input)
+print(sm_input)
```
dlaehnemann, line 13: `snakemake.input` must be a list of strings, snakemake should even check this. So I think these two lines are not necessary and you could just use `snakemake.input[0]` in the line just below when retrieving the extension.

dlaehnemann:
Thanks @AntonieV , this looks good to merge for me. And good catch that you could merge the tests!

@johanneskoester , feel free to approve and merge.

PR #120, opened 2020-06-22
wrapper for picard CollectMultipleMetrics

AntonieV:
Wrapper for CollectMultipleMetrics with picard. Since only a single name or path without file extension is expected as output, this is specified using a wildcard. As an additional option, the amount of allocated memory can be specified.


bio/picard/collectmultiplemetrics/test/Snakefile
```diff
@@ -0,0 +1,30 @@
+rule collect_multiple_metrics:
+    input:
+         bam="mapped/{sample}.bam",
+         ref="genome.fasta"
+    output:
+        multiext("{path}{sample}",
```
dlaehnemann, line 6: Nice idea with the multiext, to make this long list a bit saner! :D

One thing: I don't think we need the `{path}` wildcard here in the test example. I would just go for an explicit path of `"stats/{sample}`. Users can change this to whatever they want when using this wrapper---and so can we, if we want this for our pipeline. This would also remove the necessity for the `wildcard_constraints` below.

bio/picard/collectmultiplemetrics/wrapper.py
```diff
@@ -0,0 +1,18 @@
...
+__copyright__ = "Copyright 2020, Antonie Vietor"
+__email__ = "antonie.v@gmx.de"
+__license__ = "MIT"
+
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+
+out = str(snakemake.wildcards.path) + str(snakemake.wildcards.sample)
+
```
dlaehnemann, line 11: Maybe we could add a sensible default for `mem_gb` here, in case it is not defined in the resources. It's good that it is specified in the example, and this should ensure that everybody does define it, but in case someone deletes it, we could catch it here and set it to 3 GB anyway.

bio/picard/collectmultiplemetrics/test/Snakefile
```diff
@@ -0,0 +1,30 @@
...
+                 ".base_distribution_by_cycle.pdf",
+                 ".insert_size_metrics",
+                 ".insert_size_histogram.pdf",
+                 ".quality_by_cycle_metrics",
+                 ".quality_by_cycle.pdf",
+                 ".quality_distribution_metrics",
+                 ".quality_distribution.pdf"
+                 )
+    resources:
+        mem_gb=3
```
dlaehnemann, line 18: Maybe add a comment here, explaining that you can restrict the total resources a pipeline is allowed to use and include this link: <https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#resources>

bio/picard/collectmultiplemetrics/wrapper.py
```diff
@@ -0,0 +1,18 @@
+__author__ = "Antonie Vietor"
+__copyright__ = "Copyright 2020, Antonie Vietor"
+__email__ = "antonie.v@gmx.de"
+__license__ = "MIT"
+
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+
+out = str(snakemake.wildcards.path) + str(snakemake.wildcards.sample)
```
dlaehnemann, line 10: Ah, I had overlooked the setting of `out`, sorry. And I had missed the complexity that CollectMultipleMetrics can produce a number of different outputs. I would think that generally, all metrics files must go into the same directory, so it might be worth noting that in a comment above the `output: ` specification in the Snakefile. And then out can simply be parsed from the first output file:
```
from os import path
out = path.join( path.dirname(snakemake.output[0]), snakemake.wildcards.sample)
```

bio/picard/collectmultiplemetrics/wrapper.py
```diff
@@ -0,0 +1,18 @@
...
+
+shell(
+    "(picard -Xmx{snakemake.resources.mem_gb}g CollectMultipleMetrics "
+    "I={snakemake.input.bam} "
+    "O={out} "
+    "R={snakemake.input.ref} "
+    "{snakemake.params}) {log}"
```
dlaehnemann, line 17: I think we should add in a line with:
```
    "PROGRAM={programs} "
```

This requires a parsing of which output files are requested and matching them to the respective possible outputs. The possible `PROGRAMS` values are given in the docs: https://broadinstitute.github.io/picard/command-line-overview.html#CollectMultipleMetrics

We should thus include something like this above:
```
progs = set()
for file in snakemake.output:
    if "alignment_summary" in file:
        progs.add("CollectAlignmentSummaryMetrics")
    elif ...
    else:
        sys.exit("Unknown type of metrics file requested, for possible metrics files, see https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/picard/collectmultiplemetrics.html")
programs = "PROGRAM=" + PROGRAM=".join(progs)
```

Am I making sense?

dlaehnemann:
Based on our discussion, it would be great to add a full table of possible extensions and the `PROGRAM`s that they imply to the `meta.yaml` description. Maybe this can be done in the form of a simple table:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#tables

bio/picard/collectmultiplemetrics/wrapper.py
dlaehnemann: Maybe also add a check for:
```
if "PROGRAM=" in snakemake.params:
    sys.exit("Please do not specify the PROGRAM= parameter via `params`. The selection of PROGRAMs is handled via extensions in `output`. For available extensions, see: https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/picard/collectmultiplemetrics.html")
```

bio/picard/collectmultiplemetrics/test/Snakefile
dlaehnemann: Add comment above `multiext("stats/{sample}",` that insists on:

* the `{sample}` wildcard being the last thing in the general `multiext` string
* specifying which tools to run via the extensions below
* linking to the snakemake wrapper docs page to find out which extensions can be requested

bio/picard/collectmultiplemetrics/wrapper.py
dlaehnemann: Based upon our discussion: This could be made a bit more general with something like:
```
out = snakemake.output[0].split( snakemake.wildcards.sample )[0]
```

bio/picard/collectmultiplemetrics/test/Snakefile
```diff
@@ -0,0 +1,37 @@
...
+                 ".insert_size_histogram.pdf",
+                 ".quality_by_cycle_metrics",
+                 ".quality_by_cycle.pdf",
+                 ".quality_distribution_metrics",
+                 ".quality_distribution.pdf"
+                 )
+    resources:
+        # this parameter restricts the total resources a pipeline is allowed to use
+        # this value is set to 3 GB by default
+        # For usage information about resources see https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#resources
```
dlaehnemann, line 27: Maybe shorten to:
```
        # This parameter (default 3 GB) can be used to limit the total resources a pipeline is allowed to use, see:
        #     https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#resources
```

PR #121, opened 2020-06-23
GATK BaseRecalibrationSpark

christopher-schroeder:
The spark version of baserecalibration


bio/gatk/baserecalibratorspark/meta.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - bam file
+output:
+  - recalibrated bam file
+notes: |
+  * The `java_opts` param allows for additional arguments to be passed to the java compiler, e.g. "-Xmx4G" for one, and "-Xmx4G -XX:ParallelGCThreads=10" for two options.
+  * The `extra` param allows for additional program arguments for baserecalibratorspark.
+  * The `spark_runner` param = "LOCAL"|"SPARK"|"GCS" allows to set the spark_runner. Set the parameter to "LOCAL" or don't set it at all to run on local machine.
+  * The `spark_master` param allows to set the URL of the Spark Master to submit the job. Set to "local[number_of_cores]" for local execution. Don't set it at all for local execution with number of cores determined by snakemake.
+  * The 'spark_extra' param allows for additional spark arguments.
+  * For more inforamtion see, https://gatk.broadinstitute.org/hc/en-us/articles/360036897372-BaseRecalibratorSpark-BETA-
```
dlaehnemann, line 18: Typo, should be `information`.

Also, maybe make the link clickable in the docs with restructured text, i.e.:
```
  * For more information see the `BaseRecalibratorSpark documentation <https://gatk.broadinstitute.org/hc/en-us/articles/360036897372-BaseRecalibratorSpark-BETA->`_
```

bio/gatk/baserecalibratorspark/meta.yaml
```diff
@@ -0,0 +1,18 @@
+name: gatk BaseRecalibrator
```
dlaehnemann, line 1: Maybe make the name `gatk base recalibration spark` to set it apart from only `BaseRecalibrator` as a lone step and without Spark. I.e.:
```
name: gatk base recalibration (Spark)
```

bio/gatk/baserecalibratorspark/wrapper.py
```diff
@@ -0,0 +1,40 @@
...
+        "-- --spark-runner {spark_runner} --spark-master {spark_master} {spark_extra} "
+        "{log}"
+    )
+
+    log = snakemake.log_fmt_shell(stdout=True, stderr=True, append=True)
+    shell(
+        "gatk --java-options '{java_opts}' ApplyBQSR -R {snakemake.input.ref} -I {snakemake.input.bam} "
+        "--bqsr-recal-file {recal_table} "
+        "-O {snakemake.output.bam} {log}"
+    )
```
dlaehnemann, line 40: Are you sure that you want `BaseRecalibratorSpark` and `ApplyBQSR` in one wrapper? From what I understand, `BaseRecalibratorSpark` can easily parallelise a lot (and this is wanted, as it's slow), whereas `ApplyBQSR` does not parallelize at all (and does not need to, because it's just a linear pass over the data). Wouldn't this lead to snakemake blocking a number of threads for `BaseRecalibratorSpark` of which `ApplyBQSR` will then only use one? Wouldn't it be more efficient, to then have `ApplyBQSR` as a separate wrapper (and simply mark the `BaseRecalibratorSpark` output as `temp()`)? Or am I missing something, here?

christopher-schroeder: > Are you sure that you want `BaseRecalibratorSpark` and `ApplyBQSR` in one wrapper? From what I understand, `BaseRecalibratorSpark` can easily parallelise a lot (and this is wanted, as it's slow), whereas `ApplyBQSR` does not parallelize at all (and does not need to, because it's just a linear pass over the data). Wouldn't this lead to snakemake blocking a number of threads for `BaseRecalibratorSpark` of which `ApplyBQSR` will then only use one? Wouldn't it be more efficient, to then have `ApplyBQSR` as a separate wrapper (and simply mark the `BaseRecalibratorSpark` output as `temp()`)? Or am I missing something, here?

Well I would prefer that, yes. On the other hand I thought it would be a good idea to be as much compatible to the non-spark wrapper as possible. In that way you can easily replace the normal wrapper be the spark version. But now that I rethink about it, it would be nice to have a wrapper "BaseRecalibration" "ApplyRecalibration", "BaseRecalibrationSpark" and one wrapper to have both combined as it is now. But then everybody gets confused.

dlaehnemann: Jep, I would be for general granularity and currently this would simply mean copy-pasting two wrappers. But maybe @johanneskoester can have a look and say what he prefers?

christopher-schroeder: I agree. I think 2 steps are better, as I was also confused at first when I started using this wrapper (because of two steps in one).

bio/gatk/baserecalibratorspark/wrapper.py
```diff
@@ -0,0 +1,40 @@
...
+
+from snakemake.shell import shell
+
+extra = snakemake.params.get("extra", "")
+spark_runner = snakemake.params.get("spark_runner", "LOCAL")
+spark_master = snakemake.params.get(
+    "spark_runner", "local[{}]".format(snakemake.threads)
+)
+spark_extra = snakemake.params.get("spark_extra", "")
+java_opts = snakemake.params.get("java_opts", "")
```
dlaehnemann, line 18: Maybe this could be extended to parse the resources, that you can limit upon snakemake execution. In the Snakefile, you could specify
```
    threads: 4
    resources:
        mem_gb=3
```

And then add these to the `java_opts` here as `"-Xmx{snakemake.resources.mem_gb}G -XX:ParallelGCThreads={snakemake.threads}`. This has the advantage that snakemake is also aware of these resource specifications and can use them for its scheduling, effectively limiting the resource usage.

bio/gatk/baserecalibratorspark/wrapper.py
christopher-schroeder: I could do that, I only need the ok, cause I don't want to do unnecessary work,

bio/gatk/baserecalibratorspark/wrapper.py
christopher-schroeder: I simply don't know enough about the Garbage Collectors to judge if it is a good idea to increase the number of GC threads to the number of spark processes.

christopher-schroeder: I would leave the code as it is, discuss the "mem_gb" idea in general and make an extra pull request where we implement this for all existing java-wrapper in the same way, with the same naming scheme. But I like the idea to make xmx known to snakemake.

dlaehnemann: I've seen the `mem_gb` suggestion from @johanneskoester in reviews of other wrappers, hence the suggestion. And I wasn't aware that there are different levels of threads in this wrapper, so yes, might be best to tread carefully there.

bio/gatk/baserecalibratorspark/wrapper.py
christopher-schroeder: I did it now. I don't want to wait that long and get this from my todo list.

dlaehnemann: Good, @johanneskoester also agrees that this is a good idea, given the parallelization implications!

christopher-schroeder:
I would like to have this closed and implement the mem_gb thing in anseperate pull request

dlaehnemann:
Alright, will have a last look right now. Do you want to address the `mem_gb` issue right away, or should we create a tracking issue for that?

dlaehnemann:
One last question: I see that this PR includes `.dict` files, but they are not required as input to any of the Snakefiles. I would say, if they are required as inputs, they should be included in `input`, so that snakemake will ask for them if not present and the user can then include the respective wrapper to generate them. And if they are not needed, I would drop them...

bio/gatk/baserecalibratorspark/wrapper.py
christopher-schroeder: > Alright, will have a last look right now. Do you want to address the `mem_gb` issue right away, or should we create a tracking issue for that?

Please create an issue

christopher-schroeder:
> Alright, will have a last look right now. Do you want to address the `mem_gb` issue right away, or should we create a tracking issue for that?

Please create an issue

christopher-schroeder:
> One last question: I see that this PR includes `.dict` files, but they are not required as input to any of the Snakefiles. I would say, if they are required as inputs, they should be included in `input`, so that snakemake will ask for them if not present and the user can then include the respective wrapper to generate them. And if they are not needed, I would drop them...

So does the original bqsr wrapper test. I thought they must be mandatory then.

dlaehnemann:
Arggh, the docs don't say. Can you try out for which recipes they are needed and then add them to input there and remove the files from any wrappers where the tool does not require a dict? Sorry to keep on holding this up, but I would say this saves user frustration (including ourselves) in the long run...

dlaehnemann:
And they all need it. How I love undocumented, but required input...

dlaehnemann:
From the test logs, I think that `ApplyBQSR` also needs the `genome.dict` file. Also, I would then add it as a required input file to the rules in the `Snakefile`s, to make snakemake aware of this dependency.

christopher-schroeder:
> From the test logs, I think that `ApplyBQSR` also needs the `genome.dict` file. Also, I would then add it as a required input file to the rules in the `Snakefile`s, to make snakemake aware of this dependency.

yes, I have overseen this at first. Then I searched for a save way to git revert to a specific commit. :-)

bio/gatk/baserecalibratorspark/test/Snakefile
```diff
@@ -0,0 +1,18 @@
...
+    output:
+        recal_table="recal/{sample}.grp"
+    log:
+        "logs/gatk/baserecalibrator/{sample}.log"
+    params:
+        extra="",  # optional
+        java_opts="", # optional
+        #spark_runner="",  # optional
+        #spark_master="",  # optional
+        #spark_extra="", # optional
```
johanneskoester, line 16: When using the local spark runner (default) it is possible to set threads. I think this will be the default for most people, so this example rule should set the threads and explain that the local runner is used by default.

christopher-schroeder: Hm? It already does.

spark_master = snakemake.params.get(
    "spark_master", "local[{}]".format(snakemake.threads)
)

dlaehnemann: I think @johanneskoester meant also specifying the threads value in the rule in the Snakefile, e.g. with `threads: 8` or something, to demonstrate that this is possible...

christopher-schroeder: This is really exhausting and time-consuming. Here you go...

PR #126, opened 2020-06-24
cutadapt 2.10

christopher-schroeder:
Just an update from 2.5 to 2.10 for bugfixes and features.


christopher-schroeder:
pigz as dependency was obsolete, its already a dependency of cutadapt

PR #131, opened 2020-06-29
Optitype - generalized output directory

jafors:
Fix for OptiType wrapper to infer the output dir from the path of the output files.

### DIFF
--- a/bio/optitype/test/Snakefile
+++ b/bio/optitype/test/Snakefile
@@ -13,6 +13,5 @@ rule optitype:
         config="",
         # additional parameters
         extra=""
-    threads: 4
     wrapper:
         "master/bio/optitype"
--- a/bio/optitype/wrapper.py
+++ b/bio/optitype/wrapper.py
@@ -9,7 +9,7 @@ from snakemake.shell import shell
 
 extra = snakemake.params.get("extra", "")
 log = snakemake.log_fmt_shell(stdout=True, stderr=True)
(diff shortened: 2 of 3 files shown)


jafors:
Used this PR to update the test.py and add missing `--cores` parameter to some tests.

jafors:
> Looks good to me, just wondering two things:
> 
>     1. The test Snakefile [lists `threads: 4`](https://github.com/snakemake/snakemake-wrappers/blob/141a12a877197e2332bb1b79a0b084a8517066c4/bio/optitype/test/Snakefile#L16), but the `wrapper.py` doesn't use it. If Optitype is parallelizable, I would use the variable. Otherwise, I would drop it from the Snakefile to avoid confusion.

Thanks, you're right. Probably an artefact, I'll remove it!

>     2. How did the tests pass before, in the master branch, if they didn't contain the `--cores 1`? Or did you base your branch on an older version of master and just needed to catch up? Maybe you could double-check, to make sure we don't generate any merging confusion.

They don't. When you look into the github actions, you'll find that the master branch fails regularly in testing. The PR-branches just don't run every test, it seems (to save time, I guess). I don't know it the missing `--cores` is the only problem, but it should fix some tests.

dlaehnemann:
Ah, OK. Thanks for the explanation! Then it's a good idea to fix those tests, moving towards a passing master branch.

PR #132, opened 2020-06-30
wrapper for bedtools genomeCoverageBed

AntonieV:
with separate tests for bam and bed files


bio/bedtools/genomecov/wrapper.py
```diff
@@ -0,0 +1,29 @@
...
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+
+genome = ""
+input_file = ""
+
+if (os.path.splitext(snakemake.input[0])[-1]) == ".bam":
+    if "-ibam" not in (snakemake.input[0]):
+        input_file = "-ibam " + snakemake.input[0]
```
jafors, line 16: I don't really get the `if`-statement here. Can there be a case where `-ibam` is already in the input?

bio/bedtools/genomecov/wrapper.py
```diff
@@ -0,0 +1,29 @@
...
+
+if len(snakemake.input) > 1:
+    if (os.path.splitext(snakemake.input[0])[-1]) == ".bed":
+        input_file = "-i " + snakemake.input.get("bed")
+        genome = " -g " + snakemake.input.get("ref")
```
jafors, line 21: You'll probably not need the leading whitespace before the `-g` here.

bio/bedtools/genomecov/wrapper.py
AntonieV: Oh right, that should be snakemake.params, thanks for the review.

bio/bedtools/genomecov/wrapper.py
```diff
@@ -12,13 +12,13 @@
 input_file = ""
 
 if (os.path.splitext(snakemake.input[0])[-1]) == ".bam":
-    if "-ibam" not in (snakemake.input[0]):
+    if "-ibam" not in (snakemake.params):
         input_file = "-ibam " + snakemake.input[0]
```
jafors: Ah, I see. But what if the user does sth. like this `params: "-ibam -bg"`?

I would suggest removing the `if`-statement and always adding `-ibam` in front of the bam input. You should be able to trust the user to not use `-ibam` as a standalone param (as explained in the bedtools docs).

AntonieV: That sounds good, thanks.

jafors:
Ok, I think we are good to go then!

PR #134, opened 2020-07-02
Update for Wrapper Picard CollectMultipleMetrics

AntonieV:
@dlaehnemann : An update for picard's CollectMultipleMetrics wrapper. As you suggested the extensions were created as dict in the wrapper and now match to the full extension at the end of a path. The table for assigning tools to their extensions moved to description in meta.yaml file and has received some design changes.

### DIFF
--- a/bio/picard/collectmultiplemetrics/wrapper.py
+++ b/bio/picard/collectmultiplemetrics/wrapper.py
@@ -12,56 +12,45 @@ res = snakemake.resources.get("mem_gb", "3")
 if not res or res is None:
     res = 3
 
+exts_to_prog = {
+    ".alignment_summary_metrics": "CollectAlignmentSummaryMetrics",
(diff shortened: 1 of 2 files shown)


bio/picard/collectmultiplemetrics/wrapper.py
```diff
@@ -12,56 +12,45 @@
...
         )
-programs = " PROGRAM=" + "PROGRAM=".join(progs)
+
+programs = " PROGRAM=" + " PROGRAM=".join(progs)
 
 out = str(snakemake.wildcards.sample)  # as default
 output_file = str(snakemake.output[0])
-for ext in extensions:
+for ext in exts_to_prog:
     if ext in output_file:
```
dlaehnemann, line 54: I think this line is redundant with the following line, so it can be dropped without changing the behaviour.

AntonieV:
Thanks for review :)

PR #135, opened 2020-07-03
wrapper for deepTools computeMatrix

AntonieV:
A new wrapper for deepTools to compute scores per genome regions as a matrix.


bio/deeptools/computematrix/wrapper.py
```diff
@@ -0,0 +1,37 @@
...
+out_tab = snakemake.output.get("matrix_tab")
+out_bed = snakemake.output.get("matrix_bed")
+
+input_files = "-R "
+for bed in bed_files:
+    input_files = input_files + bed + " "
+
+input_files = input_files + "-S "
+for bigwig in bigwig_files:
+    input_files = input_files + bigwig + " "
```
jafors, line 21: While well readable, it seems a bit "too much" for me. Maybe you can shorten it up trying something like:
`"-R {} -S {}".format(" ".join(bed_files), " ".join(bw_files))`.

bio/deeptools/computematrix/wrapper.py
```diff
@@ -0,0 +1,37 @@
...
+
+output_files = "-o " + snakemake.output.get("matrix_gz")
+
+if out_tab and out_tab is not None:
```
jafors, line 25: `If out_tab:` is already good enough here. If out_tab is actually `None` (or an empty string/list) , this will already return `False` and skip the downstream code.

bio/deeptools/computematrix/wrapper.py
```diff
@@ -0,0 +1,37 @@
...
+
+input_files = "-R "
+for bed in bed_files:
+    input_files = input_files + bed + " "
+
+input_files = input_files + "-S "
+for bigwig in bigwig_files:
+    input_files = input_files + bigwig + " "
+
+output_files = "-o " + snakemake.output.get("matrix_gz")
```
johanneskoester, line 23: this concatenatiion can be replaced by directly putting the -o into the shell command below. Same for other parts in this script. Rule of thumb: try to keep as many parameters in the shell command at the bottom itself, only refer to filenames via braces (if possible). This way, readability is maximized.

bio/deeptools/computematrix/wrapper.py
jafors: Ok,looks like I missed the most important thing while talking about formatting ;) It is actually completely fine to use the input directly as sth like `-R {bed_files} -S {bigwig_files}` or `-R {snakemake.input.bed} -S {snakemake.input.bigwig}` in the shell command. This also improves readability, as mentioned above.

AntonieV:
Thanks for review :)

bio/deeptools/computematrix/wrapper.py
```diff
@@ -0,0 +1,29 @@
...
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+out_tab = snakemake.output.get("matrix_tab")
+out_bed = snakemake.output.get("matrix_bed")
+
+optional_output = ""
+
+if out_tab:
+    optional_output = optional_output + " --outFileNameMatrix " + out_tab + " "
```
johanneskoester, line 16:
```suggestion
    optional_output += " --outFileNameMatrix {out_tab} ".format(out_tab=out_tab)
```

bio/deeptools/computematrix/wrapper.py
```diff
@@ -0,0 +1,29 @@
...
+
+if out_bed:
+    optional_output = optional_output + " --outFileSortedRegions " + out_bed + " "
```
johanneskoester, line 19:
```suggestion
    optional_output += " --outFileSortedRegions {out_bed} ".format(out_bed=out_bed)
```

bio/deeptools/computematrix/test/Snakefile
```diff
@@ -0,0 +1,23 @@
+rule compute_matrix:
+    input:
+         # Please note that the -R and -S options are defined via input files
+         bed = expand("{sample}.bed", sample = ["a", "b"]),
+         bigwig = expand("{sample}.bw", sample = ["a", "b"])
```
jafors, line 5: Another really minor thing: As I understand, it is convention in Snakefiles to avoid whitespaces when assigning a key in `input`, `output` or `params`.
That would be `bed=expand("{sample}.bed", sample = ["a", "b"])` and so on.

bio/deeptools/computematrix/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  For more information about ``deepTools``, also see the `source code <https://github.com/deeptools/deepTools>`_.
+
+    +----------------------------+-----------------------------+---------------------+-------------+
+    | computeMatrix option       | Output format               | Name of output      | Recommended |
+    |                            |                             |                     |             |
+    |                            |                             | variable to be used | extension   |
+    +============================+=============================+=====================+=============+
+    | -\-outFileName, -out, -o   | gzipped matrix file, this   | matrix_gz           | ".gz"       |
+    |                            |                             |                     |             |
+    |                            | output file is required     |                     |             |
```
dlaehnemann, line 14: As with the newer `deepTools` wrappers, I'd move the required to the third column and just put `(required)` there for brevity.

bio/deeptools/computematrix/meta.yaml
```diff
@@ -0,0 +1,35 @@
+name: deepTools computeMatrix
+description: >
+  ``DeepTools``'s computeMatrix calculates scores per genome region and can be used as input for other tools or for the generation of a ``deepTools`` plotHeatmap or ``deepTools`` plotProfiles.
+  For usage information about computeMatrix, please see  ``deepTools``'s `documentation <https://deeptools.readthedocs.io/en/develop/content/tools/computeMatrix.html>`_.
+  For more information about ``deepTools``, also see the `source code <https://github.com/deeptools/deepTools>`_.
```
dlaehnemann, line 5: Maybe we can also make this consistent with my suggestions in the newer `deepTools` wrappers.

PR #140, opened 2020-07-08
deepTools plotHeatmap

AntonieV:
A new wrapper for deepTools: plotHeatmap


bio/deeptools/plotheatmap/test/Snakefile
```diff
@@ -0,0 +1,20 @@
+rule plot_heatmap:
+    input:
+         # matrix file from deepTools computeMatrix tool
+         "matrix.gz"
+    output:
+        # Please note that --outFileSortedRegions and --outFileNameMatrix are exclusively defined via output files.
+        # Usable output variables, their extensions and which option they implicitly call are listed here:
+        #         https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/deeptools/plotheatmap.html.
+        # Through the output variables image file and more output options for the heatmap can be selected.
```
dlaehnemann, line 9: :thinking: I don't understand this sentence. Maybe I'm missing something? Could you rephrase it somehow, maybe that would help me understand?

AntonieV: This should only mean that by selecting the different output variables further optional outputs can be created. Actually, it is even unnecessary, because it is more precise in the table, which is linked here. Therefore I removed it.

dlaehnemann: Jep, that works for me! :D

PR #143, opened 2020-07-21
Specified input call to bam field

jafors:
Use the desired input field ("bam") in the shell command instead of the compete `{snakemake.input}`.


dlaehnemann:
Does this work as is with the named input file in the example Snakefile? Because then keeping it as it is would also allow using an unnamed input with the wrapper. This would keep this more flexible, while the specificity is not necessarily needed to disambiguate anything here, right?

PR #145, opened 2020-07-23
add bcftools mpileup wrapper

mbhall88:
In addition to adding this wrapper, I would also like to suggest changing the `bcftools call` wrapper to stop using `samtools mpileup` (which is deprecated) and instead only make that wrapper run `bcftools call`. I know the bcftools docs give the example of piping the output of `mpileup` into `call`, but `mpileup` can take a long time to run so I prefer having it run separately so I have that intermediate file on hand in case `call` needs to be rerun. If others don't want to keep that `mpileup` file lying around they can just tag it with [`temp()`](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#protected-and-temporary-files) obviously.

### DIFF
--- a/bio/bcftools/mpileup/wrapper.py
+++ b/bio/bcftools/mpileup/wrapper.py
@@ -23,9 +23,8 @@ if "--no-reference" not in options:
         )
     options += " --fasta-ref {}".format(ref)
 
-options += " --threads {}".format(snakemake.threads)
-
 shell(
-    "bcftools mpileup {options} --output {snakemake.output.pileup} "
+    "bcftools mpileup {options} --threads {snakemake.threads} "
+    "--output {snakemake.output.pileup} "
     "{snakemake.input.alignments} 2> {snakemake.log}"
 )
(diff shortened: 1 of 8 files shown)


bio/bcftools/mpileup/wrapper.py
```diff
@@ -0,0 +1,31 @@
...
+if "--no-reference" not in options:
+    ref = snakemake.input.get("ref", "")
+    if not ref:
+        raise MissingReferenceError(
+            "The --no-reference option was not given, but no fasta reference was "
+            "provided."
+        )
+    options += " --fasta-ref {}".format(ref)
+
+options += " --threads {}".format(snakemake.threads)
```
jafors, line 26: Maybe use `--threads {snakemake.threads}` in the shell command instead.
For maximal readability, it would probably be nicer to use as many parameters as possible directly in the shell script instead of concatenating everything into `options`.

mbhall88: Yeah, fair point. I went a bit overboard in my quest for succinctness :sweat_smile:

mbhall88:
@jafors  requested changes are made.

PR #151, opened 2020-07-31
Find peaks fix

jafors:
Fixed an error in findPeaks to allow correct usage of `-style`.


jafors:
> The fix looks good. Wanna maybe add a link to the homer docs to the `meta.yaml`, here (the docs are referenced in the params comment, so maybe a link should be somewhere close-by in the wrapper docs -- and links to docs are good practice, anyway, I'd say):

I'll use this PR to update all HOMER docs. Good point! ;)

Issue #150, opened 2020-07-30
Modularity of wrappers?


vsoch:
In the same spirit of having https://github.com/snakemake-workflows, is there any concern here about maintaining versions / containers for wrappers? I just cloned the repository locally and it wasn't tiny, but my larger concern is that if we also serve containers from here (that have automated builds for the wrappers) that will be more complicated than having a https://github.com/snakemake-wrappers organization, and storing wrappers under the complete uri. E.g., the current bio/arriba would be at https://github.com/snakemake-wrappers/bio/arriba. If some web interface (or similar) is then desired to show the wrappers, it could be rendered at snakemake-wrappers.github.io, and use a simple query to the API to update the listing.

PR #152, opened 2020-08-06
add rasusa wrapper

mbhall88:
This PR will eventually try and add a `rasusa` wrapper (it currently works). But it is not to be merged yet as it relates to work in https://github.com/snakemake/snakemake/pull/532 and #150 

Note: I also reordered some of the tests to try and maintain alphabetical ordering. Not sure if this is being adhered to anymore?


vsoch:
This looks great! We will want to update for any possible changes after discussion in https://github.com/snakemake/snakemake-wrappers/pull/153. For example, if we are able to build the container from the environment.yaml, you might not need to provide it. It's also not clear if we would want `container: Dockerfile`, leaving it out entirely, or something else.

mbhall88:
Just wondering if this PR is required for the container stuff anymore? I think you had a different way of testing this out didn't you @vsoch ?

vsoch:
I had a PR in to snakemake (which would be required here) but it looks like it's gone stale. I'm not sure if this work is still a goal for snakemake, or some other strategy is going to be used. Probably we should ask @johanneskoester.

vsoch:
But @mbhall88 you can probably remove the container stuff and contribute as a wrapper without it! Don't let our slowness / this feature development hold you back! :)

dlaehnemann (approved):
The recipe looks good. Just two questions:

1. There's a lot of formatting going on in `test.py`. Are you sure you are using the latest `black`? Just want to avoid unnecessary changes, in case this will be reverted with the next best merge...
2. Just out of curiosity: What can rasusa do for me, that [`samtools view -s`](https://www.htslib.org/doc/samtools-view.html) cannot.

mbhall88:
> The recipe looks good. Just two questions:
> 
>     1. There's a lot of formatting going on in `test.py`. Are you sure you are using the latest `black`? Just want to avoid unnecessary changes, in case this will be reverted with the next best merge...

I'm using `black, version 20.8b1` which seems to be the latest? Anyway, would the CI checks have detected if the formatting was incorrect?

> 
>     2. Just out of curiosity: What can rasusa do for me, that [`samtools view -s`](https://www.htslib.org/doc/samtools-view.html) cannot.

The difference is that samtools/seqtk only subsample based on the *number* of reads - whereas rasusa subsamples to a given (theoretical) *coverage*. Which is sort of fine if you have Illumina data and you know all your reads are the same length, and you can be bothered to calculate how many you would need to get to a certain coverage. But this breaks down when working with long reads where they are never all the same length.  

I'm hoping to write this up as an application note in *Bioinformatics* or something similar soon.
I talk about this a little in the [benchmark section](https://github.com/mbhall88/rasusa#paired-end-input).

dlaehnemann:
The [GitHub Action does run a `black check`](https://github.com/snakemake/snakemake-wrappers/blob/a53e5257f470aea1a917ab02a4fc75f9810b2f6f/.github/workflows/main.yml#L25) (not sure which version, would have to check the output of the GitHub Action run on this branch, I guess). And this PR passes, so it should all be fine and I'll merge.

And thanks for the elaboration on rasusa, this does sound good. It could even be used to take peaks out of crazily over-covered regions while leaving everything else alone. This could avoid slowing down downstream tools with excessive local coverage. Definitely worth a write-up!

mbhall88:
>  It could even be used to take peaks out of crazily over-covered regions while leaving everything else alone. 

It won't really do this. For that, you would need an alignment file and to use something like VariantBam (see [here](https://github.com/mbhall88/rasusa/issues/17#issuecomment-708165373) for more info).

PR #155, opened 2020-08-17
Macs2 callpeak

AntonieV:
A new wrapper for the macs2 callpeak tool.


bio/macs2/callpeak/wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+    "_peaks.broadPeak",
+    "_peaks.gappedPeak",
+    "_treat_pileup.bdg",
+    "_control_lambda.bdg",
+}
+
+for ext in exts:
+    if out_file.endswith(ext):
+        out_name = out_file[: -len(ext)]
+        out_dir = os.path.dirname(out_file)
```
jafors, line 31: I think you could just use the required `_peaks.xls` extension here.
Sth like:
`out_file = [o for o in snakemake.output if o.endswith("_peaks.xls")][0]`
That should save you from trying all possible extensions.

bio/macs2/callpeak/wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+        list(
+            map(os.unlink, (os.path.join(out_dir, i) for i in os.listdir(out_dir)))
+        )  # removes old result files
+        os.makedirs(out_dir, exist_ok=True)
```
jafors, line 35: Is this necessary? Does `macs callpeak` not allow to overwrite already existing output files of the same name?
And can't you create the output directory by using `--outdir {out_dir}` in the shell-command?

If I am wrong here, please tell me :)

bio/macs2/callpeak/wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+        break
+
+if in_contr:
+    opt_input = " -c {contr}".format(contr=in_contr)
+
+if " --broad" in params:
+    for out in snakemake.output:
+        if out.endswith("_peaks.narrowPeak") or out.endswith("_summits.bed"):
```
jafors, line 43: Syntactic sugar (similar for the other cases):
`if any(out.endswith(("_peaks.narrowPeak", "_summits.bed")) for out in snakemake.output):`

bio/macs2/callpeak/wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+                "Add --broad option to params if these files are needed.\n"
+            )
+
+for out in snakemake.output:
+    if out.endswith("_treat_pileup.bdg") or out.endswith("_control_lambda.bdg"):
+        if (
+            "--bdg" not in params
+            and " -B " not in params
+            and not params.endswith(" -B")
+        ):
```
jafors, line 62: Similar to above, this can be written as:

`if all(p not in params for p in ["--bdg", "-B"]):`

Python is case-sensitive here, `-B` will not be falsely matched against a lowercase b as in e.g. `--broad` if that was your concern.

dlaehnemann: Also, it should be OK to have `-B` both at the start or at the end of the `params`, right? So no whitespace should be needed in the matching and the `not params.endswith(" - B")` should not be necessary? Or am I missing something?

bio/macs2/callpeak/test/Snakefile
```diff
@@ -0,0 +1,27 @@
...
+                 "_peaks.broadPeak",
+                 "_peaks.gappedPeak",
+                 # if --broad is not set in params:
+                 # "_peaks.narrowPeak",
+                 # "_summits.bed"
+                 )
+    log:
+        "logs/macs2/callpeak.log"
+    params:
+        "-f BAM -g hs --nomodel --bdg --broad"
```
dlaehnemann, line 25: Looking at the great table you provide in the docs, I hope that I understand correctly, that certain command-line parameters lead to certain (sometimes mutually exclusive) output, right? I think the optimal solution is then to only allow one way of specifying output, either via the parameters, or via the extensions. From the Snakemake point-of-view, we want to know which output to expect, so I would opt for specifying the wanted extensions and inferring the command-line flags from them. This means that we will have to:

* [ ] check for incompatible extension combinations and throw an error if encountered (basically what you are already doing, just with adjusted wording reflecting that the specified extensions are the problem)
* [ ] provide two example rules and test each of them separately, e.g. one that requests the extensions that imply `--bdg` and `--broad` and one without the extensions implying `--bdg` and requesting the extensions produced when no `--broad` flag is set.

bio/macs2/callpeak/environment.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
+  - bioconda
+  - defaults
+dependencies:
+  - macs2 ==2.2.7.1
```
dlaehnemann, line 5: The continuous integration error comes from conda being unable to resolve this environment. Maybe conda chokes on the long version number, which is one entry longer than semantic versioning allows, so maybe try (it's just a hunch, but who knows...;):
```suggestion
  - macs2 ==2.2.7
```

bio/macs2/callpeak/wrapper.py
dlaehnemann: Also, I had another look above and the suggestion about handling the output specification via output extensions should make these checks here simpler -- basically checking that none of the output-specification flags are set in params, because we do this implicitly by parsing the extensions.

AntonieV:
Thanks for the code reviews and the great suggestions @jafors and @dlaehnemann.

bio/macs2/callpeak/wrapper.py
AntonieV: The --outdir command is a great idea, but I would suggest keeping the removal of old result files. Files with the same extension will be overwritten, but files from previous runs will be preserved if the --bdg or --broad options are changed.

dlaehnemann: Hmm, I think we can let Snakemake handle the deletion of old files, as we will now enforce setting of outputs via the requested output extensions. If you want different output, you have to actually change the rule. Thus, optimally, a user would have separate rules for separate sets of output files and Snakemake can again handle files separately for each rule. Of course, this leaves some space for confusion when a user edits a rule and forgets to delete old output. But I would regard this as a corner case we don't need to handle in the wrapper. What do you think?

bio/macs2/callpeak/wrapper.py
```diff
@@ -0,0 +1,66 @@
...
+            "For usable extensions please see https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/macs2/callpeak.html.\n"
+        )
+    else:
+        params = params.replace("--broad", "")
+
+if any(
+    out.endswith(("_peaks.broadPeak", "_peaks.gappedPeak")) for out in snakemake.output
+):
+    if "--broad" not in params:
+        params += " {} ".format("--broad")
```
dlaehnemann, line 47: I think this would be slightly more succinct without the `.format()`:
```suggestion
        params += " --broad "
```

bio/macs2/callpeak/wrapper.py
```diff
@@ -0,0 +1,66 @@
...
+
+if any(
+    out.endswith(("_treat_pileup.bdg", "_control_lambda.bdg"))
+    for out in snakemake.output
+):
+    if all(p not in params for p in ["--bdg", "-B"]):
+        params += " {} ".format("--bdg")
```
dlaehnemann, line 54: As above, I'd not use `.format()` just to add whitespace:
```suggestion
        params += " --bdg "
```

bio/macs2/callpeak/wrapper.py
```diff
@@ -0,0 +1,66 @@
...
+    if any(
+        out.endswith(("_peaks.broadPeak", "_peaks.gappedPeak"))
+        for out in snakemake.output
+    ):
+        sys.exit(
+            "Output files with _peaks.narrowPeak and/or _summits.bed extensions cannot be created together with _peaks.broadPeak and/or _peaks.gappedPeak extended output files.\n"
+            "For usable extensions please see https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/macs2/callpeak.html.\n"
+        )
+    else:
+        params = params.replace("--broad", "")
```
dlaehnemann, line 41: Hmm, I think this should probably be a check for the presence of `--broad`, throwing an error if positive. While the current version will simply work in more instances, I would prefer for the wrapper to enforce a clear usage. Or does anything speak against this?

bio/macs2/callpeak/wrapper.py
```diff
@@ -0,0 +1,66 @@
...
+
+if any(
+    out.endswith(("_treat_pileup.bdg", "_control_lambda.bdg"))
+    for out in snakemake.output
+):
+    if all(p not in params for p in ["--bdg", "-B"]):
+        params += " {} ".format("--bdg")
+else:
+    for i in ["--bdg", "-B"]:
+        params = params.replace(i, "")
```
dlaehnemann, line 57: As above, I would make this a check for the presence of either flag and throw an informative error if encountered.

bio/macs2/callpeak/meta.yaml
```diff
@@ -0,0 +1,48 @@
+
+name: macs2 callpeak
+description: >
+  ``MACS2 callpeak`` model-based analysis tool for ChIP-sequencing that calls peaks from alignment results.
+  For usage information about ``MACS2 callpeak``, please see the `documentation <https://github.com/macs3-project/MACS/blob/master/README.md>`_ and the command line help.
+  For more information about ``MACS2``, also see the `source code <https://github.com/macs3-project/MACS>`_ and `published article <https://genomebiology.biomedcentral.com/articles/10.1186/gb-2008-9-9-r137>`_.
+  Please note that the number of output files and their extension depends on the selected options (please see table below).
```
dlaehnemann, line 7: Maybe adjust this line to reflect that requested extensions will invoke the flags / options automatically and that note that some extensions are incompatible with each other?

bio/macs2/callpeak/environment.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
+  - bioconda
+  - defaults
```
dlaehnemann, line 3: Macs2 depends on a number of `conda-forge` packagess, so I think this should be part of the mix, here:
```suggestion
  - conda-forge
  - bioconda
  - defaults
```

dlaehnemann (approved):

jafors (approved):
Fine with me!

PR #156, opened 2020-08-17
Gridss

christopher-schroeder:
Four Gridss wrapper to setup the reference, preprocess, assemble and call breakends.
Intermediate files are produced within an working directory with a special structure, which cannot be altered.
All tests are working by now, but the test-workflows, which are used as examples for the documentation are not perfect, any suggestion to improve them is highly appreciated!


bio/gridss/assemble/wrapper.py
```diff
@@ -0,0 +1,45 @@
...
+    raise ValueError("Please set params.workingdir to provide a working directory.")
+
+if not snakemake.input.reference:
+    raise ValueError("Please set input.reference to provide reference genome.")
+
+for ending in (".amb", ".ann", ".bwt", ".pac", ".sa"):
+    if not path.exists(f"{reference}{ending}"):
+        raise ValueError(
+            f"{reference}{ending} missing. Please make sure the reference was properly indexed by bwa."
+        )
```
jafors, line 30: I'm not sure if we should use f-strings. Yes, Python 3.6 is old news, but I remember Johannes making a point that we want to be as backwards compatible as possible.

bio/gridss/setupreference/test/Snakefile
```diff
@@ -0,0 +1,11 @@
...
+    input:
+        reference="reference/genome.fasta",
+        dictionary="reference/genome.dict",
+        indices=multiext("reference/genome.fasta", ".amb", ".ann", ".bwt", ".pac", ".sa")
+    output:
+        multiext("reference/genome.fasta", ".gridsscache", ".img")
+    log:
+        "log/gridss/setupreference.log"
+    wrapper:
+        "file:/vol/huge/christo/snakemake-wrappers/bio/gridss/setupreference"
```
jafors, line 11: This should probably be
```
    wrapper:
        "master/bio/gridss/setupreference"
```
Right?

bio/gridss/call/wrapper.py
```diff
@@ -0,0 +1,47 @@
...
+            f"{reference}{ending} missing. Please make sure the reference was properly indexed by bwa."
+        )
+
+if not path.exists(f"{dictionary}"):
+    raise ValueError(
+        f"{reference}{ending} missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard"
+    )
+
+shell(
+    "(export JAVA_OPTS='-XX:ActiveProcessorCount=8' & "
```
jafors, line 38: Should the Java opts be chooseable as a parameter or is this specific option simply the ideal choice and should stay fixed?

jafors (changes requested):
That's a lot of wrappers and checks. Thanks!

Some first thoughts below!

bio/gridss/assemble/wrapper.py
christopher-schroeder: you are right, I will replace them

bio/gridss/call/wrapper.py
christopher-schroeder: Well simply: I don't know. I set this to {threads} but it might be that we have to include a cpu limiter for real data.

jafors (approved):

PR #159, opened 2020-08-20
Strling

christopher-schroeder:
Strling wrapper


bio/strling/call/test/Snakefile
```diff
@@ -0,0 +1,18 @@
...
+        bin="extract/{sample}.bin",
+        reference="reference/genome.fasta",
+        fai="reference/genome.fasta.fai",
+        bounds="merged/group-bounds.txt" # optional, produced by strling merge
+    output:
+        "call/{sample}-bounds.txt",
+        "call/{sample}-genotype.txt",
+        "call/{sample}-unplaced.txt"
+    params:
+        prefix="call/{sample}"
```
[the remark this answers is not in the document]
dlaehnemann: Ah, now I see that the `bounds` file is probably optional output, depending on the `bounds=` named input parameter? Then maybe work with named output parameters, as well.

bio/strling/extract/meta.yaml
```diff
@@ -0,0 +1,4 @@
+name: "strling extract"
+description: Extract informative read pairs to a binary format for a single sample (same as above, you can use the same bin files).
```
dlaehnemann, line 2: As above, I'd add a link to the docs and a link to the repo. Maybe also mention the general purpose of sterling (calling STRs), so someone browsing the wrappers knows what this is for, in general.

bio/strling/call/wrapper.py
```diff
@@ -0,0 +1,48 @@
...
+if not reference:
+    raise ValueError("Please provide a fasta 'reference' input.")
+
+if not bounds:  # optional
+    bounds_string = ""
+else:
+    bounds_string = f"-b {bounds}"
+
+if not path.exists(reference + ".fai"):
+    raise ValueError("Please index the reference and generate the *.bai file")
```
dlaehnemann, line 39: I'd use similar wording as with the `.bam.bai` file to make it clear that the same basename is necessary for both files.

bio/strling/extract/wrapper.py
```diff
@@ -0,0 +1,46 @@
...
+extra = snakemake.params.get("extra", "")
+
+# Check inputs/arguments.
+bam = snakemake.input.get("bam", None)
+reference = snakemake.input.get("reference", None)
+index = snakemake.input.get("index", None)
+
+if not bam:
+    raise ValueError("Please provide a 'bam' input.")
+
```
dlaehnemann, line 24: Shouldn't this wrapper also have the `.bam.bai` check?

bio/strling/extract/wrapper.py
```diff
@@ -0,0 +1,46 @@
...
+if not reference:
+    raise ValueError("Please provide a fasta 'reference' input.")
+
+if not path.exists(reference + ".fai"):
+    raise ValueError("Please index the reference and generate the *.fai file")
```
dlaehnemann, line 29: As above, I'd use wording  to make it clear that the same basename is necessary for both files.

bio/strling/extract/test/Snakefile
```diff
@@ -0,0 +1,13 @@
+rule strling_extract:
+    input:
+        bam="mapped/{sample}.bam",
+        bai="mapped/{sample}.bam.bai",
+        reference="reference/genome.fasta",
+        fai="reference/genome.fasta.fai",
+        index="reference/genome.fasta.str" # optional
+    output:
+        "extract/{sample}.bin"
+    log:
```
dlaehnemann, line 10: Maybe add an empty string as `params.extra` here, to demonstrate that you could add your own extra parameters?

```suggestion
   params:
       extra="" # optionally add further command line arguments
   log:
```

bio/strling/index/meta.yaml
```diff
@@ -0,0 +1,4 @@
+name: "strling index"
+description: Creates a bed file of large STR regions in the reference genome. This step is peformed automatically as part of strling extract. However, when running multiple samples, it is more efficient to do it once, then pass the file to strling extract using the -g option.
```
dlaehnemann, line 2: As above, I'd add a link to the documentation and the repo, here.

bio/strling/merge/meta.yaml
```diff
@@ -0,0 +1,4 @@
+name: "strling merge"
+description: Prepare joint call str loci across all samples. Requires minimum read evidence from at least one sample.
```
dlaehnemann, line 2: As above, I'd add a link to the docs and to the repo. And a note about what strling does, in general

bio/strling/merge/test/Snakefile
```diff
@@ -0,0 +1,13 @@
+rule strling_merge:
+    input:
+        bins=["extract/A.bin", "extract/B.bin"],
+        reference="reference/genome.fasta",
+        fai="reference/genome.fasta.fai",
+    output:
+        "merged/group-bounds.txt"
+    params:
+        prefix="merged/group"
```
dlaehnemann, line 9: As above, can't we just infer the prefix from the output file and its fixed suffix? That way, we don't run the risk of having a user accidentally specify something different in `output` and `params`.

bio/strling/merge/wrapper.py
```diff
@@ -0,0 +1,35 @@
...
+prefix = snakemake.params.get("prefix", None)
+
+if not bins or len(bins) < 2:
+    raise ValueError("Please provide at least two 'bins' as input.")
+
+if not reference:
+    raise ValueError("Please provide a fasta 'reference' input.")
+
+if not path.exists(reference + ".fai"):
+    raise ValueError("Please index the reference and generate the *.fai file")
```
dlaehnemann, line 30: As above, I'd adjust the wording to make it clear, that we want identical basenames in `.fa` and `.fa.fai`.

dlaehnemann (changes requested):
Nice recipes with lots of great sanity checks! Thanks!

I just have a bunch of mostly cosmetic suggestions, nothing that should take much time to address.

bio/strling/call/wrapper.py
```diff
@@ -0,0 +1,48 @@
...
+extra = snakemake.params.get("extra", "")
+
+# Check inputs/arguments.
+bam = snakemake.input.get("bam", None)
+bin = snakemake.input.get("bin", None)
+reference = snakemake.input.get("reference", None)
+bounds = snakemake.input.get("bounds", None)
+prefix = snakemake.params.get("prefix", None)
+
+if not bam:
```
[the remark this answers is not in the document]
christopher-schroeder: Well yes, you could do that. But then you could argue that you also have to check the prefix param to be not a list. I will add this check, but in general: should we really check for everything?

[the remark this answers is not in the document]
dlaehnemann: Well, I would argue that any easy check that gives a more descriptive error than snakemake would by default, makes sense. But I also think that a good default Snakefile example is more important for the wrappers than checking every little possibility...

bio/strling/call/test/Snakefile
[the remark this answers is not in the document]
christopher-schroeder: I did it and test for equal prefix. And no, bounds is always generated.

bio/strling/call/test/Snakefile
```diff
@@ -0,0 +1,18 @@
...
+        bin="extract/{sample}.bin",
+        reference="reference/genome.fasta",
+        fai="reference/genome.fasta.fai",
+        bounds="merged/group-bounds.txt" # optional, produced by strling merge
+    output:
+        "call/{sample}-bounds.txt", # must end with -bounds.txt
+        "call/{sample}-genotype.txt", # must end with -genotype.txt
+        "call/{sample}-unplaced.txt" # must end with -unplaced.txt
+    params:
+        prefix="call/{sample}"
```
dlaehnemann, line 14: This is not used any more, right?
```suggestion
```

bio/strling/extract/meta.yaml
```diff
@@ -0,0 +1,4 @@
+name: "strling extract"
+description: STRling (pronounced like “sterling”) is a method to detect large STR expansions from short-read sequencing data. "Extract" retrieves informative read pairs to a binary format for a single sample (same as above, you can use the same bin files). (https://strling.readthedocs.io/en/latest/run.html)
```
dlaehnemann, line 2: I'd skip the `like` and put every sentence on a separate line, for cleaner version control.
```suggestion
description: >
STRling (pronounced “sterling”) is a method to detect large STR expansions from short-read sequencing data.
``extract`` retrieves informative read pairs to a binary format for a single sample (same as above, you can use the same bin files).
Documentation at: https://strling.readthedocs.io/en/latest/run.html
```

bio/strling/index/meta.yaml
```diff
@@ -0,0 +1,4 @@
+name: "strling index"
+description: STRling (pronounced like “sterling”) is a method to detect large STR expansions from short-read sequencing data. "Index" creates a bed file of large STR regions in the reference genome. This step is peformed automatically as part of strling extract. However, when running multiple samples, it is more efficient to do it once, then pass the file to strling extract using the -g option. (https://strling.readthedocs.io/en/latest/run.html)
```
dlaehnemann, line 2: As above, I'd skip the `like` and put every sentence on a separate line for cleaner version control. And I'd add what the acronym STR actually stands for:
```suggestion
description: >
    STRling (pronounced like “sterling”) is a method to detect large short tandem repeat (STR) expansions from short-read sequencing data.
    ``index`` creates a bed file of large STR regions in the reference genome.
    This step is performed automatically as part of ``strling extract``.
    However, when running multiple samples, it is more efficient to do it once, then pass the file to strling extract using the ``-g`` option.
    Documentation at: https://strling.readthedocs.io/en/latest/run.html
```

bio/strling/merge/meta.yaml
```diff
@@ -0,0 +1,4 @@
+name: "strling merge"
+description: STRling (pronounced like “sterling”) is a method to detect large STR expansions from short-read sequencing data. "Merge" prepares joint call str loci across all given samples. Requires minimum read evidence from at least one sample. (https://strling.readthedocs.io/en/latest/run.html)
```
dlaehnemann, line 2: As above, I'd suggest:
```suggestion
description: >
    STRling (pronounced “sterling”) is a method to detect large short tandem repeat (STR) expansions from short-read sequencing data.
    ``merge`` prepares joint calling of STR loci across all given samples.
    Requires minimum read evidence from at least one sample.
    Documentation at: https://strling.readthedocs.io/en/latest/run.html
```

bio/strling/call/meta.yaml
```diff
@@ -0,0 +1,4 @@
+name: "strling call"
+description: STRling (pronounced like “sterling”) is a method to detect large STR expansions from short-read sequencing data. "Call" calls genotypes/estimate allele sizes for all loci in each sample. (https://strling.readthedocs.io/en/latest/run.html)
```
dlaehnemann, line 2: Ah, forgot the suggestion here:
```suggestion
description: >
    STRling (pronounced like “sterling”) is a method to detect large short tandem repeat (STR) expansions from short-read sequencing data.
    ``call`` calls genotypes/estimate allele sizes for all loci in each sample.
    Documentation at: https://strling.readthedocs.io/en/latest/run.html
```

bio/strling/call/wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+        "Please index the bam file. The index file must have same file name as the bam file, with '.bai' appended."
+    )
+
+if not reference:
+    raise ValueError("Please provide a fasta 'reference' input.")
+
+if not bounds:  # optional
+    bounds_string = ""
+else:
+    bounds_string = f"-b {bounds}"
```
dlaehnemann, line 37:
```suggestion
    bounds_string = "-b {}".format(bounds)
```

bio/strling/call/wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+
+if not any(o.endswith("-bounds.txt") for o in snakemake.output):
+    raise ValueError("Please provide a file that ends with -bounds.txt in the output.")
+
+for filename in snakemake.output:
+    if filename.endswith("-bounds.txt"):
+        prefix = filename[: -len("-bounds.txt")]
+        break
+
+if not any(o == f"{prefix}-genotype.txt" for o in snakemake.output):
```
dlaehnemann, line 52:
```suggestion
if not any(o == "{}-genotype.txt".format(prefix) for o in snakemake.output):
```

bio/strling/call/wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+    raise ValueError(
+        "Please provide an output file that ends with -genotype.txt and has the same prefix as -bounds.txt"
+    )
+
+if not any(o == f"{prefix}-unplaced.txt" for o in snakemake.output):
```
dlaehnemann, line 57:
```suggestion
if not any(o == "{}-unplaced.txt".format(prefix) for o in snakemake.output):
```

bio/strling/extract/wrapper.py
```diff
@@ -0,0 +1,53 @@
...
+
+if not path.exists(reference + ".fai"):
+    raise ValueError(
+        "Please index the reference. The index file must have same file name as the reference file, with '.fai' appended."
+    )
+
+if not index:  # optional
+    index_string = ""
+else:
+    index_string = f"-g {index}"
```
dlaehnemann, line 41:
```suggestion
    index_string = "-g {}".format(index)
```

PR #162, opened 2020-08-25
Add wrapper for EggNOGmapper 

SilasK:
Hey, this is my first or second time I try to contribute a wrapper. 
It is for the annotation tool EggNOGmapper. 

I wonder how I should test the wrapper with CI, because the tool needs big databases to function.

I also would need to finish the following wrappers before the final commit. 

Todo:
    - [ ] add a wrapper for database download
    - [ ] add a wrapper for eggNOGmapper homology_search step


SilasK:
FYI @jhcepas

johanneskoester:
That's a hard one with testing then. Maybe it is possible to have a small dummy database stored in the repo? If not how big? Can we download it on the fly while testing?

SilasK:
The eggnog.db is 37G SQLite 3.x database. 
This excludes probably to download it for testing. I think about making a rule for the downloading of the databases.

johanneskoester:
Yes, similar to the vep-cache and vep-plugins wrappers, you should also add a wrapper for downloading the database (with `cache: True`). We won't be able to really test them then. But at least we can test for syntactical correctness with dryrun.

dlaehnemann:
@SilasK we're doing a small `snakemake-wrappers` cleanup today, and I was wondering if there are any further plans on this one (and the mentioned TODOs)?

johanneskoester:
Closing for now due to inactivity. Please reopen if you have further plans to finalize this.

PR #169, opened 2020-09-01
Enable meta wrapper

jafors:
Test and doc additions for subworkflows that consist of multiple tool wrappers - currently called "meta-wrappers" (maybe "subworkflow" would even be a better choice?).

There will probably be changes and restructuring, this is just a starting point.


test.py
```diff
@@ -78,6 +79,83 @@ def run(wrapper, cmd, check_log=None):
             os.chdir(origdir)
 
 
+def run_meta(wrapper, cmd, check_log=None):
```
johanneskoester, line 82: This looks quite redundant compared to run(). Can you generalize the original run and reuse it instead?

johanneskoester (changes requested):
Awesome! Just one thing.

johanneskoester (approved):

Issue #176, opened 2020-09-11
Wrappers using unnamed params break cluster params


lkuchenb:
Some wrappers, e.g. `samtools sort` or `samtools index` consume params entirely to customize tool invocation. Others, like `trim_galore-pe` use a named param called `extra`.

The latter allows the specification of additional params for cluster execution as suggested [in the documentation](https://snakemake.readthedocs.io/en/stable/executing/cluster.html), the former breaks when additional named params are used. I'm not sure if there are use cases for additional params other than `--cluster`, as they are obviously not going to be used in the rule code itself.

lkuchenb:
@jafors Ah good. That one didn't come up in my issue searches ;-) I'll close this one then.

johanneskoester:
@lkuchenb that part of the docs was a bit outdated, sorry. Such properties should now rather be specified via `resources` instead of `params`. Params is really only for stuff that is platform independent. Hence, a missing `extra` should not hurt.

johanneskoester:
I have updated the docs, changes will show up in the next release.

lkuchenb:
Nice, that's good to know. Thanks.

PR #177, opened 2020-09-14
Add extra params

jafors:
Adding "extra" field in params for some wrappers - there are probably more to do.

Referencing #176


johanneskoester:
Please see my answer in #176. I think we should keep it as it is now, in order to avoid such a breaking change, since params should not anymore be used for cluster specific information.

PR #182, opened 2020-09-15
Bump OptiType version to 1.3.5

lkuchenb:
Until version 1.3.4, OptiType writes temporary BAM files and names them after the current date and time. In combination with a workflow engine that potentially spawns multiple optitype instances within one second, this causes name collisions.

This was fixed in 1.3.5, the temporary BAM file names now take `--prefix` into account.

### DIFF
--- a/bio/optitype/environment.yaml
+++ b/bio/optitype/environment.yaml
@@ -3,4 +3,4 @@ channels:
   - bioconda
   - defaults
 dependencies:
-  - optitype ==1.3.4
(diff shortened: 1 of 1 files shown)


jafors (commented):
Thank you for contributing! I updated the CI workflow, could you pull the current change in the master branch and update the PR, then we'll see if it passes the tests afterwards. (Just a formaility, there should be nothing wrong with a simple version udate)

jafors (approved):

PR #185, opened 2020-09-17
Update ensembl sequence

AntonieV:
Update for Reference Ensembl-Sequence Wrapper to select a single chromosome


bio/reference/ensembl-sequence/test/old_release.smk
```diff
@@ -10,3 +10,17 @@ rule get_genome:
         "logs/get_genome.log"
     wrapper:
```
dlaehnemann, line 11: Just for consistency of the examples, I'll add in the `cache: True` as well:
```suggestion
    cache: True  # save space and time with between workflow caching (see docs)
    wrapper:
```

bio/reference/ensembl-sequence/test/old_release.smk
```diff
@@ -10,3 +10,17 @@ rule get_genome:
...
+        "refs/old_release.chr1.fasta"
+    params:
+        species="saccharomyces_cerevisiae",
+        datatype="dna",
+        build="R64-1-1",
+        release="101",
+        chromosome="I"
+    log:
+        "logs/get_genome.log"
+    wrapper:
```
dlaehnemann, line 25: As above, I'll add in the `cache: True`:
```suggestion
    cache: True  # save space and time with between workflow caching (see docs)
    wrapper:
```

dlaehnemann (approved):
Thanks a bunch, this looks good!

PR #186, opened 2020-09-18
Added wrappers for Picard MarkDuplicatesWithMateCigar, mapDamage2, bcftools sort, bwa sampe/se, and samtools calmd

fgvieira:
Implementation of wrappers for `bcftools sort`, `samtools calmd`, `Picard MarkDuplicatesWithMateCigar`, `mapDamage2`, as well as a generic wrapper for both `bwa` `samse` and `sampe` (issue #215).

`Picard MarkDuplicatesWithMateCigar` wrapper was copied from `Picard MarkDuplicates` and, since the latter had no test (and I was not sure why), I did not include a test in this one either.

More info about `mapDamage2`:
https://github.com/ginolhac/mapDamage
https://ginolhac.github.io/mapDamage/


bio/mapdamage2/environment.yaml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
+dependencies:
+  - mapdamage2 =2.2.1=pyr40_0
```
johanneskoester, line 6: Please just pin to the version, not the build string (build strings are not guaranteed to stay valid forever).
```suggestion
  - mapdamage2 =2.2.1
```

johanneskoester (changes requested):
Thank you! It was actually not intentional to not have a test with picard markduplicates. Hence, can you add one for the new wrapper here, and probably copy it over to markduplicates as well?

fgvieira:
@johanneskoester sorry for all the new wrappers on a single PR, but kept making new ones and forgot to make new branches. Will do it next time...

bio/samtools/calmd/test/Snakefile
```diff
@@ -0,0 +1,11 @@
+rule samtools_calmd:
+    input:
+        bam = "{sample}.bam", # Can be 'sam', 'bam', or 'cram'
+	ref = "genome.fasta"
```
dlaehnemann, line 4: Tabs instead of whitespace indentation, should be:
```suggestion
        ref = "genome.fasta"
```

bio/samtools/calmd/test/Snakefile
```diff
@@ -0,0 +1,11 @@
+rule samtools_calmd:
+    input:
+        bam = "{sample}.bam", # Can be 'sam', 'bam', or 'cram'
```
dlaehnemann, line 3: Maybe make this a generic input name like `alignment =`? `samtools calmd` should do automatic file type detection and should even throw useful error messages if that doesn't work. And this way, users only have to adjust the extension, and not the input variable key name.

bio/samtools/sort/wrapper.py
```diff
@@ -8,14 +8,16 @@
 from snakemake.shell import shell
 
 
-prefix = os.path.splitext(snakemake.output[0])[0]
+prefix = snakemake.params.get("prefix", "")
+if not prefix:
+    prefix = os.path.splitext(snakemake.output[0])[0]
```
dlaehnemann, line 13: I would argue that this should always be appended to any specified prefix. As a prefix specified via `params.prefix` will usually be the same for all instances of the rule, this could otherwise lead to multiple instances writing to and reading from the same temporary files, e.g. when bam files with the same name exist in separate subfolders and are handled at the same time. Or maybe, to distinguish between the prefix and where to put temporary sorting files, the `params` parameter could instead be named `tmp_folder` and the path without the extension then used at that location?

test.py
```diff
@@ -482,6 +501,162 @@ def test_bwa_index():
...
+        ["snakemake", "--cores", "1", "mapped/a.se.bam", "--use-conda", "-F"],
+    )
+
+
+def test_bwa_bam_pe():
+    run(
+        "bio/bwa/sam",
+        ["snakemake", "--cores", "1", "mapped/a.pe.bam", "--use-conda", "-F"],
+    )
+
```
dlaehnemann, line 530: This is the same two tests twice, or am I missing some subtle difference?

test.py
```diff
@@ -482,6 +501,162 @@ def test_bwa_index():
...
+            "1",
+            "mapped/a.pe.picard_sort.bam",
+            "--use-conda",
+            "-F",
+            "-s",
+            "Snakefile_picard",
+        ],
+    )
+
+
```
dlaehnemann, line 659: As above, it seems like every test appears twice in exactly the same form. Did you intend to edit one variant each?

dlaehnemann (changes requested):
Thanks for this big contribution! This generally looks good, just a number of suggestions and some minor things to fix.

As you already noted, smaller PRs are probably the way to go in the future. Makes it easier and quicker to review, and easier to merge in case of conflicts.

test.py
fgvieira: Even though it is the same wrapper, the tests are for different situations, namely:
`sam` from single-end
`sam` from paired-end
`bam` from single-end
`bam` from paired-end

test.py
fgvieira: Again, even though it is a single wrapper, I added several tests for several outputs, namely:
samtools sorted `bam` from single-end
samtools sorted `bam` from paired-end
picard sorted `sam` from single-end
picard sorted `sam` from paired-end
picard sorted `bam` from single-end
picard sorted `bam` from paired-end

bio/picard/markduplicateswithmatecigar/wrapper.py
```diff
@@ -0,0 +1,15 @@
+__author__ = "Johannes Köster"
+__copyright__ = "Copyright 2016, Johannes Köster"
+__email__ = "koester@jimmy.harvard.edu"
+__license__ = "MIT"
+
+
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
```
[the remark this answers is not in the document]
fgvieira: Added JVM memory handling just like in other picard/gatk wrappers

test.py
dlaehnemann: Ah, OK. I did miss the subtle difference between `s`am and `b`am, of course. Sorry, my bad!

bio/bcftools/sort/meta.yaml
```diff
@@ -0,0 +1,4 @@
+name: bcftools sort
+description: Sort vcf/bcf file.
```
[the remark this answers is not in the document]
fgvieira: Added link to docs on all wrappers, plus some more...

dlaehnemann:
Cheers for the immediate flurry of activity! You might just be quicker than PR #204! :D

bio/samtools/calmd/test/Snakefile
fgvieira: Implemented generic input name and automated output format detection

bio/bcftools/sort/test/Snakefile
```diff
@@ -0,0 +1,14 @@
+rule bcftools_sort:
+    input:
+        "{sample}.bcf"
+    output:
+        "{sample}.sorted.bcf"
+    log:
+        "logs/bcftools/sort/{sample}.log"
+    params:
+        tmp_dir = "`mktemp -d`",
+        extras = "-Ob"
```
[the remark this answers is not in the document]
fgvieira: What if the user wants to output uncompressed `bcf` for further piping?

bio/samtools/sort/wrapper.py
fgvieira: Fixed issue when specifying prefix

johanneskoester (changes requested):
Thanks a lot for this huge contribution! However, please remove the bwa sam wrappers. I'd rather have the existing bwa wrappers dynamically react on the chosen extension of the output file.

bio/samtools/sort/wrapper.py
```diff
@@ -4,18 +4,24 @@
 __license__ = "MIT"
 
 
-import os
+import os.path as p
```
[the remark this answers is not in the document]
fgvieira: Reverted back import changes

fgvieira:
Not quite sure what you mean by 
> [...] I'd rather have the existing bwa wrappers dynamically react on the chosen extension of the output file.

Current `bwa sampe` and `bwa samse` wrappers have the same output file extension. In fact, they are identical except for the number of input `.fastq` and `.sai` files.

The only thing that the `bwa sam` wrapper I include here does, is to infer which algorithm to use (`sampe` or `samse`) from the number of input `fastq` and `sai` files.

johanneskoester:
> Not quite sure what you mean by
> 
> > [...] I'd rather have the existing bwa wrappers dynamically react on the chosen extension of the output file.
> 
> Current `bwa sampe` and `bwa samse` wrappers have the same output file extension. In fact, they are identical except for the number of input `.fastq` and `.sai` files.
> 
> The only thing that the `bwa sam` wrapper I include here does, is to infer which algorithm to use (`sampe` or `samse`) from the number of input `fastq` and `sai` files. Just like we have a single `bwa mem` wrapper for both PE and SE, I think it makes sense to also have a single wrapper for `bwa sam`.

Ah, sorry, I misunderstood that. The problem is the name, which somehow suggests dealing with SAM files. Maybe call it `bwa samxe` instead?

fgvieira:
sounds good...

bio/bcftools/sort/test/Snakefile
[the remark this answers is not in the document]
dlaehnemann: Not sure if you have seen this suggestion, @fgvieira? Once this is in, I think we can merge the whole PR.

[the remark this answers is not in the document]
dlaehnemann: Ah, wait. I now see that it is implemented. Maybe describe the option of using it somewhere? Either in the description of the wrapper, or by including `uncompressed_bcf=false` with an explanatory comment in the example rule?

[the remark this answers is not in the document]
fgvieira: I've added the `uncompressed_bcf=false` to the example rule

bio/bcftools/sort/wrapper.py
```diff
@@ -0,0 +1,54 @@
+__author__ = "Filipe G. Vieira"
+__copyright__ = "Copyright 2021, Filipe G. Vieira"
```
dlaehnemann, line 2: Sorry, 2020 is not yet over...
```suggestion
__copyright__ = "Copyright 2020, Filipe G. Vieira"
```

bio/bwa/samxe/wrapper.py
```diff
@@ -0,0 +1,87 @@
+"""Snakemake wrapper for both bwa samse and sampe."""
+
+__author__ = "Filipe G. Vieira"
+__copyright__ = "Copyright 2021, Filipe G. Vieira"
```
dlaehnemann, line 4:
```suggestion
__copyright__ = "Copyright 2020, Filipe G. Vieira"
```

bio/mapdamage2/wrapper.py
```diff
@@ -0,0 +1,30 @@
+__author__ = "Filipe G. Vieira"
+__copyright__ = "Copyright 2021, Filipe G. Vieira"
```
dlaehnemann, line 2:
```suggestion
__copyright__ = "Copyright 2020, Filipe G. Vieira"
```

bio/samtools/calmd/wrapper.py
```diff
@@ -0,0 +1,14 @@
+__author__ = "Filipe G. Vieira"
+__copyright__ = "Copyright 2021, Filipe G. Vieira"
```
dlaehnemann, line 2:
```suggestion
__copyright__ = "Copyright 2020, Filipe G. Vieira"
```

dlaehnemann (approved):

johanneskoester (approved):

johanneskoester:
Awesome, thanks a lot!

dlaehnemann:
Yes, thanks again @fgvieira. Cool additions and improvements to the wrapper collection!

Issue #187, opened 2020-09-21
No "code" displayed in wrappers' documentation


tdayris:
**Snakemake-wrappers 0.66.0, snakemake 5.24.0**
The disparition of the displayed code appeared after addition of meta wrappers. So, the latest version of snakemake-wrappers

**Describe the bug**
There is no code displayed anymore in any wrapper's documentation. The code is still displayed in meta-wrappers.

**Minimal example**
See missing documentation in:
https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/bedtools/slop.html#code

See available documentation in:
https://snakemake-wrappers.readthedocs.io/en/stable/meta-wrappers/star_arriba.html#code

tdayris:
I found the solution in the PR.

PR #188, opened 2020-09-21
Missing code in wrappers' doc. Error #187

tdayris:
Hi,

This is a fix to the error labelled #187 
It was just a matter of namespace overloading.

Thanks in advance for your reviews


jafors (approved):
Thanks a lot, I don't know how I missed this.

PR #189, opened 2020-09-21
Java (picard) -Xmx parameter

tdayris:
Hi,

I was wondering if this is a good idea or not: based on the `resources` parameter in snakemake rules, then build `-Xmx` parameter automatically for any java tool that allow this kind of parameters.

Here, the conda wrapper of almost any picard subcommands handles `-Xm*`, and `-D*` options, so `Xmx` is handled.

If you agree with this, I am willing to do the same in Picard, SnpSift, SnpEff, GATK and Trinity (max_memory). If you see any other tool that would need such changes, please tell me and I'll be happy to make the changes.

Thanks in advance for your reviews.

PS: The modifications in `generate_docs.py` come from a PR i did this morning. I don't know how to hide it since it has been already accepted.

### DIFF
--- a/bio/picard/markduplicates/wrapper.py
+++ b/bio/picard/markduplicates/wrapper.py
@@ -8,8 +8,16 @@ from snakemake.shell import shell
 
 log = snakemake.log_fmt_shell(stdout=True, stderr=True)
 
+memory = ""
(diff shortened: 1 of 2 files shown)


johanneskoester (approved):
This is an awesome idea, I would very much welcome it!

PR #190, opened 2020-09-22
Add wrappers for vg and wgsim

cartoonist:
I created the same PR a while ago when the repository was on BitBucket. It was stalled because vg was not on conda at that time. Now, they are on conda. I have updated those wrappers with the latest version of vg and wgsim.


bio/vg/construct/environment.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
+  - bioconda
+  - defaults
+dependencies:
+  - vg >=1.27.0
```
johanneskoester, line 5:
```suggestion
  - vg ==1.27.0
```
We never know what happens in the future, hence wrappers are usually pinned to an exact version.

johanneskoester (changes requested):
Awesome, I am really happy to finally being able to merge this. Just one request about the version in each of the conda envs below.

bio/vg/construct/environment.yaml
cartoonist: That makes sense. I pinned `vg` and `wgsim` to the exact version.

cartoonist:
Just a kind reminder. I applied the requested changes.

johanneskoester (approved):

johanneskoester:
Thanks a lot for the reminder!

PR #195, opened 2020-09-25
feat: added wrapper for bgzip

williamrowell:
Added wrapper for using bgzip to compress a file.


bio/bgzip/test/Snakefile
```diff
@@ -0,0 +1,12 @@
+rule bgzip:
+    input:
+        "{prefix}.vcf"
+    output:
+        "{prefix}.vcf.gz"
+    params:
+        extra: ""  # optional
```
jafors, line 7:
```suggestion
    params:
        extra=""  # optional
```

jafors (changes requested):
Thanks for the effort in all your PRs. I'll start with this one, solid work, just one thing below in the `params`.

jafors:
Oh, and I just saw that you are missing testcases for most of your wrappers. Please remember to commit the test data and the updated test.py when you provide a new wrapper.

williamrowell:
Just wanted to drop a note that I'm not ignoring you, just busy at the moment.  Thanks for following up.  In some cases (like this one, for instance), I should be able to easily add tests, but others will likely take more time.

bio/bgzip/test/Snakefile
williamrowell: Fixed this error and reformatted Snakefile with snakefmt.

williamrowell:
Sorry this took a while to get back to, and thanks for your patience.  I made the change you suggested, and added a test case to test.py. This is the first time I've added a test case.  I will work on some of the other PRs as well.

jafors (approved):

jafors (changes requested):
Hi, thanks for the additional work! One thing I just saw - could you add a smaller testcase? 14MB seems a bit to large for a test.vcf - actually, one or two records would be sufficient.

williamrowell:
Sorry, I had originally grabbed the example VCF from the tabix wrapper, not realizing it was 14MB.  I just replaced the test case with a minimal with a handful of entries.

bio/bgzip/wrapper.py
```diff
@@ -0,0 +1,19 @@
+__author__ = "William Rowell"
+__copyright__ = "Copyright 2020, William Rowell"
+__email__ = "wrowell@pacb.com"
+__license__ = "MIT"
+
+
+from snakemake.shell import shell
+
+extra = snakemake.params.get("extra", "")
```
jafors, line 9: The wrapper should probably do some parameter checks in `extra` - if e.g. someone wants to use `-c|--stdout`, there needs to be a `> {snakemake.output}` appended to the end of the shell call.
You could do this with a new variable like `output` which contains `> {snakemake.output}` if  `-c|--stdout` in `extra` and an empty string otherwise.

jafors (changes requested):
Thanks! Sorry to bother you further, but I already proposed an additional improvement which we should think about. Hopefully, this should not be too much of a hassle.

bio/bgzip/wrapper.py
johanneskoester: I think that should even be the only behavior. This makes the choice of the output file completely free.

williamrowell:
I've updated, changed the default behavior to `bgzip -c`, and added a test case.  It's ready for review.

johanneskoester (approved):

PR #196, opened 2020-09-25
added pbsv discover and call commands

williamrowell:
pbsv is a suite of tools to detect and call structural variants from PacBio CLR or HiFi reads.  I have added wrappers for the discover and call subcommands.


johanneskoester (changes requested):
Thanks a lot! Please add test cases, and reformat the `Snakefile`s to adhere to snakefmt's style (except comments, which are currently sometimes messed up by snakefmt).

PR #197, opened 2020-09-25
added mosdepth for coverage calculation

williamrowell:
Added a wrapper for mosdepth for coverage calculations.


johanneskoester (changes requested):
Thanks a lot! Please format with snakefmt (but check the comments, snakefmt sometimes still messes them up). For the test case, please add it to `test.py`, otherwise it won't be called.

fgvieira:
Hi @williamrowell, any idea when you will add these changes?
It would be quite nice to have `mosdepth` added the wrappers.

fgvieira:
@johanneskoester, I can add the test case to `test.py`. If you merge this PR, I'll take it from there and can do it this week.
Or how do you thing it is best to proceed?

williamrowell:
Sorry, I've been swamped.  If you can point to a good test BAM, I'll add the test case this week.

fgvieira:
It looks like you already have some BAM files in the wrapper, no?
I think that @johanneskoester meant that you need to add the tests to `test.py`, so that the wrapper gets actually tested.

williamrowell:
@johanneskoester @fgvieira

Currently passes tests but fails linting because of the reserved term "prefix".  I used "prefix" because this is the name of the mosdepth parameter.  What is the preferred way of solving this?

fgvieira:
I would

williamrowell:
I agree with all except for the `prefix` fix, unless we switch to named outputs, since `{prefix}.mosdepth.summary.txt` seems to be the only output with a fixed name.  Other named outputs could be optional and not used by the wrapper.

```python
prefix = re.sub("\.mosdepth\.summary\.txt", "", snakemake.output.summary)
```

fgvieira:
I agree, you can use named outputs and use `summary` to infer the prefix (just like you suggested).

fgvieira:
@williamrowell, it would also be nice to have a `param` to specify precision, that is then passed to `mosdepth` as:
```
MOSDEPTH_PRECISION=5 mosdepth ...
```

dlaehnemann:
It would definitely be nice to have mosdepth in here and thanks for all the work on this. Any time in the near-term future to pick this up, or would it be OK for me to pick this PR up and move it along?

fgvieira:
I think you can move it along... as far as I can see, it is only missing to fix `prefix` and `MOSDEPTH_PRECISION`.
Let me know if there is anything I can help with.

dlaehnemann (approved):
OK, I think this is good to go. Added some more functionality and tests for everything. And all tests, including formatting and linting, are passing. :tada:

williamrowell:
Thanks for picking this up.  I just didn't have time to add the rest of the features.

dlaehnemann:
Yeah. I really like this tool, but its command line interface and file handling scheme requires quite a bit of fiddling, here...

johanneskoester (approved):

PR #200, opened 2020-09-25
added wrapper for samtools fasta/fastq

williamrowell:
Added wrapper for samtools fasta and samtools fastq.  Because the syntax is so similar, added "outputtype" param to specify output.


johanneskoester (changes requested):
Thanks a lot! Just one thing: I think the wrapper should be called `bio/samtools/fastx` which, in my experience seems to be the convention when referring to both fasta and fastq, instead of using an `n`.

johanneskoester (changes requested):
Further, please add a test case to `test.py`.

williamrowell:
I've renamed this, updated, and added a test case.  It's ready for review.

johanneskoester (approved):

johanneskoester:
Thank you!

PR #201, opened 2020-09-25
feat: added wrapper for bcftools stats

williamrowell:
Added bcftools stats subcommand wrapper.


johanneskoester (commented):
Thanks a lot! Please add a test case with small test data and a corresponding function in `test.py`.

johanneskoester (changes requested):
See above.

williamrowell:

johanneskoester (approved):

PR #202, opened 2020-09-25
feat: added wrapper for seqtk seq subcommand

williamrowell:
seqtk seq can be used to easily convert or reformat fasta/q files.


johanneskoester (changes requested):
Please format the `Snakefile` with snakefmt, and add a test case to `test.py`.

johanneskoester:
And thanks a lot!

williamrowell:

johanneskoester (approved):

PR #203, opened 2020-09-25
feat: added wrapper for using tabix to query an indexed file

williamrowell:
The current tabix wrapper can only be used to index a file, but tabix can also be used to query an indexed file, like `tabix -h my.vcf.gz chr1 > chr1.vcf.gz` or `tabix coverage.bed.gz chr20 > chr20.bed.gz`.  Added this as a separate wrapper.


johanneskoester (changes requested):
Thanks a lot! Please add a test case to `test.py`.

williamrowell:
I added test cases and reformatted a few of these old PRs, but there are merge conflicts because my test.py files are out of date.  I'm not sure of the least complicated way to straighten this out.

williamrowell:

johanneskoester (approved):

PR #204, opened 2020-09-25
Java (multiple tools) -Xmx parameter

tdayris:
Hi again,

As promised, all the memory additions in the previously discussed wrappers (#189 )

* Please keep an eye to the mb → gb conversion for trinity: I did a dummy division by 1024 and rounded it down with the `int()` function. If you know a better way to do it, I'll make the changes.
* Like the last time, there is the modification pushed on `generate_docs.py` still considered as 'new' despite being already accepted.

Many thanks in advance for your reviews

Edit: merge conflict solved


dlaehnemann (changes requested):
@tdayris , this is really cool. Thanks for taking the time to update all these recipes.

Your strategy to convert from Mb to Gb seems good, the `-Xmx` parameter seems to [work with powers of `1024`](https://stackoverflow.com/questions/32855984/does-java-xmx1g-mean-109-or-230-bytes/32858015#32858015).

One general thing I would suggest, is using the `resources` field in the snakemake `rule`s, so that snakemake can directly use the arguments for restricting global memory usage (by setting `--resources mem_mb=4096` should you only have 8G of RAM on your machine, for example) and requesting the amount of memory from a cluster system (as part of a `qsub` command that uses `resources.mem_mb`, for example). So instead of checking `snakemake.params.get("extra", "")`, you would check for `snakemake.resources.get("mem_mb", "")`.
Also, to showcase that users can set this parameter, I would always add it to the respective `test/Snakefile`, because that is what's rendered as the example in the docs and users will usually copy-paste this. You could simply set this to something like:
```
    # optional specification of memory usage of the JVM that snakemake will respect with global
    # resource restrictions (https://snakemake.readthedocs.io/en/latest/snakefiles/rules.html#resources)
    # and which can be used to request RAM during cluster job submission as `{resources.mem_mb}`:
    # https://snakemake.readthedocs.io/en/latest/executing/cluster.html#job-properties
    resources:
        mem_mb=1024
```
Or if unsure about a good default value, you could comment the two resource specification lines out.

Another question---although I am not necessarily set on this---would be if we want to generally use `mem_gb` instead of `mem_mb` wherever the bioconda recipes allow it. For any of the tools that I have used, I usually deal in `Gb`s of memory, rather than `Mb`s. But if you disagree, feel free to keep it as is, with `mem_mb`. Especially, as the snakemake docs example for `--resources` deals in `mem_mb`.

And finally, with a quick scan across the wrappers, I identified some extra ones that would benefit from your update -- I checked that their bioconda scripts would respect a setting of `-Xmx`:

* jannovar
* snpeff
* trimmomatic
* varscan

However, this might still be an incomplete list. But the more we get done at once, the better. Also, should you need help with all the stuff I request, feel free to ping me in!

bio/trinity/wrapper.py
```diff
@@ -10,6 +10,10 @@
 
 extra = snakemake.params.get("extra", "")
 max_memory = snakemake.params.get("max_memory", "10G")
+if "mem_mb" in snakemake.resources.keys():
+    if snakemake.resources["mem_mb"] > 10240:
```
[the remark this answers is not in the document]
tdayris: Hi,

Yes, the trailing zero is intended : trinity's default memory value is 10Gb (see line above). I did not want to change the previous behavior for user who did not set any memory value.

tdayris:
Hi,

Many thanks for your review !

* The wrappers you pointed out were also edited. Now trimmomatic, varscan, snpeff and jannovar have the automatic `-Xmx` parameter filled by the wrappers.
* The information about memory is, indeed, taken from `snakemake.resources`, the check in `snakemake.params.get("extra", "")` is here to avoid to give `-Xmx` parameter twice if the user set one value as extra parameters. If you think I shall not handle this case, please tell me.

tdayris:
Hi,

Least changes:
* Memory can be passed with either `mem_mb` and `mem_gb`
* Picard wrappers have been added to the list of tools modified
* java temp was added in this PR
* Two picard tools did not have proper tests in `test.py`, these were added
* memory information is **always** taken from `snakemake.resources`, unless specified by user in extra (in order to preserve retro-compatibility)

Snakefile test-cases will be updated soon.

dlaehnemann:
Really cool work!

One thought I had upon browsing through the current state of changes: I wonder if it might be a good idea to have a mechanism by which wrappers can share common code snippets. It would be great if we could simply have one line of code in each of the Java recipes that would import the full-blown Java memory (and further options) solution, i.e. this code snippet:
```python
java_opts = snakemake.params.get("java_opts", "")


# Getting memory in megabytes, if java opts is not filled with -Xmx parameter
# By doing so, backward compatibility is preserved
if "mem_mb" in snakemake.resources.keys() and not "-Xmx" in java_opts:
    java_opts += " -Xmx{}M".format(snakemake.resources["mem_mb"])

# Getting memory in gigabytes, for user convenience. Please prefer the use
# of mem_mb over mem_gb as advised in documentation.
elif "mem_gb" in snakemake.resources.keys() and not "-Xmx" in java_opts:
    java_opts += " -Xmx{}G".format(snakemake.resources["mem_gb"])

# Getting java temp directory from output files list, if -Djava.io.tmpdir
# is not provided in java parameters. By doing so, backward compatibility is
# not borken.
if "java_temp" in snakemake.output.keys() and not "-Djava.io.tmpdir" in java_opts:
    java_opts += " -Djava.io.tmpdir={}".format(snakemake.output["java_temp"])
```
Maybe @jafors, @tedil, @johanneskoester or someone else has a good idea to do this in a proper Pythonic way? Some sort of mini `common.py` or `utils.py` module in the main repo folder that provides such code, maybe? And I guess it would need to be a function call on the `snakemake` object?

tdayris:
Hi again,

Very good idea indeed. Shareable resources for java options, minimal threading requirements, and others would be very interesting.

Least changes:
* Snakefiles updated
* minor bug-fixes

I think this PR is complete now.

bio/gatk/applybqsr/wrapper.py
```diff
@@ -9,9 +9,26 @@
...
 
+# Getting memory in megabytes, if java opts is not filled with -Xmx parameter
+# By doing so, backward compatibility is preserved
+if "mem_mb" in snakemake.resources.keys() and not "-Xmx" in java_opts:
+    java_opts += " -Xmx{}M".format(snakemake.resources["mem_mb"])
+
+# Getting memory in gigabytes, for user convenience. Please prefer the use
+# of mem_mb over mem_gb as advised in documentation.
+elif "mem_gb" in snakemake.resources.keys() and not "-Xmx" in java_opts:
+    java_opts += " -Xmx{}G".format(snakemake.resources["mem_gb"])
```
dlaehnemann, line 20: I would prefer for `snakemake.resources` to take precedence before `snakemake.params.extra`, as it is the "most snakemakey way" of resource specification. And if someone went to the lengths of specifying resources, they probably expect these resources to be respected and a remaining option setting in `params` is probably outdated and should be removed. So the best solution is probably to simply throw an error and exit  when both `resources` and `params` want to specify Java memory, e.g. something like this under both the `if` and `elif` clause:
```
if "-Xmx" in java_opts or "-Xmx" in extra:
    sys.exit("Please specify memory requirements via params.mem_mb or params.mem_gb only, you also specified some via params.extra or params.java_opts.")
```

Which obviously leads to a bunch of extra code and probably increases the importance for centralised utility functions.... ;)

bio/gatk/baserecalibrator/test/Snakefile
```diff
@@ -11,5 +11,11 @@ rule gatk_baserecalibrator:
     params:
         extra="",  # optional
         java_opts="", # optional
+    # optional specification of memory usage of the JVM that snakemake will respect with global
+    # resource restrictions (https://snakemake.readthedocs.io/en/latest/snakefiles/rules.html#resources)
+    # and which can be used to request RAM during cluster job submission as `{resources.mem_mb}`:
+    # https://snakemake.readthedocs.io/en/latest/executing/cluster.html#job-properties
```
dlaehnemann, line 17: Maybe add in the possibility to specify memory in Gb RAM:
`    # RAM can also be specified in Gibioytes RAM via `resources.mem_gb`

bio/gatk/applybqsr/wrapper.py
```diff
@@ -9,9 +9,26 @@
...
+
+# Getting java temp directory from output files list, if -Djava.io.tmpdir
+# is not provided in java parameters. By doing so, backward compatibility is
+# not borken.
+if "java_temp" in snakemake.output.keys() and not "-Djava.io.tmpdir" in java_opts:
```
dlaehnemann, line 25: Similar to the suggestion regarding memory specification, I would argue that specifying `-Djava.io.tmpdir` in two locations (and possibly differently) should rather throw an error and exit. This maintains backward compatibility, but also ensures that an old specification is removed once a `java_temp` spec is added.

bio/gatk/applybqsr/wrapper.py
```diff
@@ -9,9 +9,26 @@
...
+    java_opts += " -Xmx{}M".format(snakemake.resources["mem_mb"])
+
+# Getting memory in gigabytes, for user convenience. Please prefer the use
+# of mem_mb over mem_gb as advised in documentation.
+elif "mem_gb" in snakemake.resources.keys() and not "-Xmx" in java_opts:
+    java_opts += " -Xmx{}G".format(snakemake.resources["mem_gb"])
+
+# Getting java temp directory from output files list, if -Djava.io.tmpdir
+# is not provided in java parameters. By doing so, backward compatibility is
+# not borken.
```
dlaehnemann, line 24: Typo:
```suggestion
# not broken.
```

bio/gatk/baserecalibrator/test/Snakefile
dlaehnemann: Ah, and maybe we move this whole comment block one row below and indent one level further? This way, it more clearly refers to the `mem_mb` entry which is then right below it.

bio/jannovar/wrapper.py
```diff
@@ -11,6 +11,21 @@
 log = snakemake.log_fmt_shell(stdout=False, stderr=True)
```
dlaehnemann, line 11: I think, the `resources: mem_mb=something` specification is still missing from `bio/jannovar/test/Snakefile`?

dlaehnemann (commented):
Once again, thanks for this major effort! And thanks for even picking up on some missing tests and further Java wrappers along the way!

As before, this all looks good and I merely have some suggestions. Feel free to disagree on any of them.

johanneskoester:
> Really cool work!
> 
> One thought I had upon browsing through the current state of changes: I wonder if it might be a good idea to have a mechanism by which wrappers can share common code snippets. It would be great if we could simply have one line of code in each of the Java recipes that would import the full-blown Java memory (and further options) solution, i.e. this code snippet:
> 
> ```python
> java_opts = snakemake.params.get("java_opts", "")
> 
> 
> # Getting memory in megabytes, if java opts is not filled with -Xmx parameter
> # By doing so, backward compatibility is preserved
> if "mem_mb" in snakemake.resources.keys() and not "-Xmx" in java_opts:
>     java_opts += " -Xmx{}M".format(snakemake.resources["mem_mb"])
> 
> # Getting memory in gigabytes, for user convenience. Please prefer the use
> # of mem_mb over mem_gb as advised in documentation.
> elif "mem_gb" in snakemake.resources.keys() and not "-Xmx" in java_opts:
>     java_opts += " -Xmx{}G".format(snakemake.resources["mem_gb"])
> 
> # Getting java temp directory from output files list, if -Djava.io.tmpdir
> # is not provided in java parameters. By doing so, backward compatibility is
> # not borken.
> if "java_temp" in snakemake.output.keys() and not "-Djava.io.tmpdir" in java_opts:
>     java_opts += " -Djava.io.tmpdir={}".format(snakemake.output["java_temp"])
> ```
> 
> Maybe @jafors, @tedil, @johanneskoester or someone else has a good idea to do this in a proper Pythonic way? Some sort of mini `common.py` or `utils.py` module in the main repo folder that provides such code, maybe? And I guess it would need to be a function call on the `snakemake` object?

+1 on @dlaehnemann's last review. Once that is done, we can merge. And awesome work on this PR. Regarding common functions, this is indeed a good idea. The idiomatic solution is to create a Python package for that, called e.g. snakemake-wrapper-utils. I'll do that ASAP.

johanneskoester:
> > Really cool work!
> > One thought I had upon browsing through the current state of changes: I wonder if it might be a good idea to have a mechanism by which wrappers can share common code snippets. It would be great if we could simply have one line of code in each of the Java recipes that would import the full-blown Java memory (and further options) solution, i.e. this code snippet:
> > ```python
> > java_opts = snakemake.params.get("java_opts", "")
> > 
> > 
> > # Getting memory in megabytes, if java opts is not filled with -Xmx parameter
> > # By doing so, backward compatibility is preserved
> > if "mem_mb" in snakemake.resources.keys() and not "-Xmx" in java_opts:
> >     java_opts += " -Xmx{}M".format(snakemake.resources["mem_mb"])
> > 
> > # Getting memory in gigabytes, for user convenience. Please prefer the use
> > # of mem_mb over mem_gb as advised in documentation.
> > elif "mem_gb" in snakemake.resources.keys() and not "-Xmx" in java_opts:
> >     java_opts += " -Xmx{}G".format(snakemake.resources["mem_gb"])
> > 
> > # Getting java temp directory from output files list, if -Djava.io.tmpdir
> > # is not provided in java parameters. By doing so, backward compatibility is
> > # not borken.
> > if "java_temp" in snakemake.output.keys() and not "-Djava.io.tmpdir" in java_opts:
> >     java_opts += " -Djava.io.tmpdir={}".format(snakemake.output["java_temp"])
> > ```
> > 
> > 
> > Maybe @jafors, @tedil, @johanneskoester or someone else has a good idea to do this in a proper Pythonic way? Some sort of mini `common.py` or `utils.py` module in the main repo folder that provides such code, maybe? And I guess it would need to be a function call on the `snakemake` object?
> 
> +1 on @dlaehnemann's last review. Once that is done, we can merge. And awesome work on this PR. Regarding common functions, this is indeed a good idea. The idiomatic solution is to create a Python package for that, called e.g. snakemake-wrapper-utils. I'll do that ASAP.

Here we go: https://github.com/snakemake/snakemake-wrapper-utils
@tdayris if you like, please PR your common java helper snippets as functions into this package.

tdayris:
Thanks a lot! What a quick answer! I'll commit the changes before the end of the week.

tdayris:
Hi, changes have been made, but I can't make through the tests since there is a very small [bug](https://github.com/snakemake/snakemake-wrapper-utils/pull/3) in the snakemake-wrapper-utils

dlaehnemann:
Your PR is merged and a new release of the `snakemake-wrapper-utils` is up on the bioconda channel!

bio/gatk/baserecalibrator/test/Snakefile
tdayris: Request included in `snakemake-wrapper-utils`

bio/jannovar/wrapper.py
tdayris: You were right, it is now included.

bio/gatk/applybqsr/wrapper.py
tdayris: Suggestion included within the snakemake-wrapper-utils

bio/gatk/applybqsr/wrapper.py
tdayris: Suggestion included within the snakemake-wrapper-utils, thanks !

tdayris:
Hi, once again, thank you very much for your help with this work.

Changes since last time:
* snakemake-wrapper-utils have replaced the copy-pasted code
* snakemake-wrapper-utils included in the `environment.yaml`-files

Feel free to point out any mistake I have made

bio/jannovar/test/Snakefile
```diff
@@ -6,6 +6,8 @@ rule jannovar:
         "jannovar/{sample}.vcf.gz"
     log:
         "logs/jannovar/{sample}.log"
+    resources:
```
dlaehnemann, line 9:
```suggestion
    # optional specification of memory usage of the JVM that snakemake will respect with global
    # resource restrictions (https://snakemake.readthedocs.io/en/latest/snakefiles/rules.html#resources)
    # and which can be used to request RAM during cluster job submission as `{resources.mem_mg}`:
    # https://snakemake.readthedocs.io/en/latest/executing/cluster.html#job-properties
    resources:
```

dlaehnemann (approved):
Very cool! Many thanks for the systematic update of Java memory management options and all the additions of docs links!

The only thing that could still be amended, would be adding `resources: mem_mb = 1024` entries to the `gatk3` wrappers:
* realignertargetcreator
* printreads
* baserecalibrator
* indelrealigner

But by now, these are kind of legacy wrappers (with gatk4 long released as the default). So it's maybe not necessary to update them. So just let me know if you want to add this (if so, note that I did one little commit adding a comment in the jannovar Snakefile, so pull you branch before changing it locally) or if I should just merge as is.

tdayris:
Thanks you, the resources information were added in GATK3. You are right.

dlaehnemann:
@tdayris Did you push this already? Because I can't see it here in the PR, yet.

tdayris:
Wrong branch, my fault ! here are the modifications

dlaehnemann:
Cheers! And thanks again for this major effort! :tada:

PR #205, opened 2020-09-27
update for macs2 callpeak wrapper for --broad option

AntonieV:
Update for macs2 callpeak wrapper in case the --broad-cutoff value is set but the --broad option is missing in params.


dlaehnemann (approved):
LGTM! Thanks!

PR #206, opened 2020-09-30
Homer annotatepeaks

AntonieV:
A new wrapper for Homer annotatePeaks.pl.


bio/homer/annotatePeaks/wrapper.py
```diff
@@ -0,0 +1,95 @@
...
+    cmp_Liftover: "-cmpLiftover",
+    adv_ann: "-ann",
+    mfasta: "-mfasta",
+    mbed: "-mbed",
+    mlogic: "-mlogic",
+}
+
+requires_motives = False
+for i in opt_files:
+    if i and not i == "":
```
jafors, line 64:
```suggestion
for i in opt_files:
    if i:
```

jafors: `i and not i == ""` is a tautology, `if i:` is totally sufficient here.

jafors (changes requested):
Holy cow, that's an impressive wrapper. Very detailed for all cases. For now, I have just a very small addition.

bio/homer/annotatePeaks/wrapper.py
AntonieV: Many thanks for reviewing the PR and your comments.

bio/homer/annotatePeaks/wrapper.py
```diff
@@ -0,0 +1,95 @@
...
+    adv_ann: "-ann",
+    mfasta: "-mfasta",
+    mbed: "-mbed",
+    mlogic: "-mlogic",
+}
+
+requires_motives = False
+for i in opt_files:
+    if i:
+        extra += " {flag} {file}".format(flag=opt_files[i], file=i)
```
dlaehnemann, line 65: Instead of loading every possible input explicitly above, we could implicitly load and check them here:
```suggestion
    file = snakemake.input.get(i, "")
    if file:
        extra += " {flag} {file}".format(flag=opt_files[i], file=file)
```

dlaehnemann (approved):
Very impressive wrapper, indeed. Just two little suggestion for an improvement. But in any case good to go!

bio/homer/annotatePeaks/wrapper.py
AntonieV: Many thanks for reviewing and your suggestions, it looks much cleaner and more compact! Some additional adjustments were necessary for this.

jafors (approved):

PR #207, opened 2020-10-02
added wrapper for dada2 filterAndTrim

cpauvert:
Hi,

DADA2 process sequences data from metabarcoding projects in order to construct OTU/ASV tables. I am not its developer nor contributor but a mere user which intend to use wrappers potential for my analyses. More info on DADA2 are available [here](https://benjjneb.github.io/dada2/index.html). 

Here is the first wrapper of an (intended) series of 7-8 DADA2 wrappers.

The pytest was conclusive, so I hope everything is complete. Do not hesitate to answer if more specifications or corrections are needed.

Best,


bio/dada2/filter-trim/test/Snakefile
```diff
@@ -0,0 +1,17 @@
+rule dada2_filter_trim:
+    input:
+        # Paired-end files without primers sequences
+        fwd=expand("trimmed/{sample}.1.fastq", sample=["a","b"]),
+	rev=expand("trimmed/{sample}.2.fastq", sample=["a","b"])
```
jafors, line 5: A difference in indentation.

```suggestion
        fwd=expand("trimmed/{sample}.1.fastq", sample=["a","b"]),
        rev=expand("trimmed/{sample}.2.fastq", sample=["a","b"])
```

bio/dada2/filter-trim/test/Snakefile
```diff
@@ -0,0 +1,17 @@
...
+    output:
+        filt=expand("filtered/{sample}.1.fastq", sample=["a","b"]),
+        filt_rev=expand("filtered/{sample}.2.fastq", sample=["a","b"]),
+        stats="reports/dada2/filter-trim.tsv"
+    params:
+        maxEE = 1,
+        truncLen = 0,
+        minLen = 1 # Set to 1 for the toy example but default is 20
```
jafors, line 13: Similar to `input` and `output`, spaces are not needed.
```suggestion
    params:
        maxEE=1,
        truncLen=0,
        minLen=1 # Set to 1 for the toy example but default is 20
```

bio/dada2/filter-trim/environment.yaml
```diff
@@ -0,0 +1,7 @@
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
+dependencies:
+  - bioconductor-dada2 = 1.16.0
```
jafors, line 6: We try to fix the exact version, rather use `==` instead of `=`.

```suggestion
dependencies:
  - bioconductor-dada2 ==1.16.0
```

jafors (changes requested):
Thanks a lot for this ongoing set of wrappers! I started with a few remarks.

bio/dada2/filter-trim/environment.yaml
cpauvert: Thanks! I did all your corrections in a single commit for clarity!

cpauvert:
Thanks for the reviews @jafors ! I might have to correct the same errors in the others wrappers.
Best, 
Charlie

bio/dada2/filter-trim/test/Snakefile
```diff
@@ -0,0 +1,17 @@
+rule dada2_filter_trim:
+    input:
+        # Paired-end files without primers sequences
+        fwd=expand("trimmed/{sample}.1.fastq", sample=["a","b"]),
+        rev=expand("trimmed/{sample}.2.fastq", sample=["a","b"])
+    output:
+        filt=expand("filtered/{sample}.1.fastq", sample=["a","b"]),
+        filt_rev=expand("filtered/{sample}.2.fastq", sample=["a","b"]),
+        stats="reports/dada2/filter-trim.tsv"
```
jafors, line 9: I guess the samples `a` and `b` are filtered independently? I'm thinking if the `expand` on two samples is an ideal example. On the one hand, it shows that the tool can take multiple samples at once, on the other hand, it looks a bit more complicated to me, and in a workflow, I'd probably filter all my samples in separate runs and don't use the `expand`, but that might just be my sole preference.
Any opinions from you or others?

jafors (changes requested):
Thanks for the updates! See below for one more thing to discuss. I'm not adamant on this, but would rather like to hear your opinion.

bio/dada2/filter-trim/test/Snakefile
cpauvert: Thanks for your inputs! Yes samples are filtered independently. I wanted to provide them all at once for convenience and to reduce wall clock time. The `dada2` takes "some" time to load, so I figured that loading once for this step would be better.

Sorry for any unneeded complexity, I am more trained in `R` than in `Snakemake`  so feel free to suggest a proper way to do, I won't be offended. If I understood correctly, you'd remove the `expand` and let `Snakemake` filter samples from the wildcards guesses?

jafors: Exactly. If you had, let's say 5 samples, you'd run the rule 5 times, which could run in parallel, given a sufficient number of cores.
If loading the `dada2` package takes some time and you don't have many cores available, it might indeed be faster to process samples in the same run, but I'd guess that for an example, it would probably be sufficient to even just use one sample and infer it over the wildcards. Then, you can also get a sample-specific report, which might be helpful in detecting problems with a specific sample or sth. like this (though I don't how the report looks at all, this is just an educated guess ;) )

cpauvert: Thanks for your explanation @jafors I'll follow your advice. At the same time I will add the log rule which I had postponed.

cpauvert:
Hi @jafors ! Following your inputs, I've corrected the `Snakefile` and set it on a per-sample basis instead of the `expand` call. I've even added the `log` rule which proves to be harder than thought with `R` wrapper compared to `Python` ones.
All the best, 
Charlie

jafors (changes requested):
Nice! Thank you very much. If you no longer need the `b`-samples, you could delete the fastq files and we're golden.

cpauvert:
Done! Thanks for guiding me to improve this wrapper. I shall apply the same to the others from the set!

cpauvert:
Set to draft because I realize that several parts needed to be well think ahead integration.

1 . Try to obey the _do not repeat yourself_ rule, especially when wrappers won't differ between single-end and paired-end processing.
2. Find a way (maybe from the #31 tximport) to include options without being redundant
3. Craft separate rules for paired-end data processing
I will work on it before submitting.
Best,

cpauvert:
I'm closing this pull request because it is now replaced by #220 and #218 for a different handling of reads (pe vs se). Thank you @jafors for the tips that made me rethink the other wrappers.
Best, 
Charlie

PR #209, opened 2020-10-05
added wrapper (1/10) for dada2 plotQualityProfile

cpauvert:
Hi,
Following #207 I propose here a wrapper for the `plotQualityProfile` function that should be executed in order to set a proper trimming length for the paired end sequences.
Toy examples are symlinks to reduce data load.
Best,


cpauvert:
Following @jafors advices on #207 the `expand` was replaced by samples wildcards. The test was therefore simplified. No logs needed here.
Thanks in advance for reviews,
Charlie

cpauvert:
This PR was not synchronizing with my last commits unfortunately. So it is now replaced by #230  and this one and the branch are closed.

PR #213, opened 2020-10-06
added dada2 wrapper (4/10) for derepFastq

cpauvert:
Hi,
A small wrapper in line with the started `dada2` integration into wrappers.
Best,
Charlie


cpauvert:
@dlaehnemann this one is also ready :)

bio/dada2/dereplicate-fastq/meta.yaml
```diff
@@ -0,0 +1,11 @@
+name: dada2_dereplicate_fastq
+description: |
+  `DADA2 <https://benjjneb.github.io/dada2/index.html>`_
+  Dereplication of FASTQ files using dada2 derepFastq function. Optional parameters are documented in the `manual <https://www.bioconductor.org/packages/release/bioc/manuals/dada2/man/dada2.pdf>`_.
+authors:
+  - Charlie Pauvert
+input: a FASTQ file
+params:
+  - R optional arguments passed to the wrapped function as named Python arguments
```
dlaehnemann, line 9: I'm not sure how clear this info is, maybe rephrase to something like:
```suggestion
  - optional arguments for derepFastq(), please provide them as python key=value pairs
```

bio/dada2/dereplicate-fastq/meta.yaml
```diff
@@ -0,0 +1,11 @@
...
+output: a RDS file containing a `derep-class` object
```
dlaehnemann, line 10: This really is nit-picking, now, and I only add this, as I already made a suggestion above. But as the sound R in RDS starts with a vowel, this should read `an RDS`. Or simply skip the `a` for brevity.

dlaehnemann (approved):
Looking good. Only minor suggestions, feel free to decline to address those... :D

bio/dada2/dereplicate-fastq/meta.yaml
cpauvert: This is nicer thanks! It's included.
But on a side note, these information in the `meta.yaml` do not show up in the documentations AFAIK.

dlaehnemann: Hmm, we should double-check this. You can compile the docs locally and have a look. There's a quick walk-through in the lower part of this contribution docs section:
https://snakemake-wrappers.readthedocs.io/en/stable/index.html#testing-locally

I would have thought, that it simply renders all `meta.yaml` entries as subheadings, but it might be that only certain names are rendered...

cpauvert: Yes thanks I had already did built locally. I have a WIP branch and some suggestions in mind. I might list it properly as an issue though.

cpauvert:
Great thank you @dlaehnemann !

PR #217, opened 2020-10-13
added dada2 wrapper (5/10) for dada (aka sample inference) function

cpauvert:
Hi,
A small wrapper in line with the started `dada2` integration into wrappers. Parameters can be passed to the R function with a Python dictionary.
Best,
Charlie


bio/dada2/sample-inference/meta.yaml
```diff
@@ -0,0 +1,13 @@
+name: dada2_learn_errors
```
dlaehnemann, line 1: This is the same as for the `dada2/learn-errors` wrapper. Shouldn't this instead read something like:
```suggestion
name: dada2_sample_inference
```

bio/dada2/sample-inference/meta.yaml
```diff
@@ -0,0 +1,13 @@
...
+description: |
+  `DADA2 <https://benjjneb.github.io/dada2/index.html>`_
+  Inferring sample composition using dada2 dada function. Optional parameters are documented in the `manual <https://www.bioconductor.org/packages/release/bioc/manuals/dada2/man/dada2.pdf>`_.
+authors:
+  - Charlie Pauvert
+input:
+  - derep: RDS file withe the dereplicated sequences
```
dlaehnemann, line 8: Typo:
```suggestion
  - derep: RDS file with the dereplicated sequences
```

bio/dada2/sample-inference/test/Snakefile
```diff
@@ -0,0 +1,19 @@
+rule dada2_sample_inference:
+    input:
+    # Dereplicated (aka unique) sequences of the sample
+        derep="uniques/{fastq}.RDS",
+        model="results/dada2/model_1.RDS" # Error model
```
dlaehnemann, line 5: Just a thought: You chose the same name for the input file as for the function argument name with `derep`, maybe do the same here (for consistency;):
```suggestion
        err="results/dada2/model_1.RDS" # Error model
```
Obviously, the `wrapper.R` needs the respective adjustment...

bio/dada2/sample-inference/meta.yaml
dlaehnemann: Or simply `dada2_dada` instead. As `dada` is the function that is called by this wrapper.

dlaehnemann (changes requested):
Just some minor suggestions, should be quickly addressed. This process is getting much smoother... :D

bio/dada2/sample-inference/meta.yaml
cpauvert: Oups that was a typo, thanks for spotting it! As for the name, I preferred to name the rule as per the step name in the [tutorial](https://benjjneb.github.io/dada2/tutorial.html) rather than the function. In order to ease the readability for the (coming soon) meta-wrapper ;)

bio/dada2/sample-inference/test/Snakefile
cpauvert: Yes that's possible. However, I just called it `model=` in the `dada2_learn_errors` rule earlier :/ What do you suggest. Change both?

dlaehnemann: Ah, yeah. It should be consistent in that sense, as well. So unless this disagrees with any named parameters in `dada2_learn_errors`, I would go for changing both.

Maybe you can start a final dada2 cleanup PR that will only be finalised and merged once all the individual PRs are merged? This way we can collect changes to already merged PRs in one place and reduce the number PRs we need to track.

bio/dada2/sample-inference/meta.yaml
dlaehnemann: So `sample_inference` [works here](https://benjjneb.github.io/dada2/tutorial.html#sample-inference), right?

You might just want to double-check in the end. I changed the name of wrapper (6/10), that also had `dada2_learn_errors` to `dada2_merge_pairs` and you might want to go for `dada2_merge_paired_reads`, [according to the tutorial](https://benjjneb.github.io/dada2/tutorial.html#merge-paired-reads). So maybe something for the cleanup PR, as well (see below).

bio/dada2/sample-inference/test/Snakefile
cpauvert: Yeah sorry about that. Perfect! I like the idea of a final cleanup PR!

dlaehnemann: No worries. That's exactly what all these code reviews are for.

dlaehnemann: Also, one more thought for the cleanup: It seems like you can link to sections of the tutorial, by appending `#all-lower-caps-section-title-with-dashes-for-whitespace`. So with the names already corresponding to the tutorial section, it might be nice to adjust the links at the start of the description to point to the respective tutorial sections. E.g.:
https://benjjneb.github.io/dada2/tutorial.html#merge-paired-reads

What do you think?

cpauvert: Hum nice suggestion but the manual provide the additional arguments whereas the tutorial doesn't. However I found out that I could pinpoint to the function within the manual using this:


https://www.bioconductor.org/packages/release/bioc/manuals/dada2/man/dada2.pdf#derepFastq

We could keep the link to the tutorial for the meta-wrapper then.
What do you think?

dlaehnemann: Yes, you're right. For the optional arguments, we should point to the PDF manual.

So maybe point to both the PDF manual and the tutorial, simply adding a sentence about the tutorial?

The PDF-link doesn't seem to work for me. It opens the PDF, but doesn't jump to the function's help section. So maybe it's browser-specific (I'm using Firefox)

bio/dada2/sample-inference/meta.yaml
cpauvert: Done in the last commit! Thanks

bio/dada2/sample-inference/test/Snakefile
cpauvert: Ok too bad. Yes I can add both links then ;)

dlaehnemann (approved):
Looking good!

PR #218, opened 2020-10-13
Added dada2 wrapper (2.A/10) for filterAndTrim (se version)

cpauvert:
Hi,
Yet another wrapper for dada2, this one is meant to filter the single-end reads. 
Best,

PR #220, opened 2020-10-14
added dada2 wrapper (2/10) for filterAndTrim (se/pe versions)

cpauvert:
Hi, 
This is a proposed wrapper for dada2. It builds upon a previous (closed) work of @jafors and me (#207) but this one is more integrated with the others wrappers, especially in terms of dealing with paired-end sequences.
I hope it will be useful.
Best,


cpauvert:
@dlaehnemann 
Included the *new* way to pass additional arguments (using python named arguments and `do.call()`).
I realize that one wrapper was sufficient for pe and se, but two rules in the Snakefile were necessary (just like we did in #209).
Once merged, I shall close the #218 
Best, 
Charlie

cpauvert:
This wrapper is ready @dlaehnemann ;)
If you can have a go at it, that would be nice!
Charlie

bio/dada2/filter-trim/test/Snakefile
```diff
@@ -0,0 +1,45 @@
...
+        "master/bio/dada2/filter-trim"
+
+rule dada2_filter_trim_pe:
+    input:
+        # Paired-end files without primers sequences
+        fwd="trimmed/{sample}.1.fastq",
+        rev="trimmed/{sample}.2.fastq"
+    output:
+        filt="filtered-pe/{sample}.1.fastq",
+        filt_rev="filtered-pe{sample}.2.fastq",
```
dlaehnemann, line 30: I think this is missing a `/`:
```suggestion
        filt_rev="filtered-pe/{sample}.2.fastq",
```

bio/dada2/filter-trim/test/Snakefile
```diff
@@ -0,0 +1,45 @@
...
+        "logs/dada2/filter-trim-se/{sample}.log"
+    # Even though this is an R wrapper, use named arguments in Python syntax
+    # here, to specify extra parameters. Python booleans (`arg1=True`, `arg2=False`)
+    # and lists (`list_arg=[]`) are automatically converted to R.
+    # For a named list as an extra named arguments, use a python dict
+    #   (`named_list={name1=arg1}`).
+    params:
+        maxEE=1, truncLen=7, minLen=1
+        # Set the number of kept bases to 7 for the toy example
+        # Set minLen to 1 for the toy example but default is 20
```
dlaehnemann, line 18: I'd put one parameter per line and then put the comments right above:
```suggestion
        maxEE=1,
        # Set the number of kept bases to 7 for the toy example
        truncLen=7,
        # Set minLen to 1 for the toy example but default is 20
        minLen=1
```

bio/dada2/filter-trim/test/Snakefile
```diff
@@ -0,0 +1,45 @@
...
+        "logs/dada2/filter-trim-pe/{sample}.log"
+    # Even though this is an R wrapper, use named arguments in Python syntax
+    # here, to specify extra parameters. Python booleans (`arg1=True`, `arg2=False`)
+    # and lists (`list_arg=[]`) are automatically converted to R.
+    # For a named list as an extra named arguments, use a python dict
+    #   (`named_list={name1=arg1}`).
+    params:
+        maxEE=1, truncLen=[7,6], minLen=1
+        # Set the number of kept bases to 7 for the toy example
+        # Set minLen to 1 for the toy example but default is 20
```
dlaehnemann, line 42: As above, but also explaining what the list `[7,6]` here means?

bio/dada2/filter-trim/meta.yaml
```diff
@@ -0,0 +1,15 @@
+name: dada2_filter_trim
+description: |
+  `DADA2 <https://benjjneb.github.io/dada2/index.html>`_
+  Quality filtering of single or paired-end reads using dada2 filterAndTrim function. Optional parameters are documented in the `manual <https://www.bioconductor.org/packages/release/bioc/manuals/dada2/man/dada2.pdf>`_.
+authors:
+  - Charlie Pauvert
+input:
+    fwd: a forward FASTQ file (potentially compressed) without primers sequences
+    rev: a (facultative) reverse FASTQ file (potentially compressed) without primers sequences
```
dlaehnemann, line 9: Minor preference of word choice in English (we also have `fakultativ` in German, so I see where this word choice is coming from;):
```suggestion
    rev: an (optional) reverse FASTQ file (potentially compressed) without primers sequences
```

bio/dada2/filter-trim/meta.yaml
```diff
@@ -0,0 +1,15 @@
...
+output:
+    filt: a compressed filtered forward FASTQ file
+    filt_rev: a (facultative) compressed filtered reverse FASTQ file
```
dlaehnemann, line 12:
```suggestion
    filt_rev: an (optional) compressed filtered reverse FASTQ file
```

bio/dada2/filter-trim/wrapper.R
```diff
@@ -0,0 +1,54 @@
...
+        # Add them to the list of arguments
+        args<-c(args, extra)
+    } else{
+        message("Optional R parameters should be passed as named Python arguments")
+        message("in the Snakefile. Check the example below:")
+        message("params:\n\tverbose=True, foo=[1,42]")
+        message("Using defaults parameters from dada2::filterAndTrim()")
+    }
+} else{
+    message("No optional parameters. Using defaults parameters from dada2::filterAndTrim()")
```
dlaehnemann, line 43: Typo:
```suggestion
    message("No optional parameters. Using default parameters from dada2::filterAndTrim()")
```

test.py
```diff
@@ -113,6 +113,16 @@ def test_dada2_learn_errors():
...
+def test_dada2_filter_trim_se():
+    run(
+        "bio/dada2/filter-trim",
+        ["snakemake", "--cores", "1", "filtered-se/a.1.fastq", "--use-conda", "-F"]
+    )
+def test_dada2_filter_trim_pe():
+    run(
+        "bio/dada2/filter-trim",
+        ["snakemake", "--cores", "1", "filtered-pe/a.1.fastq", "--use-conda", "-F"]
+    )
```
dlaehnemann, line 125: I guess it's mostly for readability, but usually you would add an empty row before and after each function definition in Python (and this might incidentally help with the merge conflicts, although I'm not sure):
```suggestion

def test_dada2_filter_trim_se():
    run(
        "bio/dada2/filter-trim",
        ["snakemake", "--cores", "1", "filtered-se/a.1.fastq", "--use-conda", "-F"]
    )

def test_dada2_filter_trim_pe():
    run(
        "bio/dada2/filter-trim",
        ["snakemake", "--cores", "1", "filtered-pe/a.1.fastq", "--use-conda", "-F"]
    )

```

dlaehnemann (commented):
Looks good. Some minor suggestions and typos, most of which you should be able to simply commit by using the suggestions right here.

bio/dada2/filter-trim/meta.yaml
cpauvert: Haha I'm busted!

bio/dada2/filter-trim/test/Snakefile
cpauvert: Thanks

bio/dada2/filter-trim/wrapper.R
cpauvert: Good catch! Oops that one is gonna come frequently in the others then

bio/dada2/filter-trim/test/Snakefile
cpauvert: Done in the comment :)

bio/dada2/filter-trim/meta.yaml
```diff
@@ -0,0 +1,15 @@
...
+description: |
+  `DADA2 <https://benjjneb.github.io/dada2/index.html>`_
+  Quality filtering of single or paired-end reads using dada2 filterAndTrim function. Optional parameters are documented in the `manual <https://www.bioconductor.org/packages/release/bioc/manuals/dada2/man/dada2.pdf>`_.
+authors:
+  - Charlie Pauvert
+input:
+    fwd: a forward FASTQ file (potentially compressed) without primers sequences
+    rev: an (optional) reverse FASTQ file (potentially compressed) without primers sequences
+output:
+    filt: a compressed filtered forward FASTQ file
```
dlaehnemann, line 11: The output files in the example rules are simply `.fastq` and thus uncompressed. Does this dada2 function recognize the file extension and compresses or doesn't accordingly? If so this should be mentioned here. And maybe one of the example rules (either se or pe) could showcase compressed output?

bio/dada2/filter-trim/meta.yaml
```diff
@@ -0,0 +1,15 @@
+name: dada2_filter_trim
+description: |
+  `DADA2 <https://benjjneb.github.io/dada2/index.html>`_
+  Quality filtering of single or paired-end reads using dada2 filterAndTrim function. Optional parameters are documented in the `manual <https://www.bioconductor.org/packages/release/bioc/manuals/dada2/man/dada2.pdf>`_.
+authors:
+  - Charlie Pauvert
+input:
+    fwd: a forward FASTQ file (potentially compressed) without primers sequences
```
dlaehnemann, line 8: `primer sequences` only needs the s in sequences (don't ask me to explain why...;)

bio/dada2/filter-trim/test/Snakefile
```diff
@@ -0,0 +1,52 @@
...
+        fwd="trimmed/{sample}.1.fastq"
+    output:
+        filt="filtered-se/{sample}.1.fastq",
+        stats="reports/dada2/filter-trim-se/{sample}.tsv"
+    log:
+        "logs/dada2/filter-trim-se/{sample}.log"
+    # Even though this is an R wrapper, use named arguments in Python syntax
+    # here, to specify extra parameters. Python booleans (`arg1=True`, `arg2=False`)
+    # and lists (`list_arg=[]`) are automatically converted to R.
+    # For a named list as an extra named arguments, use a python dict
```
dlaehnemann, line 13: `argument` without the `s`. Sorry for this typo.

bio/dada2/filter-trim/wrapper.R
```diff
@@ -0,0 +1,54 @@
...
+)
+# Test if paired end input is passed
+if(!is.null(snakemake@input[["rev"]]) & !is.null(snakemake@output[["filt_rev"]])){
+        args<-c(args,
+            rev = snakemake@input[["rev"]],
+            filt.rev = snakemake@output[["filt_rev"]]
+            )
+}
+# Check if extra params are passed
+if(length(snakemake@params) > 0 ){
```
dlaehnemann, line 30: Is this check even necessary, as you do the `is.list()` check below, anyways? So maybe we can remove this outer checking level for leaner code?

dlaehnemann (commented):
Some typos and two questions. Getting there... :D

bio/dada2/filter-trim/meta.yaml
cpauvert: > The output files in the example rules are simply `.fastq` and thus uncompressed. Does this dada2 function recognize the file extension and compresses or doesn't accordingly? If so this should be mentioned here.

Exactly! I will add a bit on that.

>  And maybe one of the example rules (either se or pe) could showcase compressed output?

Ok!

cpauvert: Actually after checking, the dada2 function compress the output by default unless you set `compress = FALSE`, so I will rename the output files accordingly.

bio/dada2/filter-trim/wrapper.R
cpauvert: Actually this is the opposite. Snakemake initializes a `list()` for `params:` by default ([link](https://snakemake.readthedocs.io/en/stable/_modules/snakemake/script.html#RScript)) so my `is.list()` is irrelevant. But checking whether the list is empty or not indicates to the user and the log that the function runs on the default parameters. Will clean now, thanks

bio/dada2/filter-trim/test/Snakefile
```diff
@@ -0,0 +1,52 @@
+rule dada2_filter_trim_se:
+    input:
+        # Single-end files without primers sequences
+        fwd="trimmed/{sample}.1.fastq"
```
dlaehnemann, line 4: One last thing: Maybe make this a compressed `fastq` file, so that compressed input is showcased? And adding a compressed version `a.1.fastq.gz` will not really add much space usage, so should be absolutely fine.

bio/dada2/filter-trim/meta.yaml
dlaehnemann: Ah, then we should add one bit of automagic behaviour:
If the output extension is `.fastq.gz`, the default compression should happen. If the output extension is `.fastq`, the extra option `compress=FALSE` should be set in `wrapper.R`. And if `params` contains `compress=FALSE`, an error should be thrown that the user should specify this only via the output's extension. Wherever possible, we try to implement implicit specifications of parameters via the `input` and `output` file extensions and via `wildcards`.

bio/dada2/filter-trim/test/Snakefile
cpauvert: Done!

bio/dada2/filter-trim/meta.yaml
cpauvert: Ok I understand. Should this automagic lies in the `Snakefile` or in the `wrapper.R` ? If you have any examples (even Python) it could help :)

dlaehnemann: This is usually in the `wrapper.py` / `wrapper.R`. Let me find an example...

dlaehnemann: A very involved example is this one:
https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/salmon/quant.html#code

But your case should be much easier. First check, whether `extra` contains `compress`, and if so throw an error. and then simply check in `wrapper.R` if the last extension in the `output` is `.gz` and if not, add `compress=FALSE` to the arguments you pass via `do.call()`.

cpauvert: Thanks for the tips! The last commit a34d377 does it

bio/dada2/filter-trim/wrapper.R
```diff
@@ -0,0 +1,76 @@
...
+log.file<-file(snakemake@log[[1]],open="wt")
+sink(log.file)
+sink(log.file,type="message")
+
+library(dada2)
+
+# Custom function to check if homogeneous compression suffixes are given
+all_compressed<-function(f){
+	if(!is.null(f)){
+		compressed<-grepl(".gz$|.bz2$", f)
```
dlaehnemann, line 19: From the docs of `filterAndTrim()` I understand, that compression is only possible in `gzip` format. So the only thing to check for should be `.gz`. And with the [`endsWith()` function](https://stat.ethz.ch/R-manual/R-devel/library/base/html/startsWith.html), we might not even need the `all_compressed()` helper function, but can handle everything with the `all()` call below.

cpauvert: Hum, you're right for `gzip` I inverted with the possibe input. And also for the `endsWith()` that's nice.
But I devised the `all_compressed` function, in case not all files provided by the user were consistent. But perhaps that's not an expected behavior. What do you think?

bio/dada2/filter-trim/wrapper.R
```diff
@@ -0,0 +1,76 @@
...
+    stop("Remove the `compress` option from the Snakefile.\n",
+    "The `compress` option is implicitly set here from the file extension.")
+    } else {
+    # Check if output files are given as compressed files
+    # ex: in se version, all(TRUE, NULL) gives TRUE
+    extra[["compress"]]<-all(
+    			all_compressed(args[["filt"]]),
+    			all_compressed(args[["filt.rev"]])
+    			)
+    }
```
dlaehnemann, line 61: I just realized, that it is actually OK to have different `compress` status for the two paired-end output files, according to the docs. I don't know why a user would want that, but no need to force the user to avoid something that the software permits. So the following should do the trick:
```suggestion
    extra[["compress"]]<-c(
        endsWith(args[["filt"]], '.gz'),
        endsWith(args[["filt.rev"]], 'gz')
    )
```

The only case I am not sure about is the single-end case, because then `compress` is a two-entry vector. This could be easier, if we remove the explicit naming of inputs and simply check whether we have one or two inputs to decide whether it is single end or paired end input. So something like:
```
input:
    "{sample}.1.fastq.gz",
    "{sample}.1.fastq.gz"
output:
    filtered=["filtered/{sample}.1.fastq.gz", "filtered/{sample}.2.fastq.gz"],
    stats="reports/{sample}.tsv"
```
And then check the length of `snakemake@input` to decide whether only `fwd` is assigned to, or if a second entry is available for `rev`.

dlaehnemann: In the latter case, we can simply use `extra[["compress"]]<-lapply(snakemake@output[["filtered"]], endsWith, suffix='.gz')` to get a list of appropriate length.

cpauvert: Ok for your suggestion! But no need to remove the explicit naming.
Actually we can exploit the `NULL` property!
For instance: `c(TRUE, NULL)` gives `c(TRUE)` in R!
Meaning that combining this magic and yours we get something like this:
```r
extra[["compress"]]<-c(
        endsWith(args[["filt"]], '.gz'),
        if(is.null(args[["filt.rev"]])) NULL else {endsWith(args[["filt.rev"]], 'gz')}
    )
```

cpauvert: Done in the last commit!

bio/dada2/filter-trim/wrapper.R
```diff
@@ -0,0 +1,59 @@
...
+            filt.rev = snakemake@output[["filt_rev"]]
+            )
+}
+# Check if extra params are passed
+if(length(snakemake@params) > 0 ){
+    # Keeping only the named elements of the list for do.call()
+    extra<-snakemake@params[ names(snakemake@params) != "" ]
+# Check if 'compress=' option is passed
+if(!is.null(extra[["compress"]])){
+    stop("Remove the `compress` option from the Snakefile.\n",
```
dlaehnemann, line 35:
```suggestion
    stop("Remove the `compress=` option from `params`.\n",
```

bio/dada2/filter-trim/wrapper.R
```diff
@@ -0,0 +1,59 @@
...
+    # Check if output files are given as compressed files
+    # ex: in se version, all(TRUE, NULL) gives TRUE
+    extra[["compress"]]<-c(
+	   endsWith(args[["filt"]], '.gz'),
+	   if(is.null(args[["filt.rev"]])) NULL else {endsWith(args[["filt.rev"]], 'gz')}
+       )
+    }
+    # Add them to the list of arguments
+    args<-c(args, extra)
+} else{
```
dlaehnemann, line 47:
```suggestion
    # Check if 'compress=' option is passed
    if(!is.null(extra[["compress"]])){
        stop("Remove the `compress=` option from `params`.\n",
            "The `compress` option is implicitly set here from the file extension.")
    } else {
        # Check if output files are given as compressed files
        # ex: in se version, all(TRUE, NULL) gives TRUE
        extra[["compress"]]<-c(
	      endsWith(args[["filt"]], '.gz'),
	      if(is.null(args[["filt.rev"]])) NULL else {endsWith(args[["filt.rev"]], 'gz')}
        )
    }
    # Add them to the list of arguments
    args<-c(args, extra)
} else {
```

dlaehnemann (approved):
This looks good. I adjusted two minor things and will wait for the tests to pass. Then this is ready to merge! :tada:

cpauvert:
Actually the failed test is due to `dada2`:
```
$ cat logs/dada2/filter-trim-pe/a.log
Le chargement a nécessité le package : Rcpp
Error in writeFastq(fqF, fout[[1]], "w", compress = compress) : 
  'compress' must be 'logical(1) (TRUE or FALSE)'
Calls: do.call ... mcmapply -> .mapply -> <Anonymous> -> writeFastq -> writeFastq
Exécution arrêtée
```
It complains about `compress` being of length two, but according to the manual this is possible. **But** when running the package example():

```r
library(dada2)
example(filterAndTrim)
filterAndTrim(testFastqs, filtFastqs, maxN=0, maxEE=2, verbose=TRUE,compress=c(T,F))
```

I got the same error :

```
Overwriting file:/tmp/RtmpwbRA08/file31045a6753e.fastq.gz
Error in writeFastq(fq, fout, "w", compress = compress) : 
  'compress' must be 'logical(1) (TRUE or FALSE)'
```
So perhaps we could still use the `all()` as in [here](https://github.com/snakemake/snakemake-wrappers/pull/220#discussion_r514294153) in the meantime. I would raise an issue on the repository of dada2.

bio/dada2/filter-trim/wrapper.R
```diff
@@ -0,0 +1,59 @@
...
+    if(!is.null(extra[["compress"]])){
+        stop("Remove the `compress=` option from `params`.\n",
+            "The `compress` option is implicitly set here from the file extension.")
+    } else {
+        # Check if output files are given as compressed files
+        # ex: in se version, all(TRUE, NULL) gives TRUE
+        extra[["compress"]]<-all(
+	      endsWith(args[["filt"]], '.gz'),
+	      if(is.null(args[["filt.rev"]])) NULL else {endsWith(args[["filt.rev"]], 'gz')}
+        )
```
dlaehnemann, line 43: Thanks for tracking down the test-failing behaviour. With the new information that different values are not possible for forward and reverse, I would opt for checking for consistency with something like this:
```suggestion
        compressed <- c(
            endsWith(args[["filt"]], '.gz'),
            if(is.null(args[["filt.rev"]])) NULL else {endsWith(args[["filt.rev"]], 'gz')}
        )
        if ( all(compressed) ) {
            extra[["compress"]] <- TRUE
        } else if ( any(compressed) ) {
            stop("Either all or no fastq output should be compressed. Please check `output.filt` and `output.filt_rev` for consistency.")
        }
```

cpauvert: Yes this makes sense!

bio/dada2/filter-trim/wrapper.R
```diff
@@ -0,0 +1,64 @@
...
+        # ex: in se version, all(TRUE, NULL) gives TRUE
+        compressed <- c(
+            endsWith(args[["filt"]], '.gz'),
+            if(is.null(args[["filt.rev"]])) NULL else {endsWith(args[["filt.rev"]], 'gz')}
+        )
+        if ( all(compressed) ) {
+            extra[["compress"]] <- TRUE
+        } else if ( any(compressed) ) {
+            stop("Either all or no fastq output should be compressed. Please check `output.filt` and `output.filt_rev` for consistency.")
+        }
```
cpauvert, line 48:
```suggestion
        if ( all(compressed) ) {
            extra[["compress"]] <- TRUE
        } else if ( any(compressed) ) {
            stop("Either all or no fastq output should be compressed. Please check `output.filt` and `output.filt_rev` for consistency.")
        } else {
            extra[["compress"]] <- FALSE
        }
```

For consistency, I added the all `FALSE`

dlaehnemann (approved):
Looks like we made it! :D

bio/dada2/filter-trim/wrapper.R
dlaehnemann: This is actually important, as the default behaviour is compressed output. So this is the one case, where we actually have to explicitly provide this named argument. So +1 for consistency! :D

cpauvert:
Yes! Thank you very much for your inputs @dlaehnemann ! I updated the docs syntax for the others as well. Do not hesitate to spark a chat if something needs more clarification!
Best, 
Charlie

PR #221, opened 2020-10-14
Wrapper for bowtie2-build

standage:
The `bowtie2` aligner is already included in Snakemake wrappers, but this PR adds a wrapper for the `bowtie2-build` command.


standage:
Ready for review!

bio/bowtie2/build/environment.yaml
```diff
@@ -0,0 +1,7 @@
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
+dependencies:
+  - bowtie2 ==2.4.1
```
dlaehnemann, line 6: Maybe unpin the patch version, unless this particular patch version is needed (e.g. for a version-specific index or something?)?
```suggestion
  - bowtie2 =2.4
```

bio/bowtie2/build/wrapper.py
```diff
@@ -0,0 +1,14 @@
...
+__license__ = "MIT"
+
+
+from snakemake.shell import shell
+
+extra = snakemake.params.get("extra", "")
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+shell(
+    "bowtie2-build --threads {snakemake.threads} {snakemake.params.extra} "
+    "{snakemake.input[0]} {snakemake.params.base}"
```
dlaehnemann, line 13: You could refer directly to the reference input by name:
```suggestion
    "{snakemake.input.reference} {snakemake.params.base}"
```

bio/bowtie2/build/test/Snakefile
```diff
@@ -0,0 +1,16 @@
+rule bowtie2_build:
+    input:
+        reference="genome.fasta"
+    output:
+        expand(
+            "genome.{ext}",
+            ext=("1.bt2", "2.bt2", "3.bt2", "4.bt2", "rev.1.bt2", "rev.2.bt2"),
+        ),
+    log:
```
dlaehnemann, line 9: I would say this calls for the use of the [snakemake `multiext()` function](https://snakemake.readthedocs.io/en/latest/snakefiles/rules.html#the-multiext-function).

bio/bowtie2/build/test/Snakefile
```diff
@@ -0,0 +1,16 @@
...
+        "logs/bowtie2_build/build.log"
+    params:
+        base="genome",  # prefix for index files
```
dlaehnemann, line 12: Instead of specifying the same basename as in `output` here via `params`, I would rather parse `snakemake.output[0]` to strip off the specific extension in `wrapper.py` and then pass the resulting `base` to bowtie. This way there's no risk that the `output` spec and a separate `base` spec can be different and break the wrapper.

dlaehnemann (changes requested):
Thanks for adding this missing wrapper! It generally looks good, I just have a couple of suggestions to further improve it.

bio/bowtie2/build/environment.yaml
standage: I'd be happy to do so, but the bowtie2 align wrapper pins this particular version. In any case, I suggest that these two wrappers should be consistent.

bio/bowtie2/build/test/Snakefile
standage: Done in 5d33a4b.

bio/bowtie2/build/test/Snakefile
standage: Is the change in 5d33a4b what you had in mind?

bio/bowtie2/build/wrapper.py
standage: Done in 5d33a4b.

bio/bowtie2/build/environment.yaml
dlaehnemann: Jep, that sounds perfectly reasonable. So lets keep it pinned as is and these should then be updated in sync. Maybe we can add a comment here (and correspondingly in the `align` wrapper's `environment.yaml`), noting this -- thus whenever anyone does an update, they are reminded to do it in sync:
```suggestion
  # if you update this version, please also update the version at bio/bowtie2/align/environment.yaml
  - bowtie2 ==2.4.1
```

bio/bowtie2/build/test/Snakefile
dlaehnemann: Jup, exactly the right amount of automagic behaviour! :D

bio/bowtie2/build/environment.yaml
standage: Done in 8fd850d.

dlaehnemann (approved):
Looks good. Thanks again!

PR #222, opened 2020-10-14
Ensure pigz dependency is installed

standage:
The `seqtk/subsample` wrapper didn't work for me because I didn't have pigz installed in my environment. This PR updates the corresponding env files to make sure this doesn't happen.

### DIFF
--- a/bio/seqtk/subsample/pe/environment.yaml
+++ b/bio/seqtk/subsample/pe/environment.yaml
@@ -3,3 +3,4 @@ channels:
   - conda-forge
 dependencies:
   - seqtk ==1.3
+  - pigz ==2.3.4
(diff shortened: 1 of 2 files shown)


bio/seqtk/subsample/pe/environment.yaml
```diff
@@ -3,3 +3,4 @@ channels:
   - conda-forge
 dependencies:
   - seqtk ==1.3
+  - pigz ==2.3.4
```
dlaehnemann, line 6: Maybe unpin the patch version, to allow using newer patch versions that might have bugs removed or even security problems patched. Or is there a particular need for this exact version?
```suggestion
  - pigz =2.3
```
Accordingly, for the `se` wrapper.

standage: Nope, minor version should suffice.

standage: Done in c62ca19.

dlaehnemann (approved):
Looks good, thanks!

PR #225, opened 2020-10-15
added dada2 wrapper (8/10) for removeBimera function

cpauvert:
Hi,
Yet another dada2 wrapper suggested. Additional arguments to the R function can be provided through a Python dictionary the params rule. Reviews and suggestions are obviously welcome :)
Best,


bio/dada2/remove-chimeras/meta.yaml
```diff
@@ -0,0 +1,12 @@
+name: dada2_remove_chimeras
+description: |
+  `DADA2 <https://benjjneb.github.io/dada2/index.html>`_
+  Removing chimeras sequences from the sequence table data using dada2 removeBimeraDenovo function. Optional parameters are documented in the `manual <https://www.bioconductor.org/packages/release/bioc/manuals/dada2/man/dada2.pdf>`_.
```
dlaehnemann, line 4:
```suggestion
  Remove chimera sequences from the sequence table data using dada2 removeBimeraDenovo function. Optional parameters are documented in the `manual <https://www.bioconductor.org/packages/release/bioc/manuals/dada2/man/dada2.pdf>`_.
```

bio/dada2/remove-chimeras/wrapper.R
```diff
@@ -0,0 +1,41 @@
...
+       args<-c(args, extra)
+} else{
+    message("No optional parameters. Using default parameters from dada2::removeBimeraDenovo()")
+}
+
+# Remove chimeras
+seqTab.nochim<-do.call(removeBimeraDenovo, args)
+
+# Store the estimated errors as RDS files
+saveRDS(seqTab.nochim, snakemake@output[[1]],compress = T)
```
dlaehnemann, line 36: I usually try to avoid `.` in variable names, even though this seems to be common in classic R code. But in many other languages, this syntax is reserved for referencing substructure of a data structure or for calling functions on a variable. So I would usually opt for an underscore instead. This is, however, merely a question of preference -- so feel free to ignore:
```suggestion
# Remove chimeras
seqTab_nochimeras<-do.call(removeBimeraDenovo, args)

# Store the estimated errors as RDS files
saveRDS(seqTab_nochimeras, snakemake@output[[1]],compress = T)
```
Also, while I'm at it, why not make the variable even more readable by skipping the `chim` abbreviation...

dlaehnemann (approved):
Generally looks good, two language / stylistic suggestions.

bio/dada2/remove-chimeras/meta.yaml
cpauvert: Good catch!

bio/dada2/remove-chimeras/wrapper.R
cpauvert: Yes I don't mind to change. I was sticking to the tutorial. Will change

cpauvert:
Thanks!

PR #227, opened 2020-10-15
added dada2 wrapper (10/10) for addSpecies function

cpauvert:
Hi,
The last dada2 wrapper of the series. Additional arguments to the R function can still be provided through a Python dictionary the params rule. Reviews and suggestions are obviously welcome :)
Best,


bio/dada2/add-species/test/Snakefile
```diff
@@ -0,0 +1,18 @@
...
+    input:
+        taxtab="results/dada2/taxa.RDS", # Taxonomic assignments
+        refFasta="resources/example_species_assignment.fa.gz" # Reference FASTA 
+    # Even though this is an R wrapper, use named arguments in Python syntax
+    # here, to specify extra parameters. Python booleans (`arg1=True`, `arg2=False`)
+    # and lists (`list_arg=[]`) are automatically converted to R.
+    # For a named list as an extra named argument, use a python dict
+    #   (`named_list={name1=arg1}`).
+    params:
+        verbose=True
```
dlaehnemann, line 11: I like having an example `key=value` pair here, to showcase how to specify named parameters. But maybe keep this as commented out per default, as you did in other wrappers? I am not sure, how output this will generate for the log file?
```suggestion
    #params:
    #    verbose=True
```

dlaehnemann: But maybe also something to do in the final cleanup PR, simply checking if all the wrappers are consistent in that regard.

dlaehnemann (approved):
Just one minor thing, but we could also leave this for a final dada2 cleanup PR.

bio/dada2/add-species/test/Snakefile
cpauvert: Yes we can comment here, this is no big deal for the log. I will check in the final PR.

cpauvert:
Resolved the `test.py` conflict. I will reorder and tidy the dada2 tests function in the clean PR

cpauvert:
Thanks!

PR #230, opened 2020-10-16
added wrapper (1/10) for dada2 plotQualityProfile

cpauvert:
Hi,
Here is a wrapper for examining quality profiles of a single FASTQ sample file (in **s**ingle **e**nd version) or the list of two samples (in **p**aired **e**nd version).
Reviews and contributions are welcome,
Best, 
Charlie


bio/dada2/quality-profile/environment.yaml
```diff
@@ -0,0 +1,7 @@
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
+dependencies:
+  - bioconductor-dada2 == 1.16.0
```
dlaehnemann, line 6: As mentioned in the other review. Unless it is necessary to pin the exact patch version, I'd unpin it:
```suggestion
  - bioconductor-dada2 == 1.16
```

bio/dada2/quality-profile/test/Snakefile
```diff
@@ -0,0 +1,10 @@
+rule dada2_quality_profile:
+    input:
+        # FASTQ file without primers sequences
+        "trimmed/{sample}.{orientation}.fastq"                               # se version
+        # expand("trimmed/{{sample}}.{orientation}.fastq",orientation=[1,2]) # pe version
+    output:
+        "reports/dada2/quality-profile/{sample}.{orientation}-quality-profile.png" # se version
+        #"reports/dada2/quality-profile/{sample}-quality-profile.png"              # pe version
+    wrapper:
+        "master/bio/dada2/quality-profile"
```
dlaehnemann, line 10: I would simply create two example rules to showcase both the single end and the paired end case (and then test both in `test.py`). For example:
```suggestion
rule dada2_quality_profile_se:
    input:
        # FASTQ file without primers sequences
        "trimmed/{sample}.1.fastq"
    output:
        "reports/dada2/quality-profile/{sample}.1-quality-profile.png"
    wrapper:
        "master/bio/dada2/quality-profile"

rule dada2_quality_profile_pe:
    input:
        # FASTQ file without primers sequences
        expand("trimmed/{{sample}}.{orientation}.fastq",orientation=[1,2])
    output:
        "reports/dada2/quality-profile/{sample}-quality-profile.png"
    wrapper:
        "master/bio/dada2/quality-profile"
```

bio/dada2/quality-profile/wrapper.R
```diff
@@ -0,0 +1,15 @@
+# __author__ = "Charlie Pauvert"
+# __copyright__ = "Copyright 2020, Charlie Pauvert"
+# __email__ = "cpauvert@protonmail.com"
+# __license__ = "MIT"
+
+# Snakemake wrapper for plotting the quality profile of reads using dada2 plotQualityProfile function.
+
+library(dada2, quietly=TRUE)
```
dlaehnemann, line 8: I'd add the logging redirection / sink code here (and add a `log:` entry to the example rules) and then remove the `quietly=TRUE` to get library loading infos into the logs. This may be important for debugging installation / loading issues.

dlaehnemann (changes requested):
Nice little recipe, pretty straightforward. I have two suggestions.

Thanks a bunch!

bio/dada2/quality-profile/wrapper.R
cpauvert: Thanks! Fixed in eef6c31

bio/dada2/quality-profile/test/Snakefile
cpauvert: That is better indeed. Done in eef6c31

bio/dada2/quality-profile/environment.yaml
cpauvert: Done in eef6c31

bio/dada2/quality-profile/wrapper.R
```diff
@@ -0,0 +1,24 @@
...
+# Plot the quality profile for a given FASTQ file or a list of files
+pquality<-plotQualityProfile(unlist(snakemake@input))
+
+# Write the plots to files
+library(ggplot2)
+ggsave(snakemake@output[[1]], pquality, width = 4, height = 3, dpi = 300)
+
+# Close the connection for the log file
+sink(type="message")
+sink()
```
dlaehnemann, line 24: Out of curiosity (and because we have not been doing this in other snakemake R wrappers / R scripts): Do you think this is necessary or do you think this could be skipped? I would think, this implicitly happens with the script ending, right? So unless you want to access the log file from within the script after closing this sink, you should not need to do this explicitly? But then, I might be missing some important implication.

dlaehnemann (approved):
Looks good. I have one thing, but it is more of a question (with possible implications for other snakemake R wrappers/scripts).

bio/dada2/quality-profile/wrapper.R
cpauvert: I've never thought of it. But I think you're right it could be skipped. I tried on a small reprex and it seems right. Even in the other  dada2 wrappers, it is either a per sample basis or based on all samples, so the log is close by the end of the script and it is fine.

dlaehnemann: Yeah, the log should always be unique to a rule's instance via the wildcards (snakemake complains, if the `log` file specification doesn't contain all of the wildcards from `output`).
In any case, I'd say it's your call whether you want to keep it around (if only to demonstrate correct usage of `sink()`, which I have just learned this way;) or want to kick it. Just let me know which way, and I can either merge as is or after a further commit.

cpauvert: Ok let's keep the proper syntax them. Then I will also add the source from where I learned it as well ;)

cpauvert:
That is alright for me then :+1:

cpauvert:
Thank you David, I will apply the discussed changes and parameters implementation in the others wrappers to be homogeneous!

dlaehnemann:
Feel free to ping me in for reviews, once you are done updating each one of them. This way we should get through them quickly!

Issue #228, opened 2020-10-15
bwa and bwa-mem2 index could infer prefix from output


fgvieira:
**Is your feature request related to a problem? Please describe.**
At least in `bwa` and `bwa-mem2` index, the destination (`prefix`) could be directly inferred from the output, being unnecessary to specify the `prefix` param.

**Describe the solution you'd like**
The wrapper would infer the prefix from `snakemake.output[0]`

@Smeds do you think it makes sense?

fgvieira:
Fixed in #232. Closing.

PR #232, opened 2020-10-19
feat: explicitly specify bwa index in bwa wrappers

tdayris:
Hi,

This is an implementation of enhancement request #228 

Many thanks in advance for your reviews

PS: I'm still having this Template merge from PR #187 . I shall not have merged it to my master branch


bio/bwa-mem2/index/wrapper.py
```diff
@@ -16,7 +16,11 @@
     raise ValueError("Please provide exactly one reference genome as input.")
 
 # Prefix that should be used for the database
-prefix = snakemake.params.get("prefix", "")
+prefix = None
+if "prefix" in snakemake.params.keys():
+    prefix = snakemake.params["prefix"]
+else:
+    prefix = splitext(snakemake.output[0])[0]
```
dlaehnemann, line 23: In correspondence with this change, I would also remove the `params.prefix` specification from the example `Snakefile` here:
https://github.com/tdayris/snakemake-wrappers/blob/661f288e5f1d150c35ca8e910c5a3fb59ef9c4a4/bio/bwa-mem2/index/test/Snakefile#L13-L14

Same for `prefix` and `index` in the other recipes. I like the idea of backward compatibility of the wrapper code, in case someone simply updates the wrapper version. But new users that copy paste the example rule from the Snakefile should have the automagic behaviour without risking that the prefix and output specs go out of sync.

bio/bwa-mem2/mem-samblaster/wrapper.py
```diff
@@ -14,6 +14,14 @@
 sort_extra = snakemake.params.get("sort_extra", "")
 samblaster_extra = snakemake.params.get("samblaster_extra", "")
 
+index = None
+if "index" in snakemake.params.keys():
+    index = snakemake.params["index"]
+elif "index" in snakemake.input.keys():
+    index = path.splitext(snakemake.input["index"])[0]
```
dlaehnemann, line 21: I guess this is intended as the standard future use, right? I would then update the example rule in the Snakefile to contain an `input: index="someindex.some_index_extension"` accordingly.

dlaehnemann (changes requested):
I really like the new behaviour and the backwards compatibility. I have some suggestions to update the example rules in the `Snakefile`s accordingly. But I also just realised, that PR #186 also addresses this. Could you double check whether you can contribute anything to the review over there? And if that PR addresses everything you're looking for, maybe close this PR here?

The changes regarding wrapper code rendering should be fine, as they are already in master in exactly the same way. No clue, why they get highlighted again.

bio/bwa-mem2/mem-samblaster/wrapper.py
johanneskoester: Actually the input should list all index files, by making use of the [multiext helper](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html?highlight=multiext#the-multiext-function).

bio/bwa-mem2/index/wrapper.py
tdayris: Changes have been made in the Snakefiles, thanks you!

bio/bwa-mem2/mem-samblaster/wrapper.py
tdayris: You are right, miltiext is now displayed in the snakefiles.

dlaehnemann (approved):
Looks good, now. I'll have a look why the tests are failing.

PR #233, opened 2020-10-19
Unicycler

tdayris:
Hi,

Unicycler is an assembler based on Spades. It handles both short and long reads and handles circular plasmids.

`samtools`, `pilon`, `racon`, `bcftools`, and `blast` are dependencies required for optional unicycler process (*e.g.* long read processing, ...)

Thank you in advance for your reviews


bio/unicycler/environment.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - conda-forge
+dependencies:
+  - bowtie2 ==2.4.1
+  - bcftools ==1.10.2
+  - spades ==3.14.1
+  - samtools ==1.10
+  - pilon ==1.23
+  - racon ==1.4.13
+  - blast ==2.10.1
+  - unicycler ==0.4.8
```
dlaehnemann, line 12: As always, I'm for unpinning the patch version of packages unless any tool needs to be on a specific patch version for a particular reason. :D

bio/unicycler/test/Snakefile
```diff
@@ -0,0 +1,8 @@
+rule test_unicycler:
+    input:
+        short1 = "reads/reads.R1.fq.gz",
+        short2 = "reads/reads.R2.fq.gz"
+    output:
+        "result/assembly.fasta"
```
dlaehnemann, line 6: Maybe make this a bit more abstract, add a `log:` entry and an empty `params.extra` to showcase that this can be used to pass further arguments, e.g.:
```suggestion
    input:
        paired = expand("reads/{{sample}}.{read}.fq.gz", read=["R1", "R2"])
    output:
        "result/{sample}.fasta"
    log:
        "logs/{sample}.log"
    params:
        extra=""
```

bio/unicycler/wrapper.py
```diff
@@ -0,0 +1,32 @@
...
+extra = snakemake.params.get("extra", "")
+
+input_reads = ""
+if "short1" in snakemake.input.keys() and "short2" in snakemake.input.keys():
+    input_reads += " --short1 {} ".format(snakemake.input["short1"])
+    input_reads += " --short2 {} ".format(snakemake.input["short2"])
+if "unpaired" in snakemake.input.keys():
+    input_reads += " --unpaired {} ".format(snakemake.input["unpaired"])
+if "long" in snakemake.input.keys():
+    input_reads += " --long {} ".format(snakemake.input["long"])
```
dlaehnemann, line 21: Ah, OK. I see the complexity, now. I would then opt for several example rules in the Snakefile that showcase different options, for example:

1. Short paired-end reads only.
2. Short paired-end and unpaired and long reads all together in one rule.

And maybe use a list in an input named `paired` instead of `short1` and `short2` separately? But that is a question of taste, I guess.

bio/unicycler/meta.yaml
```diff
@@ -0,0 +1,9 @@
+---
+name: unicycler
+description: Assemble bacterial genomes with Unicycler
```
dlaehnemann, line 3: I always like a link to the docs for quick reference.

test.py
```diff
@@ -34,20 +36,21 @@ def run(wrapper, cmd, check_log=None):
...
-
         for w in used_wrappers:
             success = False
             for ext in ("py", "R", "Rmd"):
                 script = "wrapper." + ext
                 if os.path.exists(os.path.join(w, script)):
                     os.makedirs(os.path.join(dst, w), exist_ok=True)
                     copy(w, script)
-                    success=True
+                    success = True
```
dlaehnemann, line 46: Is this an intended change? Or simply `black` doing its job?

dlaehnemann (changes requested):
Looks very good. Just some minor suggestions for improvements. Thanks a lot!

test.py
tdayris: You are right, I blacked this file as I initially did a formatting error.

bio/unicycler/meta.yaml
tdayris: Link to documentation has been added. 

I totally agree with you on **always** adding documentation and/or citation. I think this may be a future PR I'll apply for.

bio/unicycler/wrapper.py
tdayris: Short1 and Short2 have now been paired together as you suggested. Further examples will soon be added as you advised

bio/unicycler/environment.yaml
tdayris: I do not agree on this one. Even though patches usually contains minor bug fixes, it is important to my team to master every single patch version (*e.g.* for ISO standards, clinics and personal-medicine related projects).

Samtools, GATK, FastQC, Salmon, etc. All of them have patch version pinned in their respective wrappers.

bio/unicycler/test/Snakefile
tdayris: You're right, changes have been made according to your suggestions. Thanks!

bio/unicycler/wrapper.py
tdayris: Examples have been added. Thanks again for your review!

bio/unicycler/environment.yaml
dlaehnemann: I guess samtools, gatk, fastqc and salmon could or should be unpinned.

But your other argument is new to me and an important one. I would still think that you would want to get security patches and bug fixes, even in a clinics setting (or especially in such a setting). And you could still ensure consistency of results via some sort of continuous integration testing. But obviously, you have the respective hands-on experience and will know better about all the complicated details of this sort of environment.
So maybe you want to chime into this (very) recent Stackoverflow question regarding version pinning and elaborate your use case further, so it gets documented more permanently (as comments to the question, the answer, or as an altogether different answer):
https://stackoverflow.com/questions/64594146/best-practices-for-bioconda-versions-in-snakemake-wrappers

dlaehnemann (approved):
Looks good!

And I guess the [pinning / unpinning discussion](https://github.com/snakemake/snakemake-wrappers/pull/233#discussion_r516063570) is for another day. But please do chime in on it over at Stackoverflow:
https://stackoverflow.com/questions/64594146/best-practices-for-bioconda-versions-in-snakemake-wrappers

PR #234, opened 2020-10-19
Added wrapper for OpenCRAVAT

rkiminsilico:
Hi, I'm adding a snakemake wrapper for OpenCRAVAT, an open source, scalable decision support system to support variant and gene prioritization.


rkiminsilico:
Please review

bio/open-cravat/meta.yaml
```diff
@@ -0,0 +1,4 @@
+name: OpenCRAVAT
+description: Annotate variant calls with OpenCRAVAT.
```
dlaehnemann, line 2: I'd advocate for adding a link to the docs for quick reference.

bio/open-cravat/wrapper.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+def install_module(module_name, version):
+    cmd = ["oc", "module", "install", "--skip-installed", "-y"]
+    if version is not None:
+        cmd.extend(["-v", version])
+    cmd.append(module_name)
+    cmd = " ".join(cmd)
+    shell("{cmd} {log}")
+
```
dlaehnemann, line 49: I would say that installation of software (and in this case adding submodules) should be handled by the package manager, conda. Thus, I would suggest creating bioconda packages for individual submodules or packages of them that often go together. They will simply have the `open-cravat` recipe as a dependency. And then we can create separate wrappers for different `open-cravat` use cases and install all the necessary modules via the `environment.yaml` file.

bio/open-cravat/test/Snakefile
```diff
@@ -0,0 +1,12 @@
+rule run:
+    input:
+        ['example_input']
+    output:
+        ['example_input.xlsx']
+    params:
+        genome='hg38',
+        annotators=['biogrid=3.5.171'],
+        reports=['excel'],
+        output_dir='.',
```
dlaehnemann, line 10: These things should be specified via the `output`, e.g. as wildcards (e.g. `{biogrid_version}` somewhere in the file name), as file extensions (e.g. the extension `.xlsx` in an output file implies, that an Excel report is wanted) or as named outputs (could be another option for `excel`, i.e. you can specify it with `output: excel="example_input.xlsx" and then check in `wrapper.py`, whether `snakemake.output` contains an entry with key `excel`).

bio/open-cravat/test/Snakefile
```diff
@@ -0,0 +1,12 @@
+rule run:
+    input:
+        ['example_input']
```
dlaehnemann, line 3: For a single file name entry, the list brackets can be skipped:
```suggestion
        'example_input'
```

bio/open-cravat/wrapper.py
```diff
@@ -0,0 +1,302 @@
...
+            inarg_dict.update(**vars(inarg))
+        elif t == types.SimpleNamespace:
+            inarg_dict.update(**vars(inarg))
+        elif t == dict:  # {'output_dir': '/rt'}
+            inarg_dict.update(inarg)
+    inarg_dict.update(inkwargs)
+    arg_dict = get_argument_parser_defaults(parser)
+    arg_dict.update(inarg_dict)
+    args = SimpleNamespace(**arg_dict)
+    return args
```
dlaehnemann, line 39: A wrapper should not need to reimplement a full-blown command line argument parser. Instead, a wrapper should provide one or more use cases with clearly defined inputs and outputs and specific parameter settings. Any extra command line arguments can usually be set via a `params: extra=" -flag1 -argument1 argument1_value "`. And wherever possible, the behaviour of the wrapper should be determined by which inputs and outputs a user specifies and should then "automagically" set any necessary flags and arguments in the actual shell call of the tool (and throw informative errors whenever it encounters incompatible settings).

dlaehnemann (changes requested):
Wow, there is a lot of complexity and a lot of work in this wrapper. Thanks for investing so much time.

I think it is necessary to untangle some of that complexity and I see two major ways of doing so:

1. Installation of modules should be handled by conda and I'd suggest adding module-specific recipes to bioconda, that each have `open-cravat` as a dependency and thus extend it when they are installed. This should be feasible by using the [installation commands provided provided here](https://github.com/KarchinLab/open-cravat/wiki/quickstart-command-line#install-an-annotator) in a build.sh file.
2. Break this up into several wrappers for different OpenCRAVAT use cases or create several example rules for different use cases in the `Snakefile` (and then test each in `test.py`). For this, as many options and arguments should be automagically inferred from `input:` and `output:` specifications and extra arguments provided in the rule via `params: extra=""` or a limited number of named `params`.

In the current form, it is both not clear to me how exactly to use the wrapper and how to review it in detail, without digging deep into the OpenCRAVAT docs. But with the two above suggestions, both these things should get a lot easier.

Sorry for suggesting so much more work, but I think it will make using and mainting the wrapper(s) a lot easier in the long run. And thanks again for this massive effort!

bio/open-cravat/meta.yaml
rkiminsilico: committed to bfcbaac

bio/open-cravat/wrapper.py
rkiminsilico: Thanks for the suggestion. open-cravat currently has about 130 annotation modules and their location can differ by user configuration through open-cravat. open-cravat users are supposed to install annotation modules as needed and use them and there is not yet open-cravat submodules which often go together. The number of open-cravat submodules is growing as new open-cravat annotation modules are frequently added. Thus, I am not sure if installing open-cravat submodules with conda is the best way. Any advice would be appreciated.

dlaehnemann: Hmm, I see that this would mean a lot of bioconda recipes. At the same time, these would be highly similar, so adding them in a big batch should be possible. And the problem with installing modules as part of a wrapper is that this wrapper potentially gets invoked several times in parallel on the same cluster / machine and then several processes would try to install stuff to the same location in parallel, which would create a mess. If this is handled by conda, the installation is done once and the resulting environment can then be used by any number of invocations. And then you can offer a wrapper per annotation module, simply by adjusting the `environments.yaml` file to contain the needed module -- which might also help in creating more specific wrappers for particular use cases.

bio/open-cravat/test/Snakefile
rkiminsilico: Removing `outputs` parameter is possible as you described. 

`annotators` parameter seems to be a bit complicated: OpenCRAVAT has a pipeline internally, and at each step of the pipeline, files are generated. At the end of the run, intermediate files can be deleted optionally. For example, when an OpenCRAVAT run starts with `oc run input -a biogrid -t vcf`, `input.crv` and `input.crg' is created by an input format converter module of OpenCRAVAT (depending on the format of the input file). Then, `input.crx` is created by a gene model mapping module and `input.biogrid.gen` is generated as a result of running the biogrid portion of OpenCRAVAT. Then, `input.sqlite` is generated which is the result database file of the job. Then, `input.variant.vcf` is generated by the VCF-format reporter of OpenCRAVAT, as the final output of the job. This process goes without stop. To stop OpenCRAVAT after creating each intermediate file, options can be given: for example, `oc run input --startat converter --endat converter` will create `input.crv` and `input.crg` and stop. `oc run input --startat mapper --endat mapper` will create `input.crx` and stop. `oc run input --startat annotator --endat annotator -a biogrid` will create `input.biogrid.gen` and stop. `oc run input --startat aggregator --endat postaggregator` will create `input.sqlite` and stop. Finally, `oc run input --startat reporter --endat reporter -t vcf` will create `input.variant.vcf` and stop. To implement what I understand as what you suggest, I can think of the following:

```
rule report:
   input: "input.sqlite"
   output: "input.variant.vcf"
   wrapper: "master/bio/open-cravat"
rule db:
   input: ["input.crx", "input.biogrid.gen"]
   output: "input.sqlite"
   wrapper: "master/bio/open-cravat"
rule biogrid:
   input: "input.crg"
   output: "input.biogrid.gen"
   wrapper: "master/bio/open-cravat"
rule crx:
   input: "input.crv"
   output: "input.crx"
   wrapper: "master/bio/open-cravat"
rule crv:
   input: "input"
   output: ["input.crv", "input.crg"]
   wrapper: "master/bio/open-cravat"
```

This would be basically replacing/reimplementing OpenCRAVAT's internal pipeline with snakemake. OpenCRAVAT has been trying to make user experience simpler so that the same can be run with `oc run input -a biogrid -t vcf`. I am new to snakemake, so I wonder - the above Snakemake file, do snakemake users prefer that kind of fine-grained process definition?

Regarding `{biogrid_version}`, OpenCRAVAT does not have it currently as shown above with `input.biogrid.gen` and OpenCRAVAT team can have a discussion about that.

dlaehnemann: Yes, from the general idea of wrappers, I would be for implementing individual substeps of OpenCRAVAT as separate rules. In a sense, OpenCRAVAT is re-implementing what a workflow manager like Snakemake is actually built for (and probably better at doing, because that's what it is built for): coordinating the execution of a bunch of steps one after the other. Only that Snakemake will give you much more power to deploy the whole pipeline in any context from a small computer with only one core to a big server or even different cloud services, simply by changing some command-line parameters. Plus a lot of configurability of that deployment.

And if you are aiming to make the `wrapper.py` easier to review (and then maintain), you could even go for separate wrappers for different substeps (e.g. `wrapper: "master/bio/open-cravat/converter"`, `wrapper: "master/bio/open-cravat/mapper"`, etc.), as they will have very particular required inputs and outputs and probably certain default parameters. This way, each substep would have a `wrapper.py` file that only needs to handle some parameters and very defined input and output extensions.
Also, for a similar tool that offers lots of subcommands as steps in a pipeline (but implemented in R), check out the recently merged `dada2` wrappers. Just to get an idea of the general setup. For these, there are currently two pull requests open to turn the collection of wrappers into a basic pipeline by chaining them together in a meta-wrapper, see e.g.:
https://github.com/snakemake/snakemake-wrappers/pull/249

And if you really want to create a snakemake pipeline that users can use as a template where they just have to edit some configuration files and provide data annotation and can then run the whole thing (including quality control output and the possibility to create a snakemake report, afterwards), [Snakemake Workflows](https://github.com/snakemake-workflows/docs) would be what you are looking for. So this is what I would recommend for the easy user experience in the end. It keeps the granularity of the wrappers, so that the execution of the pipeline can be optimised. It keeps the individual rules / wrappers configurable while the `wrapper.py`s remain maintainable. And it clearly documents the configuration of the whole pipeline / workflow in `config.yaml` and `samples.tsv` files in a `config/` folder (and a [snakemake report](https://snakemake.readthedocs.io/en/latest/snakefiles/reporting.html#reports), if you implement that).

So in summary: I would opt for finely granular (separate) wrappers that each use OpenCRAVAT to do one step at a time. This way, when creating a snakemake workflow that uses them, snakemake can schedule small jobs in the most optimal way, no matter what the specifications of the system it runs on. To give users an easy copy-pasteable version of the basic pipeline, I would create a meta-wrapper that provides rules for all steps (using the individual wrappers). And in a final step you could then go and add bells and whistles by building a full snakemake workflow (preprocessing, quality control, further plots, a [snakemake report](https://snakemake.readthedocs.io/en/latest/snakefiles/reporting.html#reports)).

I hope this helps moving your discussions on how to go about this forward. And don't hesitate to ask further. These different levels of granularity are still emerging as a consensus, so I must say that the global view I provide here is not really documented well, anywhere. Currently, there's only the separate documentations of `snakemake`, `snakemake-wrappers` and `snakemake-workflows` that at some points mention each other...

rkiminsilico: Thanks for your detailed explanation, which helped me understanding the philosophy of snakemake a lot. I appreciate it. Following your advice, I actually implemented individual wrappers for each step of the OpenCRAVAT pipeline, but in the end, it seemed overly complicated. OpenCRAVAT has its pipeline, but basically its main use is to get some input files and then produce some output files. "annotators" are basically the same concept as VEP's plugins. I see that VEP's annotate wrapper has "plugins" parameters in "annotate" wrapper (an example [here](https://github.com/snakemake/snakemake-wrappers/blob/master/bio/vep/annotate/test/Snakefile)). OpenCRAVAT is the same kind. For example, the following will be a syntactic equivalent for OpenCRAVAT:
```
rule annotate_variants:
  input:
    "input.vcf"
  output:
    excel="input.xlsx",
    tsv="input.variant.tsv"
  params:
    annotators=["biogrid", "clinvar"],
    extra="-l hg38"
  wrapper:
    "master/bio/open-cravat/annotate"
```
If the finer control of OpenoCRAVAT which I initially envisioned could be set aside for now, can a schema like the above be approved? Just as VEP wrapper will fail if it does not have correct cache or plugins, OpenCRAVAT will simply fail with a log message if a required annotator module is not installed.

dlaehnemann: Yes, I think implementing one rule for the whole OpenCRAVAT pipeline is also possible. And regarding the `annotators`, VEP also has a mechanism for downloading them via a separate wrapper:
https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/vep/plugins.html

Thus, optimally you would create this as a second wrapper and users can then enforce such a download in a proper Snakemakey way by requiring some file (or folder) that the download rule / wrapper creates as input for the main OpenCRAVAT wrapper.

And regarding fine-grained configurability, I would reiterate what I [recommended elsewhere in this PR](https://github.com/snakemake/snakemake-wrappers/pull/234#discussion_r510296991):

* try to infer as many command line arguments from the specified `input` and `output` files and their extensions (e.g. after a quick glance over at the [OpenCRAVAT wiki page on command-line usage](https://github.com/KarchinLab/open-cravat/wiki/2.-Command-line-usage), arguments `-n` and `-d` should be inferred from the `output` file paths, stripping off the extensions; `-t` should be handled by looking at the extensions in `output`; `-l` should be provided as a required `input`; `-a` should be inferred from a required named input list `input.annotators`, which will ensure each of those is downloaded beforehand by the other wrapper suggested above; `-c` could check whether a named input `input.config` is present in the rule specification and if it is, add this to the command that is run)
* handle other command line arguments via string fields in `params`, e.g. `params.extra` -- in a rule example this would look like:
  ```
    params:
        extra = "-v"
  ```

I hope this provides a workable solution. But I think we are getting there, with only a few more iterations! :D

bio/open-cravat/wrapper.py
rkiminsilico: Thanks for the help. There are now two wrappers, `module` and `run`, for installing OpenCRAVAT modules and running OpenCRAVAT, respectively. `module` wrapper checks the directory of each module as output. `run` receives those module directories as a part of input, along with input files. `run` wrapper recognizes which should be input files and which should be module directories.

bio/open-cravat/test/Snakefile
rkiminsilico: According to your recommendation, input and output for `run` wrapper are now configured as follows:
```
modules_dir = 'modules'
rule run:
    input:
        'example_input',
        modules_dir + '/commons/hg38wgs',
        modules_dir + '/converters/cravat-converter',
        modules_dir + '/mappers/hg38',
        modules_dir + '/annotators/biogrid', 
        modules_dir + '/annotators/clinvar',
        modules_dir + '/postaggregators/tagsampler',
        modules_dir + '/postaggregators/varmeta',
        modules_dir + '/postaggregators/vcfinfo',
        modules_dir + '/reporters/excelreporter',
        modules_dir + '/reporters/tsvreporter',
        modules_dir + '/reporters/csvreporter',
    output:
        'example_input.xlsx',
        'example_input.variant.tsv', 
        'example_input.variant.csv'
    log:
        "logs/open-cravat/run.log"
    wrapper:
        'master/bio/open-cravat/report'
```
Which annotator modules to run and which reporter types are requested are all inferred from `input`. For missing modules, `module` wrapper can be used in the same Snakemake file as follows:
```
rule installmodules:
    output:
        directory(modules_dir + '/commons/hg38wgs'), 
        directory(modules_dir + '/converters/cravat-converter'), 
        directory(modules_dir + '/mappers/hg38'), 
        directory(modules_dir + '/annotators/biogrid'), 
        directory(modules_dir + '/annotators/clinvar'),
        directory(modules_dir + '/postaggregators/tagsampler'),
        directory(modules_dir + '/postaggregators/varmeta'),
        directory(modules_dir + '/postaggregators/vcfinfo'),
        directory(modules_dir + '/reporters/excelreporter'),
        directory(modules_dir + '/reporters/tsvreporter'),
        directory(modules_dir + '/reporters/csvreporter'),
    log:
        "logs/open-cravat/module.log"
    wrapper:
        'master/bio/open-cravat/module'
```

bio/open-cravat/wrapper.py
rkiminsilico: Argument parser-related parts have been removed.

rkiminsilico:
@dlaehnemann I think I have implemented what were asked. Tests passed. Please review.

bio/open-cravat/run/test/Snakefile
```diff
@@ -0,0 +1,42 @@
+rule run:
+    input:
+        'example_input',
```
johanneskoester, line 3: Can you add a comment what is the supposed input here? I would've expected it to be just a VCF file, but that does not seem to be the case?

johanneskoester: Is it BED? In any case, please use an extension for the input.

bio/open-cravat/run/test/module/test/Snakefile
```diff
@@ -0,0 +1,7 @@
+rule module:
```
johanneskoester, line 1: the module subfolder seems to be a repetition of the module rule above. Is it for testing purposes?

johanneskoester (changes requested):
Almost there, thanks a lot! See some final questions below.

bio/open-cravat/run/test/Snakefile
rkiminsilico: The test file is in the OpenCRAVAT input format (OpenCRAVAT can accept VCF files, but the test input file is in the OpenCRAVAT input format). I have added ".tsv" extension to the file since it is tab-delimited.

bio/open-cravat/run/test/module/test/Snakefile
rkiminsilico: Yes (testing purposes). That was needed for my local test. I have just removed `module` folder under `run/test` and used the repo URL for the `module` wrapper in the Snakefile at `run/test` instead.

rkiminsilico: Using the repo URL for the `module` wrapper in the `Snakefile` of `run/test` did not work. For testing purposes, I made `module_wrapper` inside `run/test` and put a wrapper file and an environment file.

rkiminsilico:
Requested changes were made. Please review.

bio/open-cravat/run/test/module/test/Snakefile
johanneskoester: Maybe instead just add the module folders to the test case? Or are those files large?

rkiminsilico: The usual required modules are about 1GB, and probably the critically required modules just for the test run may be still about 200 to 300MB. Not sure if this size is small enough?

johanneskoester (approved):

PR #238, opened 2020-10-26
Homer annotatepeaks

jafors:
This should fix a bug in the HOMER annotatePeaks wrapper.
Before, the wrapper required a specific input file even if the corresponding output was not chosen.


dlaehnemann (approved):
This looks good to me, so I'll approve it. But I would suggest to wait with merging until @AntonieV has also had a look, as she originally created the wrapper and might spot stuff that I don't see.

AntonieV:
This looks good, thanks for finding and fixing this bug!

Issue #236, opened 2020-10-20
VEP cache: specify output folder


fgvieira:
**Snakemake version**
snakemake: `5.25.0`
snakemake-wrappers: `0.65.0`

**Describe the bug**
Wrapper `vep/cache` has, as single output, a path where to download the cache:
https://github.com/snakemake/snakemake-wrappers/blob/master/bio/vep/cache/test/Snakefile#L3

Inside this folder, the wrapper will create folders with species, build and release. However, these folders are not part of the rule output path, so the rule cannot properly check if the output exists or not. That is, if the pipeline is run twice with different genomes, the second time will use the cache of the first run.

**Suggested fix**
I'd suggest having the `species`/`build`/`release` as wildcards and part of the output folder. The wrapper would have to make sure the output path is valid (would have to end in `{species}/{release}_{build}`), download the cache to a temp folder and then move to the right output folder; for example:
```
rule get_vep_cache:
    output:
        directory("resources/vep/cache/{species}/{release}_{build}")
    params:
	species = lambda wildcards: wildcards.species,
	build = lambda wildcards: wildcards.build,
	release = lambda wildcards: wildcards.release,
    log:
        "logs/vep/cache/{species}/{release}_{build}.log"
    wrapper:
        "master/bio/vep/cache"
```

The `params` could also be deprecated, if the wrapper can get `species`/`build`/`release` from the wildcards.

The `vep annotate` rule would take the full path to the cache and infer `species`/`build`/`release`.

fgvieira:
Closing.
See #241

PR #241, opened 2020-10-28
VEP cache with fully specified output folder with wildcards

fgvieira:
Implementation of #236, where the output folder has to be fully specified (including the part that is added by the ensembl script),  so that the outptut can be checked by the rule if it exists or not.

This way, we can specify the specific cache for a given species/build/release and check if it exists.

### DIFF
--- a/bio/vep/cache/meta.yaml
+++ b/bio/vep/cache/meta.yaml
@@ -2,3 +2,4 @@ name: vep download cache
 description: Download VEP cache for given species, build and release.
 authors:
   - Johannes Köster
+  - Filipe G. Vieira
--- a/bio/vep/cache/test/Snakefile
+++ b/bio/vep/cache/test/Snakefile
@@ -1,12 +1,8 @@
 rule get_vep_cache:
     output:
-        directory("resources/vep/cache")
-    params:
-        species="saccharomyces_cerevisiae",
-        build="R64-1-1",
-        release="98"
+        directory("resources/vep/cache/{species}/{release}_{build}")
     log:
-        "logs/vep/cache.log"
+        "logs/vep/cache/{species}/{release}_{build}.log"
     cache: True  # save space and time with between workflow caching (see docs)
     wrapper:
         "master/bio/vep/cache"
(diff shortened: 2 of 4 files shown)


johanneskoester:
If I am not completely misunderstanding something, I don't think this PR is needed. Specifying species and release in the wildcards is already possible now: 
```
rule get_vep_cache:
    output:
        directory("resources/vep/cache/{species}/{release}_{build}")
    params:
        species=lambda w: w.species,
        build=lambda w: w.build,
        release=lambda w: w.release,
    log:
        "logs/vep/cache/{species}/{release}_{build}.log"
    cache: True  # save space and time with between workflow caching (see docs)
    wrapper:
        "master/bio/vep/cache"
```

In other words, the current wrapper is usable in both ways:
1. one species, build, release per workflow (hence no need for the wildcards).
2. multiple species, build, release in one workflow (encoding via wildcards).

johanneskoester:
Please reopen if I did not correctly understand this.

johanneskoester:
And, in any case, still thanks a lot for taking the time to work on this! What we could easily do is to add above example to the current wrapper.

fgvieira:
Yes, you could include the wildcards but the output folder would actually end up being `resources/vep/cache/{species}/{release}_{build}/{species}/{release}_{build}`, since the `ensembl-vep` cache script creates extra folders. I guess it is not that critical, just a bit redundant.. :smile: 

And how would that play out with the snakemake `cache`, if there are no wildcards?

johanneskoester:
> Yes, you could include the wildcards but the output folder would actually end up being `resources/vep/cache/{species}/{release}_{build}/{species}/{release}_{build}`, since the `ensembl-vep` cache script creates extra folders. I guess it is not that critical, just a bit redundant.. smile
True, but you don't see that, and consuming vep wrappers handle that correctly.

> And how would that play out with the snakemake `cache`, if there are no wildcards?

The cache hashes params, code and software stack. It does not hash wildcards (as people can name them differently). Hence, it is even crucial to have these encoded in params.

PR #245, opened 2020-10-29
Update the rst template to display I/O (fixes #244)

### LINKED ISSUE #244: Displaying the wrapper input and output in the documentation
**Is your feature request related to a problem? Please describe.**
I noticed that while the `meta.yaml` are well completed, the info within is not displayed in the documentation (even in v0.67.0), especially the description of input and output of the wrapper.

**Describe the solution you'd like**
Update the Jinja `.rst` template to print out these info.

**Describe alternatives you've considered**
I will do a PR with a suggestion of template.

Best, 
Charlie

cpauvert:
Hi, 
This is a suggestion to display the info hidden in the `meta.yaml` that would benefit wrapper's users.

* Style homogeneization of dependencies display in wrapper docs
* Added a new section for Input / Output if documented in `meta.yaml`
* Style modification for wrapper names in meta-wrapper docs

## Before 
Before screenshot

## After
After screenshot

Best, 
Charlie


dlaehnemann:
Looks good. One small suggestion: I would put any extra stuff that is rendered underneath the example. We could even move the dependencies underneath the example and only keep the general description above it. I think the example is usually what most people need, and they will only look at the details if they need them to adjust or debug something.

cpauvert:
Alright, I'll update the PR :)

cpauvert:
I reordered the sections and also added `jinja` options to be able to indent more clearly the `rst` template.

However, some `meta.yaml` raised some warnings concerning duplicates. Nonetheless, I have to _clean my own backyard_ as the saying goes because I specified the named output in the `meta.yaml` of the dada2 wrappers wrongly according to jinja (but will correct that soon):

## Deprecated syntax for `meta.yaml`

```yaml
output:
    arg1: description
    arg2: long description
```
This does not create a dictionary nor a list but a named list impossible to access from the template.

## Recommended `meta.yaml`

```yaml
output:
  - arg1: description
  - arg2: long description
```
which creates a dictionary OR

```yaml
output:
  -  description
  -  long description
```
which creates a list

I hope you folks are ok with these changes, otherwise do not hesitate to jump in for a review.

## Updated screenshot

Capture d’écran_2020-10-29_22-07-07

bio/ucsc/bedGraphToBigWig/meta.yaml
```diff
@@ -1,10 +1,10 @@
 name: bedGraphToBigWig
 description: |
-  Convert *.bedGraph file to *.bw file (see http://hgdownload.soe.ucsc.edu/admin/exe/linux.x86_64/FOOTER.txt)
+  Convert \*.bedGraph file to \*.bw file (see http://hgdownload.soe.ucsc.edu/admin/exe/linux.x86_64/FOOTER.txt)
 authors:
   - Roman Cherniatchik
 input:
-  bedGraph: Path to *.bedGraph file
+  bedGraph: Path to \*.bedGraph file
   chromsizes: Chrom sizes file, could be generated by twoBitInfo or downloaded from UCSC
```
dlaehnemann, line 8:
```suggestion
  - bedGraph: Path to \*.bedGraph file
  - chromsizes: Chrom sizes file, could be generated by twoBitInfo or downloaded from UCSC
```

docs/generate_docs.py
```diff
@@ -34,7 +34,7 @@
     TOOL_TEMPLATE = Template(f.read())
 
 with open(os.path.join(BASE_DIR, "_templates", "wrapper.rst")) as f:
-    TEMPLATE_WRAPPER = Template(f.read())
+    TEMPLATE_WRAPPER = Template(f.read(), trim_blocks=True, lstrip_blocks=True)
```
dlaehnemann, line 37: Just out of curiosity: What do these extra parameters do? And should they also be added to the `TEMPLATE_META = Template()` call below?

dlaehnemann (commented):
Sure, updates to newer syntax are always a good idea. I just spotted one instance of old syntax you seem to have missed. And I have one question about the new `Template()` parameters.

docs/generate_docs.py
cpauvert: Jinja template tags (between the curly brackets) left  a blank line once processed. Sometimes  they even let whitespaces.
Usually this is no trouble, for generating html for instance. But with rst/markdown syntax, every spaces and lines has its importance and can break the formatting.
Hence, adding these options can actually remove the blank lines and the whitespaces that would have remained otherwise. One other ugly workaround (which I did in the first commit) is to pack all the jinja tags on one line but this is hardly readable.

The template for the meta-wrapper docs needed not more complex jinja tags and was therefore fine with the defaults

dlaehnemann (approved):
Thanks for the addition!

PR #240, opened 2020-10-28
Specify the YAML loader (fixes #239)

cpauvert:
Simple fix for the warning when building the documentation. According to the [pyyaml docs]( https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation) it is also safer.
Best, 
Charlie

Issue #239, opened 2020-10-28
Warning when building docs YAMLLoadWarning


cpauvert:
**Snakemake version**
Snakemake-wrappers version 0.67.0 and Sphinx v3.2.1 

**Describe the bug**
A warning is issued by Sphinx when building the docs locally or remotely with the Github Actions.

**Logs**

```
/home/runner/work/snakemake-wrappers/snakemake-wrappers/docs/generate_docs.py:72: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  env = yaml.load(env)
```
Can be found [here](https://github.com/snakemake/snakemake-wrappers/pull/224/checks?check_run_id=1321013873#step:4:8) as well.

**Additional context**
https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation#how-to-disable-the-warning

This is not a priority, I just stumble upon several times and decided to register it. I will probably PR a minor fix soon.
Best, 
Charlie

PR #246, opened 2020-10-31
specify yaml loader in test.py

dlaehnemann:
strategy from PR #240 and issue #239 by @cpauvert

### DIFF
--- a/test.py
+++ b/test.py
@@ -31,7 +31,7 @@ def run(wrapper, cmd, check_log=None):
         if os.path.exists(os.path.join(wrapper, wrapper_file)):
             # is meta wrapper
             with open(os.path.join(wrapper, wrapper_file), "r") as wf:
-                wf = yaml.load(wf)
+                wf = yaml.load(wf, Loader=yaml.BaseLoader)
                 used_wrappers = wf["wrappers"]
         else:
             used_wrappers.append(wrapper)


test.py
```diff
@@ -31,7 +31,7 @@ def run(wrapper, cmd, check_log=None):
         if os.path.exists(os.path.join(wrapper, wrapper_file)):
             # is meta wrapper
             with open(os.path.join(wrapper, wrapper_file), "r") as wf:
-                wf = yaml.load(wf)
+                wf = yaml.load(wf, Loader=yaml.BaseLoader)
```
tedil, line 34: so which Loader do we want to use here?
from the pyyaml website:
> Loader(stream) is the most common of the above classes and should be used in most cases. stream is an input YAML stream. It can be a string, a Unicode string, an open file, an open Unicode file.

> Loader supports all predefined tags and may construct an arbitrary Python object. Therefore it is not safe to use Loader to load a document received from an untrusted source. By default, the functions scan, parse, compose, construct, and others use Loader.

> SafeLoader(stream) supports only standard YAML tags and thus it does not construct class instances and probably safe to use with documents received from an untrusted source. The functions safe_load and safe_load_all use SafeLoader to parse a stream.


> BaseLoader(stream) does not resolve or support any tags and construct only basic Python objects: lists, dictionaries and Unicode strings.

> CLoader, CSafeLoader, CBaseLoader are versions of the above classes written in C using the LibYAML library.

dlaehnemann: Well, I went for the `BaseLoader`, because I think that should suffice and should be as safe (or safer?) than `SafeLoader`. We're basically using these yaml-files to construct strings, right?

tedil (approved):

mbargull (approved):
Approving just because @tedil approved -- and to test out that people "from the outside" can approve, too :).

dlaehnemann:
So you do go for internet points, @mbargull ... :D

mbargull:
> So you do go for internet points, @mbargull ... :D

*old man waves hand*

But nah, I just have this "limit the platforms I'm active on and prefer those close to code"-thing ;).

test.py
mbargull: To actually add content-related input (and not just drive-by-approving):
You want to use `BaseLoader` if you want to read the "base" YAML but without using any of those standard "schemas". Meaning, it'll use the (I believe it's called) "failsafe schema" which only distinguishes between scalars (strings), mappings and sequences. Ergo, with will sidestep any other value interpretation, e.g., `value: 1` will be loaded as `{'value': '1'}` and explicitly *not* as `{'value': 1}`.

(`conda-build` does the same thing, which is why I know of this. It's handy because then you can just write `version: 1.2` and not have it interpret it as "string 'version' maps to a floating point number 1.2". Of course, this mean you have to handle the conversions yourself whenever you do want to interpret values as numbers, Booleans, etc.)

dlaehnemann: Yeah, that was roughly how I understood it. And from what I see, all we are doing is taking that structured input and parsing it into strings. So this should be exactly what we want... :D

mbargull: Perfect! Knowing this context, I want to approve -- oh wait, already done :).

PR #249, opened 2020-11-03
add meta-wrapper dada2-pe

cpauvert:
Hi, 
Some recently  included dada2 wrappers were bundled together to make this meta-wrapper (**pe**-version). Changes from the wrapper codes includes:

* Adjustments of filenames and add onstart message
* Necessary example files for the meta-wrapper
* Add test in test.py

Docs and local test runs fine.

However,  I have an open question so far (maybe @dlaehnemann), I want to summarize the number of reads and ASV throughout the different rules (just like in the [tutorial](https://benjjneb.github.io/dada2/tutorial.html#track-reads-through-the-pipeline) and my scripts). Initially, I planned to add in the meta-wrapper a `stats` rule that would parse all output file using an R script and output a table/plot.
But I figured that the awesome [reporting](https://snakemake.readthedocs.io/en/stable/snakefiles/reporting.html) ability of Snakemake could fit the job perfectly. But I'm not sure if I have to modify every wrapper code to report something, or could I had reporting rules to the meta-wrapper Snakefile?

Best, 
Charlie
EDIT: Added the rule graph as a summary
rule

PR #250, opened 2020-11-04
added missing dada2 wrapper for collapseNoMismatch

cpauvert:
Hi (again),
So while doing the dada2 meta-wrapper #249 I realize a wrapper was missing. This function usually does not change the results (if your amplicon sequence has not length variation unlike ITS) but it is safer to run it to avoid overestimating the number of ASV in the dataset. This function is not part of tutorial, that's why I only remenbered it when digging my own scripts.

Thanks in advance for the review, 
Best, 
Charlie


dlaehnemann (approved):
You've definitely figured out these R wrappers by now!

Tip of the hat for linking to showcases of why to use this wrapper in the dada2 issues!

cpauvert:
> You've definitely figured out these R wrappers by now!

Haha thank you. We did it together though!

> Tip of the hat for linking to showcases of why to use this wrapper in the dada2 issues!

:+1: 

At some point, I will suggest an automatic linter action for R just like the one for Python (though R does not have any PEP) but just to ensure some nice formatting

PR #252, opened 2020-11-04
Replace ifelse() by if() (fixes #251)

### LINKED ISSUE #251: Only first of samples names is passed in dada2 make-table wrapper
**Snakemake version**
current snakemake-wrappers version 0.67.0 (bbd7c0fa6286c1569f727817b928bd40f597a9ce)

**Describe the bug**
The rownames of the resulting sequence table should have the names `c("a","b")` but have `c("a",NA)` instead.

**Minimal example**
The minimal example for reproducing the bug is in the test directory of the wrapper.

cpauvert:
ifelse() reduces the length of the vector to 1 instead of keeping the length.
Minor correction to resolves #251

### DIFF
--- a/bio/dada2/make-table/wrapper.R
+++ b/bio/dada2/make-table/wrapper.R
@@ -14,9 +14,8 @@ sink(log.file,type="message")
 library(dada2)
 
 # If names are provided use them
-nm<-ifelse(is.null(snakemake@params[["names"]]),
-           NULL,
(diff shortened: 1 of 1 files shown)


dlaehnemann (approved):
That's a weird side-effect of `ifelse()` I have not seen before...

cpauvert:
Thanks for the merge! Yeah it is annoying, and I somehow keep forgetting it.

PR #256, opened 2020-11-09
Automated detection of BAI file, memory, and temp dir

fgvieira:
Added automated detection of BAI file (generated by default), memory, and temp dir.
Set BED file as an input file and not a parameter.

### DIFF
--- a/bio/gatk3/indelrealigner/test/Snakefile
+++ b/bio/gatk3/indelrealigner/test/Snakefile
@@ -1,11 +1,13 @@
 rule indelrealigner:
     input:
-        bam="mapped/{sample}.bam",
+        bam="mapped/{sample}.bam",              # input bam file (indexed)
         ref="genome.fasta",
-        known="dbsnp.vcf.gz",
+        known="dbsnp.vcf.gz",                   # vcf files of known variation (indexed)
         target_intervals="{sample}.intervals"
     output:
-        bam="realigned/{sample}.bam"
+        bam="realigned/{sample}.bam",
+        bai="realigned/{sample}.bai",
+        java_temp=temp(directory("/tmp/gatk3_indelrealigner/{sample}")),
     log:
         "logs/gatk3/indelrealigner/{sample}.log"
     params:
@@ -13,5 +15,7 @@ rule indelrealigner:
     resources:
         mem_mb = 1024
     threads: 16
+    resources:
+        mem_mb = 10240
     wrapper:
         "bio/gatk/indelrealigner"
--- a/bio/gatk3/indelrealigner/test/Snakefile
+++ b/bio/gatk3/indelrealigner/test/Snakefile
@@ -1,15 +1,15 @@
 rule indelrealigner:
     input:
         bam="mapped/{sample}.bam",
-	bai="mapped/{sample}.bai",
+        bai="mapped/{sample}.bai",
         ref="genome.fasta",
         known="dbsnp.vcf.gz",
-	known_idx="dbsnp.vcf.gz.tbi",
+        known_idx="dbsnp.vcf.gz.tbi",
         target_intervals="{sample}.intervals"
     output:
         bam="realigned/{sample}.bam",
         bai="realigned/{sample}.bai",
-	java_temp=temp(directory("/tmp/gatk3_indelrealigner/{sample}")),
+        java_temp=temp(directory("/tmp/gatk3_indelrealigner/{sample}")),
     log:
         "logs/gatk3/indelrealigner/{sample}.log"
     params:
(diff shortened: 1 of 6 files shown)


bio/gatk3/indelrealigner/test/Snakefile
```diff
@@ -1,16 +1,20 @@
 rule indelrealigner:
     input:
-        bam="mapped/{sample}.bam",
+        bam="mapped/{sample}.bam",              # input bam file (indexed)
         ref="genome.fasta",
-        known="dbsnp.vcf.gz",
+        known="dbsnp.vcf.gz",                   # vcf files of known variation (indexed)
```
dlaehnemann, line 5: If indices are needed, they should be required input:
```suggestion
        bam="mapped/{sample}.bam",
        bai="mapped/{sample}.bai",
        ref="genome.fasta",
        known="dbsnp.vcf.gz",
        known_index="dbsnp.vcf.gz.csi",
```

bio/gatk3/indelrealigner/wrapper.py
```diff
@@ -35,6 +56,6 @@
     " {input_known_string}"
     " {bed}"
     " --targetIntervals {input_target_intervals}"
-    " -o {snakemake.output}"
+    " -o {output_bam}"
```
dlaehnemann, line 59: To avoid the explicit get operation from above, you can simply make this:
```suggestion
    " -o {snakemake.output.bam}"
```

bio/gatk3/indelrealigner/wrapper.py
```diff
@@ -14,8 +14,25 @@
...
+# Getting memory in gigabytes, for user convenience. Please prefer the use
+# of mem_mb over mem_gb as advised in documentation.
+elif "mem_gb" in snakemake.resources.keys() and not "-Xmx" in java_opts:
+    java_opts += " -Xmx{}G".format(snakemake.resources["mem_gb"])
+
+# Getting java temp directory from output files list, if -Djava.io.tmpdir
+# is not provided in java parameters. By doing so, backward compatibility is
+# not broken.
+if "java_temp" in snakemake.output.keys() and not "-Djava.io.tmpdir" in java_opts:
+    java_opts += " -Djava.io.tmpdir={}".format(snakemake.output["java_temp"])
```
dlaehnemann, line 33: This is now done via the new dependency of `snakemake-wrapper-utils`. A version of this wrapper with this included was merged yesterday, so I'm afraid you'll have to merge in the master and resolve any conflicts. Sorry for the inconvenient timing...

bio/gatk3/indelrealigner/wrapper.py
dlaehnemann: While we are at it, I would do the same for `snakemake.input.bam`, `snakemake.input.ref` and `snakemake.input.target_intervals`. Nothing is ever done to parse them any further, so we can simply paste them in here.

dlaehnemann (changes requested):
Thanks a lot for the cleanup effort in these wrappers!

Just a few further suggestions -- and I'm afraid you'll have to merge in the most recent master branch changes, as these wrappers were updated in the big Java memory handling pull request that was merged yesterday:
https://github.com/snakemake/snakemake-wrappers/pull/204

bio/gatk3/indelrealigner/wrapper.py
fgvieira: Ok, I'll merge the lastest commits

bio/gatk3/indelrealigner/wrapper.py
```diff
@@ -8,34 +8,39 @@
 from snakemake.shell import shell
 from snakemake_wrapper_utils.java import get_java_opts
 
+input_known = snakemake.input.known
```
dlaehnemann, line 11: I would do this via a get, so that you don't get a key error if `known=""` is not specified (which is a valid choice, I think?):
```suggestion
input_known = snakemake.input.get("known", "")
```

dlaehnemann: Never mind, I double checked the [old docs](https://github.com/broadinstitute/gatk-docs/blob/master/gatk3-tooldocs/3.8-0/org_broadinstitute_gatk_tools_walkers_indels_IndelRealigner.html) and this is indeed an obligatory parameter and thus a key error is fine! Sorry for the noise...

dlaehnemann (approved):
Looks good, sorry for the last comment...

bio/gatk3/indelrealigner/wrapper.py
fgvieira: No problem... :smile:

PR #257, opened 2020-11-09
Add missing R dependencies

fgvieira:
Added missing R dependencies.

### DIFF
--- a/bio/mapdamage2/environment.yaml
+++ b/bio/mapdamage2/environment.yaml
@@ -7,6 +7,6 @@ dependencies:
   - mapdamage2 =2.2.1
   - r-inline
   - r-ggplot2
-  - gam
-  - Rcpp
-  - RcppGSL
+  - r-gam
+  - r-rcpp
+  - r-rcppgsl
--- a/bio/mapdamage2/environment.yaml
+++ b/bio/mapdamage2/environment.yaml
@@ -1,6 +1,12 @@
 channels:
   - bioconda
   - conda-forge
+  - r
   - defaults
 dependencies:
   - mapdamage2 =2.2.1
+  - r-inline
+  - r-ggplot2
+  - gam
+  - Rcpp
+  - RcppGSL
--- a/bio/mapdamage2/environment.yaml
+++ b/bio/mapdamage2/environment.yaml
@@ -1,7 +1,6 @@
 channels:
   - bioconda
   - conda-forge
-  - r
   - defaults
 dependencies:
   - mapdamage2 =2.2.1
--- a/bio/mapdamage2/environment.yaml
+++ b/bio/mapdamage2/environment.yaml
@@ -3,9 +3,4 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - mapdamage2 =2.2.1
-  - r-inline
-  - r-ggplot2
-  - r-gam
-  - r-rcpp
-  - r-rcppgsl
+  - mapdamage2 =2.2


bio/mapdamage2/environment.yaml
```diff
@@ -1,6 +1,12 @@
...
   - conda-forge
+  - r
   - defaults
 dependencies:
   - mapdamage2 =2.2.1
+  - r-inline
+  - r-ggplot2
+  - r-gam
+  - r-rcpp
+  - r-rcppgsl
```
dlaehnemann, line 12: From what I can see, all of these are dependencies of the mapdamage2 bioconda package, so they should be handled there:
https://bioconda.github.io/recipes/mapdamage2/README.html#package-Recipe

What was not working on your end? Or maybe you were using an older wrapper version with an older mapdamage2 version?

bio/mapdamage2/environment.yaml
```diff
@@ -1,6 +1,12 @@
 channels:
   - bioconda
   - conda-forge
+  - r
```
dlaehnemann, line 4: All the `R` packages are on `conda-forge` nowadays, so this channel should not be needed here. You can see that this is where the bioconda recipe gets its dependencies from the little anvil symbol next to the dependencies here:
https://bioconda.github.io/recipes/mapdamage2/README.html#package-Recipe

And you can find all those `R` packages here:
https://conda-forge.org/feedstocks/

dlaehnemann (commented):
I am not sure any of this should be necessary. What was the root problem that motivated these additions?

bio/mapdamage2/environment.yaml
fgvieira: I thought so too, but I got an error:
```
Missing the following R libraries 'inline', 'gam' and 'RcppGSL'
The Bayesian estimation has been disabled

Cannot use --rescale with missing R libraries
```

it seems someone had the same issue [here](https://github.com/ginolhac/mapDamage/issues/25), and they made another conda recipe but you need to specify a specifc build. Would that be better?

bio/mapdamage2/environment.yaml
fgvieira: Ok, just removed that dependency.

bio/mapdamage2/environment.yaml
```diff
@@ -4,3 +4,8 @@ channels:
   - defaults
 dependencies:
   - mapdamage2 =2.2.1
+  - r-inline
+  - r-ggplot2
+  - r-gam
+  - r-rcpp
+  - r-rcppgsl
```
dlaehnemann, line 11: From my understanding of the [`mapdamage2` install instructions in its GitHub README](https://github.com/ginolhac/mapDamage#bioconda-installation), this should work:
```suggestion
  - mapdamage2mapdamage2=2.2.1=pyr40_0
```
Although this also seems like something that should be fixable over at bioconda, rather than in this wrapper. But that is out of my depth, so I'll check in with someone more knowledgable.

bio/mapdamage2/environment.yaml
mbargull: @fgvieira, these dependencies will all get installed alongside `bioconda::mapdamage2=2.2.1`:
```
conda search --info bioconda::mapdamage2=2.2.1
Loading channels: done
mapdamage2 2.2.1 pyr40_0
------------------------
file name   : mapdamage2-2.2.1-pyr40_0.tar.bz2
name        : mapdamage2
version     : 2.2.1
build       : pyr40_0
build number: 0
size        : 48 KB
license     : MIT
subdir      : noarch
url         : https://conda.anaconda.org/bioconda/noarch/mapdamage2-2.2.1-pyr40_0.tar.bz2
md5         : d5e43d637bf9e55a3bb05e4b067fba72
timestamp   : 2020-06-01 19:18:46 UTC
dependencies: 
  - pysam
  - python >=3.5
  - r-base >=4.0,<4.1.0a0
  - r-gam
  - r-ggplot2 >=2.1.0
  - r-inline
  - r-rcpp
  - r-rcppgsl
  - seqtk
```
To me this looks like you haven't got the Conda environment activated, which means either the `mapDamage` or `Rscript` executable that are run are not the one from the environment `bioconda::mapdamage2=2.2.1` is installed into.

bio/mapdamage2/environment.yaml
fgvieira: That was how I initially had the environment, but @johanneskoester said it would be better with just the version and not the build string: https://github.com/snakemake/snakemake-wrappers/pull/186#discussion_r494117562

Well, just let me know how do you thing it would be best to have it.

dlaehnemann:
I just tried the following and it worked for me, including stating that it is `Performing Bayesian estimates`. So maybe this is some artefact of a conda setup somewhere? Or some other weird side effects?

```
mamba create -n mapdamage2 mapdamage2=2.2.1
conda activate mapdamage2
mapDamage --input test.sorted.bam --reference genome.fasta --folder test --rescale-out test.rescaled.bam
```

bio/mapdamage2/environment.yaml
dlaehnemann: Well, usually it's best not to pin too tightly. So optimally, this wrapper would even only pin to `  = mapdamage2 =2.2`. But I am at a loss what exactly is happening here. And both the bioconda package and the snakemake wrapper have passing tests. So I would think that the error you are seeing is something specific to your setup and I can't really help much with debugging there -- you'll have to dig deeper yourself.
And adding dependencies redundantly over here should not change anything in the general setup...

One option to try that sometimes helps, is completely removing your current conda installation and starting from scratch. It's kind of the nuclear option, but is has done the trick for me for some weird situations like this one.

fgvieira:
ok, so it seems it was because of a conflict between R packages.
It seems that, even though R and R packages are installed through conda, the `R_LIBS` variables takes precedence if it is defined.


Either way, I've removed the patch version and pinned it to minor version only.
best

dlaehnemann:
Ah, yeah. Thanks for the patience and keeping on digging on your end. This does seem messy, judging for the (stalling) attempts to fix this in conda:
https://github.com/conda-forge/r-base-feedstock/pull/65

dlaehnemann (approved):
This minimal change obivously makes sense, so I'll merge it!

mbargull:
> It seems that, even though R and R packages are installed through conda, the `R_LIBS` variables takes precedence if it is defined.

Ah, the `PYTHONPATH` of the R world. Yes, that makes sense; I hadn't thought about that!
@fgvieira, thank you very much for the followup -- that's quite helpful for debugging future issues of other users!

PR #258, opened 2020-11-10
Bedtools complement

AntonieV:
A new wrapper for bedtools complement


bio/bedtools/complement/meta.yaml
```diff
@@ -0,0 +1,12 @@
+name: bedtools complement
+description: >
+  ``Bedtools complement`` maps all regions of the genome which are not covered by the input.
```
dlaehnemann, line 3: I would add a link to the docs of this tools, here:
https://bedtools.readthedocs.io/en/latest/content/tools/complement.html

Same for the other `meta.yaml`s in this PR and in PR #261

bio/bedtools/complement/meta.yaml
```diff
@@ -0,0 +1,12 @@
...
+authors:
+  - Antonie Vietor
+input:
+  - BED/GFF/VCF files
+  - genome file (`genome file format <https://bedtools.readthedocs.io/en/latest/content/general-usage.html#genome-file-format>`_)
+output:
+  - .bed.complement or
+  - .vcf.complement or
+  - .gff.complement
```
dlaehnemann, line 12: I would always keep the file format extension last (see suggestion below in the Snakefile) and here I would simply keep the description as generic as for the input:
```suggestion
  - complemented BED/GFF/VCF file
```
Similarly in the other wrappers in these PRs.

bio/bedtools/complement/test/Snakefile
```diff
@@ -0,0 +1,27 @@
+rule bedtools_complement_bed:
+    input:
+        in_file="a.bed",
+        genome="genome_file"
```
dlaehnemann, line 4: Not having a file extension just looks awkward. Could we go for:
```suggestion
        genome="dummy.genome"
```
Users can then e.g. exchange `dummy` with something like `GRCh38` in their pipeline.

bio/bedtools/complement/test/Snakefile
```diff
@@ -0,0 +1,27 @@
...
+    output:
+        "results/bed-complement/a.bed.complement"
```
dlaehnemann, line 6: I would always keep the file format extension last:
```suggestion
        "results/bed-complement/a.complement.bed"
```
I'd also adjust the ordering in the log file name accordingly and in the other rules throughout these PRs.

dlaehnemann (changes requested):
Thanks, this all looks good!

I just have a couple of docs and cosmetic suggestions. These also apply to PR #261 .

dlaehnemann:
Just realized that this whole wrapper (bedtools complement) is also included in PR #261 , or am I missing something? If so, we can simply close this pull request, right?

PR #260, opened 2020-11-11
CACHE_VERSION to VERSION

aabiddanda:
In newer interations of ensembl-vep and `vep_install` the `--CACHE_VERSION` flag is now `--VERSION`, and this throws an error (this is now the default with `conda` installing ensembl-vep.


johanneskoester:
Thank you. I have slightly extended the fix, this time pinning to the vep version in the environment. Hence, closing in favor of PR #271

PR #264, opened 2020-11-13
updates author information and minimum fastp version

skurscheid:


bio/fastp/environment.yaml
```diff
@@ -3,4 +3,4 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - fastp ==0.20.0
+  - fastp >=0.20.0
```
johanneskoester, line 6: We cannot be sure about future releases. For provenance reasons, wrapper envs should always pin at least to minor versions.

johanneskoester:
```suggestion
  - fastp =0.20
```

johanneskoester (changes requested):

skurscheid (commented):
Sure, sounds absolutely reasonable and will keep this in mind!

johanneskoester (approved):

PR #266, opened 2020-11-16
Add normalization option to freebayes.

christopher-schroeder:
Sometimes freebayes does not normalize indels. This PR adds an option to activate bcftools norm for this issue,

### DIFF
--- a/bio/freebayes/wrapper.py
+++ b/bio/freebayes/wrapper.py
@@ -1,4 +1,4 @@
-__author__ = "Johannes Köster, Felix Mölder"
+__author__ = "Johannes Köster, Felix Mölder, Christopher Schröder"
 __copyright__ = "Copyright 2017, Johannes Köster"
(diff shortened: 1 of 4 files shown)


bio/freebayes/wrapper.py
```diff
@@ -11,10 +11,18 @@
 log = snakemake.log_fmt_shell(stdout=False, stderr=True)
 
 params = snakemake.params.get("extra", "")
+norm = snakemake.params.get("normalize", False)
+assert norm in [True, False]
 
 pipe = ""
 if snakemake.output[0].endswith(".bcf"):
     pipe = "| bcftools view -Ob -"
```
dlaehnemann, line 19: As both cases (norm is `True` and norm is `False`) are handled below, overwriting pipe, this line can be skipped:
```suggestion
```

dlaehnemann (approved):
Nice addition, with good defaults and all!

PR #267, opened 2020-11-16
chm_eval: remove symlink to chm_eval_kit output that does not exist in repo

dlaehnemann:
This seems like the current reason for tests still failing on the master branch.


dlaehnemann:
Hmm, the `chm_eval` test is also using the `chm_eval_kit` wrapper to get input input data, but doesn't seem to be able to load the other wrapper properly, as its data is not copied alongside `chm_eval`'s data. Can we somehow use the standard snakemake way of deploying wrappers for this test?

johanneskoester:
Fixed in commit 1c800958654163da4b372bfec3cc4e59e52c6698

johanneskoester:
Thank you anyway!

dlaehnemann:
Cheers for the cleanup!

PR #272, opened 2020-11-20
update bcftools view wrapper

mbhall88:
This wrapper didn't have a test. It also wasn't using log or threads. Also made the input and output a little more flexible.


johanneskoester (approved):
Thank you!

PR #281, opened 2021-01-25
Bamtools split wrapper.

Smeds:


Smeds:
Made a new more clean commit  (#282)

PR #282, opened 2021-01-26
Bamtools split wrapper.

Smeds:


bio/bamtools/split/test/Snakefile
```diff
@@ -0,0 +1,12 @@
+rule bamtools_split:
+    input:
+        "mapped/{sample}.bam",
+    output:
+        "mapped/{sample}.REF_xx.bam",
+    params:
+        #  -reference is the default value and can be left out, could also be '-paired' and '-tag <tag_name>'
+        split_type="-reference",
```
jafors, line 8: Could you add an `extra` field here, if someone wants to use other flags like `-refPrefix`?

```suggestion
    params:
        #  -reference is the default value and can be left out, could also be '-paired' and '-tag <tag_name>'
        split_type="-reference",
        extra=""
```

bio/bamtools/split/wrapper.py
```diff
@@ -0,0 +1,16 @@
+__author__ = "Patrik Smeds"
+__copyright__ = "Copyright 2021, Patrik Smeds"
+__email__ = "koester@jimmy.harvard.edu"
```
jafors, line 3: You can give your own mail address here. ;)

bio/bamtools/split/wrapper.py
```diff
@@ -0,0 +1,16 @@
...
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+
+split_type = snakemake.params.get("split_type", "-reference")
+
+if len(snakemake.input) != 1:
+    raise ValueError("One bam input file expected, got: " + str(len(snakemake.input)))
+
+shell("bamtools split -in {snakemake.input} " + split_type + " {log}")
```
jafors, line 16:
```suggestion
shell("bamtools split -in {snakemake.input} {split_type} {log}")
```

jafors (changes requested):
Thanks for the work, this should be a valuable addition! Find some small remarks below.

bio/bamtools/split/wrapper.py
Smeds: Done

bio/bamtools/split/test/Snakefile
Smeds: Changed split_type to extra

bio/bamtools/split/wrapper.py
Smeds: Changed split_type to extra

johanneskoester (approved):

PR #296, opened 2021-03-01
Added bam-output option

fgvieira:
Added `--bam-output` option to output  assembled haplotypes and locally realigned reads.

More info in:
https://gatk.broadinstitute.org/hc/en-us/articles/360037422451-HaplotypeCaller#--bam-output

### DIFF
--- a/bio/delly/wrapper.py
+++ b/bio/delly/wrapper.py
@@ -8,7 +8,9 @@ from snakemake.shell import shell
 
 
 exclude = (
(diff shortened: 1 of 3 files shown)


fgvieira:
It is giving a formatting error on `freebayes` but I did not change anything there.

jafors (changes requested):
Thanks for the additon. Maybe you could add the bam-output field in the example Snakefile? It can be commented out like the `known` input, just for people to easily understand how they can get the bam output.

jafors (approved):

PR #299, opened 2021-03-01
Fixed issue when input file is not required

fgvieira:
Fixed some analyses (e.g. model) where the `bam` file is not necessary.


jafors (approved):
Looks good to me!

PR #302, opened 2021-03-02
Added wrapper for GATK GenomicsDBImport

fgvieira:
Added wrapper for GATK `GenomicsDBImport` and updated the wrapper `GenotypeGVCFs` for compatibility.


jafors:
@brcopeland has promted a PR (#309) for the same tool, and both have some features that seem important/nicely done to me. Maybe you could join forces and create the "perfect" wrapper. :)

brcopeland:
I looked through both of these and I think it makes sense to use this version; it has the added functionality to update an existing GenomicsDB data store, which I did not implement.

However, there is a change to the `bio/delly` wrapper here which I assume is a mistake.

bio/gatk/genotypegvcfs/wrapper.py
```diff
@@ -14,9 +14,15 @@
 java_opts = get_java_opts(snakemake)
 
 log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+input_gvcf = snakemake.input.gvcf
+if os.path.isdir(input_gvcf):
+    input_gvcf = "gendb://" + input_gvcf
+
+
```
brcopeland, line 22: The purpose of the PR is to add a wrapper to run GenomicsDBImport.  I think this would appropriately be part of a different PR, which I have done in #310.

bio/delly/wrapper.py
```diff
@@ -8,7 +8,9 @@
 
 
 exclude = (
-    "-x {}".format(snakemake.input.exclude) if snakemake.input.get("exclude", "") else ""
+    "-x {}".format(snakemake.input.exclude)
+    if snakemake.input.get("exclude", "")
+    else ""
```
brcopeland, line 13: This change does not appear to be relevant to the PR.

bio/gatk/genotypegvcfs/environment.yaml
```diff
@@ -3,5 +3,5 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - gatk4 ==4.1.4.1
+  - gatk4 ==4.2.0.0
```
brcopeland, line 6: This change appears to be an unnecessary side effect.

brcopeland (commented):
Elaboration on prior comments.

bio/gatk/genotypegvcfs/wrapper.py
fgvieira: Yes, it could be a different PR, but had to place both on the same branch for testing purposes.

johanneskoester (approved):

PR #305, opened 2021-03-05
Added GATK VariantEval wrapper

fgvieira:


bio/gatk/varianteval/wrapper.py
```diff
@@ -0,0 +1,48 @@
...
+
+extra = snakemake.params.get("extra", "")
+java_opts = get_java_opts(snakemake)
+
+
+vcf = list(map("--eval {}".format, snakemake.input.vcf))
+
+bam = snakemake.input.get("bam", "")
+if bam:
+    bam = list(map("--input {}".format, bam))
```
jafors, line 17: Maybe we should allow `input.vcf` and `input.bam` to be strings if only one sample is used. You could check if they are strings or lists and act accordingly. Sth like

```
vcf = snakemake.input.vcf
if isinstance(vcf, str):
    vcf = "--eval  {}".format(vcf)
else:
    vcf = list(map("--eval {}".format, vcf))
```

bio/gatk/varianteval/wrapper.py
```diff
@@ -0,0 +1,48 @@
...
+
```
jafors, line 18: Maybe we should also add a `--comp` input for comparison vcfs, similar to the bam input.
See [here](https://gatk.broadinstitute.org/hc/en-us/articles/360040507171-VariantEval-BETA-#--comparison)

bio/gatk/varianteval/test/Snakefile
```diff
@@ -0,0 +1,16 @@
+rule gatk_varianteval:
+    input:
+        vcf=["calls/snvs.vcf"],
```
jafors, line 3: As mentioned in the `wrapper.py`, I feel this should be able to be a string in single sample runs.

jafors (changes requested):
Thanks for the addition! I have added some remarks, feel free to comment if you disagree!

johanneskoester (approved):

PR #306, opened 2021-03-08
Fixed small bug with log file and removed quiet flag

fgvieira:
* Remove quoting on log file, since it would assume it was another input file and actually not log anything
* Removed `quiet` flag, since it can be added in the `params`


jafors (changes requested):
Thanks! Maybe add `--quiet` to the params in the example Snakefile, so this change gets more prominently known to the user.

PR #310, opened 2021-03-12
Added the ability to run gatk GenotypeGVCFs on a GenomicsDB data store.

brcopeland:
Additionally added parameters to handle:
1. an input interval file,
2. dbSNP,
3. --tmp-dir


bio/gatk/genotypegvcfs/wrapper.py
```diff
@@ -12,11 +12,36 @@
...
+interval_file = snakemake.input.get("interval_file", "")
+if interval_file:
+    interval_file = "-L {}".format(interval_file)
+dbsnp = snakemake.input.get("known", "")
+if dbsnp:
+    dbsnp = "-D {}".format(dbsnp)
+
+# Allow for either an input gvcf or GenomicsDB
+gvcf = snakemake.input.get("gvcf", "")
+genomicsdb = snakemake.input.get("genomicsdb")
```
jafors, line 24:
```suggestion
genomicsdb = snakemake.input.get("genomicsdb", "")
```

jafors (changes requested):
Hey, thanks a lot for the additions! Maybe you could add some info to the example Snakefile, e.g. a comment that both `gvcf` or `genomicsdb` can be used as input.

bio/gatk/genotypegvcfs/wrapper.py
brcopeland: Good catch regarding getting the genomicsdb variable.  I made that change as well as added a comment to the Snakefile.

PR #311, opened 2021-03-15
Add log files to several wrappers

fgvieira:
Added `log` files to several wrappers.


jafors (approved):
Thanks for this log-athon! Looking good to me.

PR #312, opened 2021-03-22
update run_busco command to busco

sanjaynagi:
As of version 4.0.0 on onwards, the command to run busco is no longer run_busco as it used to be, and is now simply 'busco'. This breaks the wrapper. I have edited it to the correct 'busco'

### DIFF
--- a/bio/busco/environment.yaml
+++ b/bio/busco/environment.yaml
@@ -3,6 +3,6 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - python ==3.6
-  - busco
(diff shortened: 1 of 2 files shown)


jafors:
Thanks! Could you additionally fix busco to the latest version in the `environment.yaml`, so the wrapper will always use a version >=4.0.0.

sanjaynagi:
I have now fixed the busco version in environment.yaml to the latest version 5.0.0

johanneskoester (approved):

PR #314, opened 2021-03-23
BLAST makeblastdb

AntonieV:
A new wrapper for NIH BLAST to create local databases from nucleotide or protein reference sequences.


dlaehnemann (approved):
Just two small things. I'll simply apply them and see if the tests pass. If they do, I'll merge.

Thanks a bunch!

PR #316, opened 2021-03-23
Implemented running MarkDuplicates on multiple bams.

brcopeland:
As noted in https://gatk.broadinstitute.org/hc/en-us/articles/360035889471-How-should-I-pre-process-data-from-multiplexed-sequencing-and-multi-library-designs- :

> 2. Merge read groups and mark duplicates per sample (aggregation + dedup)
> 
> Once you have pre-processed each read group individually, you merge read groups belonging to the same sample into a single BAM file. You can do this as a standalone step, bur for the sake of efficiency we combine this with the per-sample duplicate marking step (it's simply a matter of passing the multiple inputs to MarkDuplicates in a single command).

### DIFF
--- a/bio/picard/markduplicates/meta.yaml
+++ b/bio/picard/markduplicates/meta.yaml
@@ -4,6 +4,6 @@ description: |
 authors:
   - Johannes Köster
 input:
-  - bam file
+  - bam file(s)
 output:
   - bam file with marked or removed duplicates
--- a/bio/picard/markduplicates/test/Snakefile
+++ b/bio/picard/markduplicates/test/Snakefile
@@ -1,6 +1,9 @@
 rule mark_duplicates:
     input:
         "mapped/{sample}.bam"
+	# optional to specify a list of BAMs; this has the same effect
+	# of marking duplicates on separate read groups for a sample
+	# and then merging
     output:
         bam="dedup/{sample}.bam",
         metrics="dedup/{sample}.metrics.txt"
--- a/bio/picard/markduplicates/wrapper.py
+++ b/bio/picard/markduplicates/wrapper.py
@@ -11,12 +11,16 @@ log = snakemake.log_fmt_shell(stdout=True, stderr=True)
 
 extra = snakemake.params
 java_opts = get_java_opts(snakemake)
+bams = snakemake.input
+if isinstance(bams, str):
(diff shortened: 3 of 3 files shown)


jafors:
In the updated wrapper, the `-I {input}` is indeed used instead of `--INPUT={input}`.
But maybe we should update the version in the environment.yaml (should be 2.25.0 at the moment) and use the "new" syntax throughout.

brcopeland:
Could you clarify where the `-I` is for me?  I've been trying to figure this out; I looked at https://github.com/snakemake/snakemake-wrappers/blob/master/bio/picard/markduplicates/wrapper.py which is what I believe it would be accessing.

brcopeland:
I made another small change because I was experiencing what I perceived as a bug; the wrapper previously passed all `params` (interpreting these as "extra") along to picard instead of only `params.extra`.  I think this is the intended functionality.

jafors:
> I made another small change because I was experiencing what I perceived as a bug; the wrapper previously passed all `params` (interpreting these as "extra") along to picard instead of only `params.extra`. I think this is the intended functionality.

Good thinking, but in this case, since `params` does not have a field `extra` in the example Snakefile, this would not include the given parameters. I think until there is no need for a parameter to be distinct from the rest (so you would need to have `params.extra` and `params.xyz`) , I would leave it as it was to not introduce a breaking change for people not using `params.extra` but just `params`.
When creating new wrappers though, I would suggest always using `extra` as a `params` field, it just looks "cleaner" to me.

brcopeland:
I could change that back but it would definitely break things for me; you can see in the current version all `params` are passed as `extra`, but this includes `params.java_opts`, which are then parsed and again passed in.

Additionally I am working on a PBS/TORQUE cluster and based my cluster profile on https://github.com/Snakemake-Profiles/pbs-torque/blob/master/%7B%7Bcookiecutter.profile_name%7D%7D/pbs-submit.py .  I recognize this may not be an optimal solution, but at line 139 there, I couldn't define `cluster` within a rule so I changed this to look in `params` instead.  The way this wrapper is written now passes on these additional parameters.

jafors:
Okay, didn't think about this, you are definitely correct! That's a pretty big reason for defining `params.extra`.

jafors (approved):

Issue #294, opened 2021-02-27
wrappers for hap.py and dipcall


nate-d-olson:
**Is your feature request related to a problem? Please describe.**
Difficulty developing an assembly benchmarking pipeline due to the dependencies associated with two key bioinformatic tools use in the pipeline hap.py and dipcall. I know there is a snakemake wrapper for the hap.py pre.py command but we are running into a regex issue when running the hap.py bioconda package in the snakemake/snakemake docker container. 

**Describe the solution you'd like**
We would like to have snakemake wrappers for hap.py, https://github.com/Illumina/hap.py, and dipcall, https://github.com/lh3/dipcall.

**Describe alternatives you've considered**
Tried using docker containers to handle dependencies but we were unable to get the pipeline to work on both macOS and Linux. Tried using the hap.py bioconda package but ran into a regex error when run in the snakemake/snakemake container. There is an issue on the hap.py github repo related to the error, https://github.com/Illumina/hap.py/issues/66, which seems to be related to specific gcc versions.

**Additional context**
As part of the Genome in a Bottle project, https://www.nist.gov/programs-projects/genome-bottle we are working to improve the usability and interpretability of our methods for benchmarking variant calls. Hap.py is used for small variant benchmarking and is the reference implementation of the GA4GH best practices for small variant benchmarking. dipcall is used to generate variant calls from a diploid assembly. We developed a proof of concept pipeline using snakemake, https://github.com/usnistgov/giab-asm-benchmarking, but the pipeline only works on macOS due to a hack we used to handle the hap.py and dipcall dependencies.

mbargull:
Hi @nate-d-olson,

> Tried using the hap.py bioconda package but ran into a regex error when run in the snakemake/snakemake container. There is an issue on the hap.py github repo related to the error, [Illumina/hap.py#66](https://github.com/Illumina/hap.py/issues/66), which seems to be related to specific gcc versions.

It looks like [version 0.3.12](https://github.com/Illumina/hap.py/releases/tag/v0.3.12) of hap.py [fixed the regex issue](https://github.com/Illumina/hap.py/pull/82) but we only got that version [into Bioconda](https://github.com/bioconda/bioconda-recipes/pull/20325) a couple of months after its release and [Snakemake wrapper creation](https://github.com/snakemake/snakemake-wrappers/pull/15).
I'll relay the need for the updated wrapper right away.
And please feel free to let us know of any further issue you might encounter after we get the updated wrapper up!

[...]

PR #317, opened 2021-03-26
Hap.py

nate-d-olson:
snakemake wrapper for main hap.py benchmarking functionality for https://github.com/snakemake/snakemake-wrappers/issues/294. This is my first attempt at creating a snakemake-wrapper, recommendations and suggestions for adhering to best practices are appreciated.


nate-d-olson:
Any recommendations for removing `params.prefix`, causing the code quality limiting error, when the prefix is used for multiple output files?

bio/hap.py/hap.py/environment.yaml
```diff
@@ -0,0 +1,7 @@
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
+dependencies:
+  - hap.py    = 0.3.14
+  - rtg-tools = 3.10.1
```
jafors, line 7:
```suggestion
  - hap.py ==0.3.14
  - rtg-tools ==3.10.1
```

jafors (changes requested):
Hey! I'll start with some linting issues (see below).

jafors:
> Any recommendations for removing `params.prefix`, causing the code quality limiting error, when the prefix is used for multiple output files?

I'd suggest inferring it from one of the output names. In your case, you would need to get the "results" from the complete file name. You can access wildcards, input and output inside of `params` using the a lambda function.
```
params:
    prefix=lambda wc, input, output: output[0].split('.')[0]
```
This should do the trick, since all output files start with the same prefix. You can however also be more specific and use for example `output["runinfo"]` instead of `output[0]`.

nate-d-olson:
Thanks @jafors for your help addressing the code quality issues. In this case for the outputs is it better to use the `multiext` function to ensure the output files have all the same prefix or define them individually. Then use the lambda expression to define the suggested lambda expression to define the prefix.

bio/hap.py/hap.py/wrapper.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Extract arguments
+extra = snakemake.params.get("extra", "")
+
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+
+shell(
+    "(hap.py"
+    " --threads {snakemake.threads}"
+    " --engine {snakemake.params.engine}"
```
jafors, line 36: If I understand hap.py correctly, there is a default engine which is used if the `--engine` parameter is not given.
If that's the case, we also should have a default value for this, or even easier, just ignore the parameter if it is not given.
Sth. like:
```
engine = snakemake.params.get("engine", "")
if engine:
    engine = "--engine {}".format(engine)

shell(
    "(hap.py"
    " --threads {snakemake.threads}"
    " {engine}"
   ....
```

bio/hap.py/hap.py/wrapper.py
```diff
@@ -0,0 +1,45 @@
...
+    " -r {snakemake.input.genome}"
+    " {extra}"
+    " -f {snakemake.input.truth_regions}"
+    " --stratification {snakemake.input.strats}"
```
jafors, line 40: Similar to the engine, I think those are optional inputs, so the wrapper should be functional without them. You can use similar structures with `get()` as with params.

jafors (changes requested):
In total, the wrapper looks already pretty good, but since hap.py has multiple optional arguments, we could go for some generalizations.

bio/hap.py/hap.py/wrapper.py
nate-d-olson: Since `truth_regions` and `stratifications` are input files, can I use `get()` with `snakemake.input` like this
 
```
truth_regions = snakemake.input.get("truth_regions")
if truth_regions:
    truth_regions = "-f {}".format(truth_regions)
```

jafors: Yes, this should work, you can use it with `input` as with `params`.

johanneskoester (approved):
Excellent work!

jafors (approved):

PR #318, opened 2021-03-30
Automatically detect threads and output format on bcftools

fgvieira:
These options are quite common across `bcftools` tools (and there might be others, like `tmp_dir` or `max_mem`) so, it might make sense to have them parsed on `snakemake_wrapper_utils`, just like it is currently done for `java`. The same could be done for `samtools`, (e.g.):

- `snakemake_wrapper_utils.bcftools`
- `snakemake_wrapper_utils.samtools`

### DIFF
--- a/bio/bcftools/concat/test/Snakefile
+++ b/bio/bcftools/concat/test/Snakefile
@@ -7,7 +7,6 @@ rule bcftools_concat:
         "logs/all.log"
     params:
         uncompressed_bcf = False,
-        bcftools_use_mem = False,
         extra = ""  # optional parameters for bcftools concat (except -o)
     threads: 4
     resources:
--- a/bio/bcftools/concat/test/Snakefile
+++ b/bio/bcftools/concat/test/Snakefile
@@ -3,6 +3,8 @@ rule bcftools_concat:
         calls=["a.bcf", "b.bcf"]
     output:
         "all.bcf"
+    log:
+        "logs/all.log"
     params:
         uncompressed_bcf = False,
         bcftools_use_mem = False,
--- a/bio/bcftools/concat/test/Snakefile
+++ b/bio/bcftools/concat/test/Snakefile
@@ -1,4 +1,4 @@
-B65;6003;1crule bcftools_concat:
+rule bcftools_concat:
     input:
         calls=["a.bcf", "b.bcf"]
     output:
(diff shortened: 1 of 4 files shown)


johanneskoester (approved):

johanneskoester:
> These options are quite common across `bcftools` tools (and there might be others, like `tmp_dir` or `max_mem`) so, it might make sense to have them parsed on `snakemake_wrapper_utils`, just like it is currently done for `java`. The same could be done for `samtools`, (e.g.):
> 
> * `snakemake_wrapper_utils.bcftools`
> * `snakemake_wrapper_utils.samtools`

This is a very good idea. Do you want to provide a PR?

fgvieira:
Sure... I can give it a try. 😄

bio/bcftools/concat/test/Snakefile
```diff
@@ -1,9 +1,14 @@
-rule bcftools_concat:
+B65;6003;1crule bcftools_concat:
```
johanneskoester, line 1: What happened here?

johanneskoester (changes requested):
snakemake-wrapper-utils 0.2.0 has been released, you can add it to the env and make use of it.

bio/bcftools/concat/test/Snakefile
fgvieira: Uff, no idea... :)

fgvieira:
@johanneskoester I am getting an error that seems to be related to `snakemake`:
`ModuleNotFoundError: No module named 'stopit'`

bio/bcftools/concat/test/Snakefile
```diff
@@ -3,7 +3,14 @@ rule bcftools_concat:
         calls=["a.bcf", "b.bcf"]
     output:
         "all.bcf"
+    log:
+        "logs/all.log"
     params:
-        ""  # optional parameters for bcftools concat (except -o)
+        uncompressed_bcf = False,
+        bcftools_use_mem = False,
```
johanneskoester, line 10:
```suggestion
        bcftools_use_mem = False,  # if True, limit memory to given mem_mb resource
```
Is that the intended meaning of this? Maybe the parameter should be renamed to `bcftools_limit_mem` then?

johanneskoester: I know, this would need a change in snakemake-wrapper-utils

johanneskoester (changes requested):

fgvieira:
Yes, the idea was to disable memory parsing, since `bcftools concat` does not support it.
However, I think it is better to disable it in the wrapper.

johanneskoester (approved):

PR #320, opened 2021-03-30
Added automated detection of output BAI and SBI files to GATK MarkDuplicatesSpark wrapper 

fgvieira:


bio/gatk/markduplicatesspark/wrapper.py
```diff
@@ -14,6 +14,17 @@
...
+
+if snakemake.output.get("bai"):
+    bai = "true"
+else:
+    bai = "false"
+
+if snakemake.output.get("sbi"):
+    sbi = "true"
+else:
+    sbi = "false"
```
jafors, line 26: Those params are set to `true` per default. Why would you need to infer them from the output files and require the user to explicitly give the index files as output?

If you would not like to index the bam, you could just add `--create-output-bam-index false` to the `extra` param.
And if the user wants to have the `.bai` file as an output, it is no problem to just add it to the output files without changing the wrapper.

jafors (changes requested):

bio/gatk/markduplicatesspark/wrapper.py
fgvieira: Yes, and that means that those files are always created, even if they are not in the `output`. My idea was to have those options dependent on the rule's output files. So, if the user does not specifies them in the output, they are not created. This would be a bit more error safe, as no files are created without the rule specifically asking for them.

On the other hand, this would also be more explicit, since the command line would clearly state `--create-output-bam-index true/false` (if `-p`).

That said, this is a minor change since, as you say, those options can be given through the `extra` param.

jafors: True. Could you then add `sbi` and `bai` to the output of the example Snakefile so that people know these paramteres exist and should be set? Thanks!

fgvieira: However, you cannot specify the `sbi` and `bai` file names, meaning that you can get a `MissingOutputException` if you specify `sbi` and `bai` in `output` (if the names do not match). :disappointed:

Unless we rename these files at the end, if the file names are different.

Do you think it is worth it?

jafors: Hmm, in this case, I'd also rather trust that users give the correct name for the sbi and bai output.
I'm still not 100% convinced we need to restrict the creation of those files to rules with explicit output files given, but I guess if it is made clear in the Snakefile that they need to be given in a specific way (like depending on the name of the bam output) it should be ok.

fgvieira: Yeah... maybe if/when GATK allows to specify `bai` and `sbi` file names this will make more sense.
I'll close it.

PR #327, opened 2021-04-02
Added support for sorting and output formats on Minimap2 wrapper

fgvieira:
If I do not specify the `sort` option, I get an error:
```
  File "/lustre/hpc/hologenomics/fgvieira/NGS/snakemake-wrappers/bio/minimap2/aligner/test/.snakemake/scripts/tmpigon5oti.wrapper.py", line 57, in <module>
    raise ValueError("Unexpected value for params.sort ({})".format(sort))
ValueError: Unexpected value for params.sort (functools.partial(<function Namedlist._used_attribute at 0x7f8eb0aa7dc0>, _name='sort'))
```

It seems sort is being set as `functools.partial(<function Namedlist._used_attribute at 0x7f8eb0aa7dc0>, _name='sort')`, even though it should be set to `none` by default:
```
sort = snakemake.params.get("sort", "none")
```

Any idea why?


dlaehnemann:
If sort is not specified at all, it should default to `"none"`. But are you referring to the case where you set `sort = ""` the empty string? I guess this would mean that sort is actually set to the empty string, so that the equality `if sort == "none"` doesn't hold. Could that be it? If so, maybe we should explicitly handle this (and default to "none" for the empty string, as well)?

fgvieira:
I get that error if `params.sort` is not specified at all:
```
rule minimap2_bam:
    input:
        target="target/{input1}.mmi", # can be either genome index or genome fasta
        query=["query/reads1.fasta", "query/reads2.fasta"]
    output:
        "aligned/{input1}_aln.bam"
    log:
        "logs/minimap2/{input1}.log"
    params:
        extra="-x map-pb",           # optional
        sort_extra=""                # optional: extra arguments for samtools/picard
    threads: 3
    wrapper:
        "master/bio/minimap2/aligner"
```

It seems as if the default `none` is not being set. Can you double check on your side?

jafors:
Seems like `params.get` is not a big fan of params named `sort`. If I rename it to `sorting`, everything seems to work. :smile:

jafors:
I get the same errors. Renaming the parameter seems to fix that, however. `Sort` seems to be an unlucky choice of name.

fgvieira:
Any suggestions?

bio/minimap2/aligner/test/Snakefile
```diff
@@ -7,7 +7,41 @@ rule minimap2:
     log:
         "logs/minimap2/{input1}.log"
     params:
-        extra="-x map-pb"  # optional
+        extra="-x map-pb",           # optional
+        sort="coordinate",           # optional: Enable sorting. Possible values: 'none', 'queryname' or 'coordinate'
```
jafors, line 11:
```suggestion
        sorting="coordinate",           # optional: Enable sorting. Possible values: 'none', 'queryname' or 'coordinate'
```

bio/minimap2/aligner/test/Snakefile
```diff
@@ -7,7 +7,41 @@ rule minimap2:
...
+    input:
+        target="target/{input1}.mmi", # can be either genome index or genome fasta
+        query=["query/reads1.fasta", "query/reads2.fasta"]
+    output:
+        "aligned/{input1}_aln.sam"
+    log:
+        "logs/minimap2/{input1}.log"
+    params:
+        extra="-x map-pb",           # optional
+        sort="none",                 # optional: Enable sorting. Possible values: 'none', 'queryname' or 'coordinate'
```
jafors, line 27:
```suggestion
        sorting="none",                 # optional: Enable sorting. Possible values: 'none', 'queryname' or 'coordinate'
```

bio/minimap2/aligner/test/Snakefile
```diff
@@ -7,7 +7,41 @@ rule minimap2:
...
+    input:
+        target="target/{input1}.mmi", # can be either genome index or genome fasta
+        query=["query/reads1.fasta", "query/reads2.fasta"]
+    output:
+        "aligned/{input1}_aln.bam"
+    log:
+        "logs/minimap2/{input1}.log"
+    params:
+        extra="-x map-pb",           # optional
+        sort="coordinate",           # optional: Enable sorting. Possible values: 'none', 'queryname' or 'coordinate'
```
jafors, line 43:
```suggestion
        sorting="coordinate",           # optional: Enable sorting. Possible values: 'none', 'queryname' or 'coordinate'
```

bio/minimap2/aligner/wrapper.py
```diff
@@ -3,14 +3,50 @@
...
 inputQuery = " ".join(snakemake.input.query)
 
+# Extract output format
+out_name, out_ext = path.splitext(snakemake.output[0])
+out_ext = out_ext[1:].upper()
+
+# Extract arguments.
+extra = snakemake.params.get("extra", "")
+
+sort = snakemake.params.get("sort", "none")
```
jafors, line 19:
```suggestion
sort = snakemake.params.get("sorting", "none")
```

jafors (commented):
This seems to work on my local machine.

jafors (approved):

PR #329, opened 2021-04-03
Fixed bug when specifying queryname sorting 

fgvieira:
and removed tmp file specification, since it can be done through `sort_extra`.

### DIFF
--- a/bio/bwa/mem/wrapper.py
+++ b/bio/bwa/mem/wrapper.py
@@ -5,6 +5,7 @@ __license__ = "MIT"
 
 
 from os import path
+import re
 import tempfile
 from snakemake.shell import shell
 
@@ -15,6 +16,11 @@ extra = snakemake.params.get("extra", "")
 sort = snakemake.params.get("sorting", "none")
 sort_order = snakemake.params.get("sort_order", "coordinate")
 sort_extra = snakemake.params.get("sort_extra", "")
+
+if re.search(r"-T\b", sort_extra) or re.search(r"--TMP_DIR\b", sort_extra):
+    sys.exit(
+        "You have specified temp dir (`-T` or `--TMP_DIR`) in params.sort_extra; this is automatically set from params.tmp_dir."
+    )
 
 log = snakemake.log_fmt_shell(stdout=False, stderr=True)
 
@@ -53,8 +59,8 @@ elif sort == "picard":
 
     # Sort alignments using picard SortSam.
     pipe_cmd = (
-        "picard SortSam {sort_extra} INPUT=/dev/stdin"
-        " OUTPUT={snakemake.output[0]} SORT_ORDER={sort_order} TMP_DIR={tmp}"
+        "picard SortSam {sort_extra} --INPUT /dev/stdin"
+        " --OUTPUT {snakemake.output[0]} --SORT_ORDER {sort_order} --TMP_DIR {tmp}"
     )
 
 else:
--- a/bio/bwa/mem/wrapper.py
+++ b/bio/bwa/mem/wrapper.py
@@ -5,18 +5,24 @@ __license__ = "MIT"
 
 
 from os import path
-
+import tempfile
 from snakemake.shell import shell
 
 
 # Extract arguments.
 extra = snakemake.params.get("extra", "")
 
-sort = snakemake.params.get("sort", "none")
+sort = snakemake.params.get("sorting", "none")
 sort_order = snakemake.params.get("sort_order", "coordinate")
 sort_extra = snakemake.params.get("sort_extra", "")
 
 log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+
+tmp_prefix = path.splitext(snakemake.output[0])[0]
+tmp_dir = snakemake.params.get("tmp_dir")
+if tmp_dir:
+    tempfile.tempdir = tmp_dir
+
 
 # Check inputs/arguments.
 if not isinstance(snakemake.input.reads, str) and len(snakemake.input.reads) not in {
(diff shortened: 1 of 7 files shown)


bio/bwa/mem/wrapper.py
```diff
@@ -36,15 +36,12 @@
...
-    pipe_cmd = "samtools sort {sort_extra} -o {snakemake.output[0]} -"
-
     # Add name flag if needed.
     if sort_order == "queryname":
         sort_extra += " -n"
 
-    prefix = path.splitext(snakemake.output[0])[0]
-    sort_extra += " -T " + prefix + ".tmp"
+    # Sort alignments using samtools sort.
+    pipe_cmd = "samtools sort {sort_extra} -o {snakemake.output[0]} -"
```
dlaehnemann, line 44: Having this after potentially adding to `sort_extra` is a good idea. But I think the `-T` prefixing should remain in there, as well, as this will avoid multiple sorts happening in the same folder from accidentally generating the same `.tmp` files. So I'd also keep this:
```
    prefix = path.splitext(snakemake.output[0])[0]
    sort_extra += " -T " + prefix + ".tmp
```

dlaehnemann (changes requested):
Good catch and nice docs extensions.

However, I think keeping the `-T` setting is required to avoid race conditions.

bio/bwa/mem/wrapper.py
fgvieira: The reason I removed it was that the user might sometimes want to have the temp files in another place (e.g. local scratch), rather than where the output is; and it can just be specified through `params.extra`.

But I guess we could check if the tmp folder is specified in `params.extra` and, if not, default to that.
Do you think that would make sense?

dlaehnemann: Ah, yes. Very good point. That makes perfect sense.

Something like `if "-T " in params.extra` should do the trick.

fgvieira: but actually, samtools already has a default temp:
```
-T PREFIX
Write temporary files to PREFIX.nnnn.bam, or if the specified PREFIX is an existing directory, to PREFIX/samtools.mmm.mmm.tmp.nnnn.bam, where mmm is unique to this invocation of the sort command.

By default, any temporary files are written alongside the output file, as out.bam.tmp.nnnn.bam, or if output is to standard output, in the current directory as samtools.mmm.mmm.tmp.nnnn.bam.
```

So it might be ok as is, no?

johanneskoester: I think samtools' default is not a good one, becuase we frequently see leftover files from that in case of crashes. Therefore, this rule should actually always default to some Python-driven TemporaryDirectory for the -T param, and not even expose it in my opinion.  What do you think?

dlaehnemann: I think I still had the problem in mind from older versions, where specifying a (tmp) folder as a prefix did not generate a unique infix. So it seems the situation is already much better. But I agree with @johanneskoester that handling this with a [python temporary directory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) is best to avoid a mess after unexpected interruptions of job executions. This should give the directory as a prefix:
```
import tempfile

with tempfile.TemporaryDirectory() as prefix:
```

And then maybe add a check to `params.extra`, that `-T ` should not be used there.

fgvieira: Added python `tempdir` to both `samtools` and `picard`

bio/bwa/mem/wrapper.py
```diff
@@ -5,19 +5,31 @@
...
 sort_extra = snakemake.params.get("sort_extra", "")
 
+if re.search(r"-T\b", sort_extra) or re.search(r"--TMP_DIR\b", sort_extra):
+    sys.exit(
+        "You have specified temp dir (`-T` or `--TMP_DIR`) in params.sort_extra; this is automatically set from params.tmp_dir."
+    )
+
 log = snakemake.log_fmt_shell(stdout=False, stderr=True)
 
+tmp_prefix = path.splitext(snakemake.output[0])[0]
```
dlaehnemann, line 27: I don't think the `tmp_prefix` is needed. The `TemporaryDirectory()` will be unique and will exist once samtools is called, so there cannot be any race conditions, if I understand the handling of existing directories correctly from the [samtools sort manual](https://www.htslib.org/doc/samtools-sort.html).

```suggestion
```

bio/bwa/mem/wrapper.py
```diff
@@ -36,32 +48,30 @@
...
-    pipe_cmd = "samtools sort {sort_extra} -o {snakemake.output[0]} -"
-
     # Add name flag if needed.
     if sort_order == "queryname":
         sort_extra += " -n"
 
-    prefix = path.splitext(snakemake.output[0])[0]
-    sort_extra += " -T " + prefix + ".tmp"
+    # Sort alignments using samtools sort.
+    pipe_cmd = "samtools sort -T {tmp}/{tmp_prefix}.tmp {sort_extra} -o {snakemake.output[0]} -"
```
dlaehnemann, line 56: In line with my above comment, I would simply go for:
```suggestion
    pipe_cmd = "samtools sort -T {tmp} {sort_extra} -o {snakemake.output[0]} -"
```

bio/bwa/mem/wrapper.py
```diff
@@ -5,19 +5,31 @@
...
+    sys.exit(
+        "You have specified temp dir (`-T` or `--TMP_DIR`) in params.sort_extra; this is automatically set from params.tmp_dir."
+    )
+
 log = snakemake.log_fmt_shell(stdout=False, stderr=True)
 
+tmp_prefix = path.splitext(snakemake.output[0])[0]
+tmp_dir = snakemake.params.get("tmp_dir")
+if tmp_dir:
+    tempfile.tempdir = tmp_dir
```
dlaehnemann, line 30: That's a neat trick! :D

test.py
```diff
@@ -76,7 +76,7 @@ def run(wrapper, cmd, check_log=None):
         os.chdir(testdir)
         if os.path.exists(".snakemake"):
             shutil.rmtree(".snakemake")
-        cmd = cmd + ["--wrapper-prefix", "file://{}/".format(d), "--conda-cleanup-pkgs"]
+        cmd = cmd + ["--wrapper-prefix", "file://{}/".format(d), "--conda-cleanup-pkgs", "--printshellcmds"]
```
dlaehnemann, line 79: Is this meant as a permanent change, or was this just for testing locally?

dlaehnemann (changes requested):
This generally looks good to me, thanks a bunch. Just two minor things.

test.py
fgvieira: I thought of it as a permanent change, as it can help debugging when you can see exactly what command is being run.

bio/bwa/mem/wrapper.py
fgvieira: I am not totally sure but, as I understood it, `TemporaryDirectory()` will return a temp directory (and create it) while `samtools sort` require a prefix (directory + part of the name). That was why I included the `tmp_prefix`.

But maybe it is not necessary...

fgvieira: I'll accept your suggestions and we'll see if it works! :smile:

test.py
dlaehnemann: You're right. Also, this is only displayed in the logs for failing tests. So it should be fine!

dlaehnemann (approved):

PR #332, opened 2021-04-06
Deepvariant 1.0

nikostr:
This PR updates the deepvariant wrapper to v1.0.0. Since the bioconda deepvariant recipe now supports GPU acceleration this means performance improvements both with regards to variant calling performance and runtime. 

The PR also adds support for generating gVCFs, includes a test for this, and resolves a bug in how the `basename` variable is computed.


johanneskoester (approved):

johanneskoester:
Thanks!

PR #333, opened 2021-04-08
Updates bwa-mem samtools to v1.12, adds `--write-index` flag

nikostr:
Hey! The bwa mem wrapper is using samtools 1.9 currently. The latest version is 1.12, and it provides the nifty `--write-index` flag. I've added an optional parameter to add this flag in the wrapper, as well as a test to see that this works. Solved this by letting the param take the value `True` to add the flag - not sure if this is the best way to do it, but it seems to work.

### DIFF
--- a/bio/bwa/mem/environment.yaml
+++ b/bio/bwa/mem/environment.yaml
@@ -4,5 +4,5 @@ channels:
   - defaults
 dependencies:
   - bwa ==0.7.17
-  - samtools ==1.9
+  - samtools ==1.12
   - picard ==2.20.1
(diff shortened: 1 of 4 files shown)


jafors:
Hi! Thanks for the addition. `--write-index` is indeed a nice feature in the new samtools version, but I feel that if I wanted to use it here, I would just add it in the rule definition. Writing:
`sort_extra="--write-index"`
should basically yield the same result as your version with the additional parameter flag. Any objections?

The samtools version update, however, is a very good idea.

nikostr:
No objections, I guess it potentially makes it more transparent :) I'll get this sorted.

nikostr:
Okay @jafors, this should be good to go now! The wrapper has been restored, so the PR now consists of the version update and a test case for index writing.

jafors (approved):

PR #336, opened 2021-04-13
Updates deepvariant to v1.1.0

nikostr:
This PR updates DeepVariant to v1.1.0. The upstream bioconda recipe does not yet support e.g. DeepTrio, but the PR https://github.com/bioconda/bioconda-recipes/pull/27988 adds the missing `--sample_name` argument to the `dv_make_examples.py` wrapper. This in turn avoids problems downstream, as samples in vcfs may have other names than `default`.

### DIFF
--- a/bio/deepvariant/environment.yaml
+++ b/bio/deepvariant/environment.yaml
@@ -2,4 +2,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - deepvariant=1.1.0
+  - deepvariant ==1.1.0
--- a/bio/deepvariant/environment.yaml
+++ b/bio/deepvariant/environment.yaml
@@ -2,4 +2,4 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - deepvariant=1.0.0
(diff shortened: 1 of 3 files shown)


nikostr:
Given the current wrapper status, the sample name in the vcf would be set to the `basename` variable, i.e. the filename of the incoming bam file, excluding the file ending. I think it might make sense to let users set this using a parameter that defaults to the current default setting to avoid breakage, but that would still allow users to control sample naming by other means than by the name of the input bam file. Should this be a separate PR?

jafors:
Feel free to go on with any other additions to deepvariant in this PR!

nikostr:
Okay, I think this should be it :) For now, the `sample_name` parameter has no real effect on the output - it is only used in the names of temporary files - but as soon as https://github.com/bioconda/bioconda-recipes/pull/27988 is approved the actual vcf samples names should be updated accordingly.

nikostr:
Now the upstream has been merged, and I've tested this locally - it works beautifully! Feel free to merge :)

jafors (approved):

PR #338, opened 2021-04-16
Update BUSCO to v5.1.2

nikostr:
This updates BUSCO from v5.0.0 to v5.1.2, and resolves issues with the old wrapper. Since the BUSCO output structure has been updated (since v4?), the old wrapper no longer works. The output of the wrapper now needs to be a directory.

An improvement to the wrapper is an optional `download_path` parameter, allowing the user to specify a download path for BUSCO, e.g  in the `resources` directory.

I have not yet tested how this performs when multiple first-time runs are initiated. There may be issues if the same download path is used for all of them.


bio/busco/wrapper.py
```diff
@@ -16,22 +17,19 @@
...
-    "busco --in {snakemake.input} --out {out_name} --force "
-    " --cpu {snakemake.threads} --mode {mode} --lineage {lineage} "
-    " {extra} {log}"
-)
+with tempfile.TemporaryDirectory(prefix="busco_", dir="") as tmp_dir:
+    shell(
+        "busco --in {snakemake.input} --out {tmp_dir} --force "
+        " --cpu {snakemake.threads} --mode {mode} --lineage {lineage} "
+        + download_path
+        + " {extra} {log}"
```
jafors, line 31:
```suggestion
        "--cpu {snakemake.threads} --mode {mode} --lineage {lineage} "
        "{download_path} "
        "{extra} {log}"
```

You can use `{download_path}` directly in the command string just as `{mode}` or `{lineage}`.

bio/busco/wrapper.py
```diff
@@ -16,22 +17,19 @@
 assert lineage is not None, "please input the path to a lineage for busco assessment"
 
 # busco does not allow you to direct output location: handle this by moving output
-outdir = path.dirname(snakemake.output[0])
-if "/" in outdir:
-    out_name = path.basename(outdir)
-else:
-    out_name = outdir
+outdir = snakemake.output[0]
```
jafors, line 20: If I read the busco manual correctly, they do not advice to give a path here, just a `prefix` name for output folders and files. The  top-level  path leading to the results directory needs to be given by `--out_path`.
So if I want my results to be in sth. like `results/busco/sample`, I would need to call busco with `--out sample --out_path results/busco`.
Am I correct here? If so, you should think about how to parse the snakemake output into those two params, maybe using `os.path`.

bio/busco/wrapper.py
```diff
@@ -16,22 +17,19 @@
...
+        + " {extra} {log}"
+    )
 
-busco_outname = "run_" + out_name
-
-# move to intended location
-shell("cp -r {busco_outname}/* {outdir}")
-shell("rm -rf {busco_outname}")
+    # move to intended location
+    shell("cp -r {tmp_dir} {outdir}")
```
jafors, line 35: I think this is not necessary if the output options are set correctly (see above).

jafors (changes requested):
Thanks a bunch. I made a few comments below.

bio/busco/wrapper.py
nikostr: Good catch! This is resolved by https://github.com/snakemake/snakemake-wrappers/pull/338/commits/f9a582a7d3ef218ab5a63fa4fdfe533c99b028f9. Much neater solution.

nikostr:
While I was at it I took the liberty of renaming `lineage_path` to `lineage`, as it is not a path that is being given there.

jafors (approved):

PR #340, opened 2021-04-17
Updata multiqc version to 1.10.1

Direct-A:
Updata multiqc version to 1.10.1

multiqc 1.9 could not parse *.zip file, which is the result of fastqc.

### DIFF
--- a/bio/multiqc/environment.yaml
+++ b/bio/multiqc/environment.yaml
@@ -3,4 +3,4 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - multiqc == 1.10.1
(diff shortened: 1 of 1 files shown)


bio/multiqc/environment.yaml
```diff
@@ -3,4 +3,4 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - multiqc ==1.9
+  - multiqc == 1.10.1
```
jafors, line 6:
```suggestion
  - multiqc ==1.10.1
```

jafors (approved):

PR #345, opened 2021-04-22
Add cosmic database download wrapper

tedil:
This PR adds a wrapper for downloading files from [COSMIC](https://cancer.sanger.ac.uk/cosmic/download).
It expects the following parameters to be set:
 - `build` (at the moment either "GRCh37" or "GRCh38")
 - `dataset` (at the moment either "cosmic" or "cell_lines")
 - `version` (at the moment "v83" to "v92" or "latest")
 - `file` (which in the example Snakefile is derived from the output via the `{db}` wildcard)

Since new builds, datasets, versions and files may be added in the future, this wrapper will query the endpoints in order to check if the requested combination is actually available (at the time this wrapper is executed).

### DIFF
--- a/bio/cosmic-db/test/Snakefile
+++ b/bio/cosmic-db/test/Snakefile
@@ -1,0 +1,16 @@
+envvars:
+    "COSMIC_EMAIL",
+    "COSMIC_PW"
+
+rule cosmic_download:
+    output:
+        "resources/{db}"
+    params:
+        build="GRCh38",
+        dataset="cosmic",
+        version="v92",
+        file=lambda wc: wc.db  # e.g. "CosmicHGNC.tsv.gz"
+    log:
+        "logs/cosmic-db/{db}.log"
+    wrapper:
+        "master/bio/cosmic-db"
--- a/bio/cosmic-db/environment.yaml
+++ b/bio/cosmic-db/environment.yaml
@@ -1,0 +1,8 @@
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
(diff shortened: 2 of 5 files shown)


tedil:
The tests fail because no credentials (i.e. environment variables `COSMIC_EMAIL` and `COSMIC_PW`) are defined. Since we probably don't want a dummy account stored in this repo's github secrets, it's best to use either a little flask app which serves as ta mock restful api for the requests in this wrapper, or use a more general approach such as https://github.com/stoplightio/prism and have that be available for any wrapper.

dlaehnemann:
I have to admit I don't fully understand the solutions you propose, but is that something that would make sense to have in [snakemake-wrapper-utils](https://github.com/snakemake/snakemake-wrapper-utils)?

tedil:
Yes, I guess that'd be a good place to put it.

bio/cosmic-db/test/Snakefile
```diff
@@ -0,0 +1,16 @@
...
+    "COSMIC_PW"
+
+rule cosmic_download:
+    output:
+        "resources/{db}"
+    params:
+        build="GRCh38",
+        dataset="cosmic",
+        version="v92",
+        file=lambda wc: wc.db  # e.g. "CosmicHGNC.tsv.gz"
```
johanneskoester, line 12: Maybe add some comment here explaining the meaning of file? And where to get the possible values from?

johanneskoester (changes requested):

PR #348, opened 2021-04-28
Bcftools filter output format

### LINKED ISSUE #347: bcftools filter requires output file ending in bcf.gz to create compressed bcfs
**Snakemake version**
This applies to the current snakemake-wrappers version (0.74.0).

**Describe the bug**
The `bcftools/filter` wrapper infers output format based on the output file ending. There does not seem to be a consensus regarding how to name uncompressed bcfs (https://github.com/samtools/hts-specs/issues/114) and adding a `.gz` to the output file name may even break certain tools (https://github.com/Clinical-Genomics/MIP/issues/516).

Not sure how to handle this in the best way. Maybe an optional `output_format` parameter that overrides the default wrapper behavior?

nikostr:
This PR fixes #347 by adding an option `output_format` parameter. If the param is not set, it defaults to behavior of the previous wrapper. This avoids problems related to the `bcf.gz` file ending.


fgvieira:
Hi @nikostr, not sure if you noticed, but there is a `snakemake-wrapper-utils` for `bcftools` that takes care of output format:
https://github.com/snakemake/snakemake-wrapper-utils/blob/master/snakemake_wrapper_utils/bcftools.py

nikostr:
Thanks for pointing this out @fgvieira ! Now updated the PR to use `snakemake-wrapper-utils`.

bio/bcftools/filter/test/Snakefile
```diff
@@ -40,14 +40,15 @@ rule bcf_filter_o_bcf:
...
     input:
         "{prefix}.bcf",
     output:
-        "{prefix}.filter.bcf.gz",
+        "{prefix}.filter.uncompressed.bcf",
     log:
-        "log/{prefix}.filter.bcf.gz.log",
+        "log/{prefix}.filter.uncompressed.bcf.log",
     params:
+        uncompressed_bcf = True,
```
jafors, line 51:
```suggestion
        uncompressed_bcf=True,
```

jafors (approved):
Very cool, thanks.

PR #353, opened 2021-05-06
Added wrapper for PrintReadsSpark

fgvieira:


bio/gatk/printreadsspark/wrapper.py
```diff
@@ -0,0 +1,27 @@
...
+)
+spark_extra = snakemake.params.get("spark_extra", "")
+java_opts = get_java_opts(snakemake)
+
+
+shell(
+    "gatk --java-options '{java_opts}' PrintReadsSpark {extra} "
+    "--reference {snakemake.input.ref} --input {snakemake.input.bam} "
+    "--output {snakemake.output.bam} "
+    "-- --spark-runner {spark_runner} --spark-master {spark_master} {spark_extra} "
```
jafors, line 25: What does the solitary `--` stand for here? Is it needed before the spark parameters?

jafors (commented):
Thanks again! I have one question below.

bio/gatk/printreadsspark/wrapper.py
fgvieira: Yes, it is the way to separate between PrintReads and spark parameters, just like in:

https://github.com/snakemake/snakemake-wrappers/blob/07a6d399384dcf58465825a99b1aa5d79f434769/bio/gatk/baserecalibratorspark/wrapper.py#L23-L29

bio/gatk/printreadsspark/environment.yaml
```diff
@@ -0,0 +1,8 @@
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
+dependencies:
+  - gatk4 ==4.2.0.0
+  - openjdk =8
+  - snakemake-wrapper-utils ==0.2.0
```
dlaehnemann, line 8: Do these need to be pinned down to these versions, or would newer patch versions also be fine to include (to get bug fixes in). I.e. something like:

```suggestion
  - gatk4 =4.2
  - openjdk =8
  - snakemake-wrapper-utils =0.2
```

Compare to the conda version matching specs:
https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#build-version-spec

bio/gatk/printreadsspark/meta.yaml
```diff
@@ -0,0 +1,18 @@
+name: gatk PrintReadsSpark
+description: |
+  Run gatk PrintReadsSpark.
```
dlaehnemann, line 3: Maybe make this describe what PrintReadsSpark actually does?
```suggestion
  Write reads from SAM format file (SAM/BAM/CRAM) that pass specified criteria to a new file. This is the version that can be run on Spark.
```

dlaehnemann (approved):
Thanks a bunch!

Two minor things. Feel free to submit those suggestions or not.

PR #356, opened 2021-05-09
Fastp: Added support for unpaired and merged reads output

fgvieira:


bio/fastp/wrapper.py
```diff
@@ -9,28 +9,62 @@
...
+# Input files
 if n == 1:
     reads = "--in1 {}".format(snakemake.input.sample)
 else:
     reads = "--in1 {} --in2 {}".format(*snakemake.input.sample)
 
+
+# Output files
 trimmed_paths = snakemake.output.get("trimmed", None)
 if trimmed_paths is not None:
```
jafors, line 29:
```suggestion
if trimmed_paths:
```

bio/fastp/wrapper.py
```diff
@@ -9,28 +9,62 @@
...
+            unpaired2 = snakemake.output.get("unpaired2", None)
+            if unpaired2:
+                trimmed += f" --unpaired2 {unpaired2}"
+
+        # Output merged PE reads
+        merged = snakemake.output.get("merged", None)
+        if merged:
+            if "--merge " not in extra:
+                raise ValueError("output.merged specified but merge option missing")
+            trimmed += f" --merged_out {merged}"
```
jafors, line 52: I like how you catch all these special cases, but with `unpaired`, `unpaired1`, `unpaired2`, `merged` and later `failed`, that's a lot of non-essential output fields which can be overwhelming for some users.
Maybe give some examples in the Snakefile (can be as comments).

jafors (changes requested):
Thanks! I have some comments below.

bio/fastp/wrapper.py
fgvieira: I've added some examples to the `Snakefile`.

I am just not sure whether the check below should be performed by the wrapper since, if `--merge` is not provided, `output.merged` will not be created and the rule will fail anyway (just with the more general "missing output" error message).

https://github.com/snakemake/snakemake-wrappers/blob/d88aa2afe48bd236e94f60ada50339db2b8f8f33/bio/fastp/wrapper.py#L51-L54

jafors: Hmm. I like this type of error message - it's probably not 100% needed, but since it is here, it makes it easier to catch such small errors.
One could also set `--merge` if the output is defined, but that would also be based on some kind of assumption that the user did not make a mistake.

jafors (approved):
Thanks for the updates!

PR #359, opened 2021-05-13
Added CollectRnaSeqMetrics wrapper.

brcopeland:


bio/picard/collectrnaseqmetrics/wrapper.py
```diff
@@ -0,0 +1,23 @@
...
+__copyright__ = "Copyright 2021, Brett Copeland"
+__email__ = "brcopeland@ucsd.edu"
+__license__ = "MIT"
+
+
+from snakemake.shell import shell
+from snakemake_wrapper_utils.java import get_java_opts
+
+strand = snakemake.params.get("strand", "NONE")
+extra = snakemake.params.get("extra", "")
```
jafors, line 11: If you want to treat `STRAND` as a specific parameter, please add an example with description in the example Snakefile. It is easier for the user to understand where to set this info without having to look into the wrapper script!
It might also be nice to always add the `extra` field to `params`, maybe even with some example flags.

bio/picard/collectrnaseqmetrics/test/Snakefile
```diff
@@ -0,0 +1,16 @@
+rule alignment_summary:
+    input:
+        bam = "mapped/a.bam",
+        refflat = "annotation.refFlat"
```
jafors, line 4:
```suggestion
        bam="mapped/a.bam",
        refflat="annotation.refFlat"
```
As mentioned in other wrappers, `snakefmt` is your friend regarding these small cosmetic issues. ;)

jafors (changes requested):
Thanks a bunch! Small comments below.

brcopeland:
The failed test doesn't appear to be related to be related to this wrapper (some conda issue?).  I'm not clear on how to resolve that.

jafors (approved):

PR #360, opened 2021-05-13
Added qualimap rnaseq wrapper.

brcopeland:


brcopeland:
Just to clarify the change made, the `qualimap` wrapper that runs the java program has a parameter, `--memory-size`, that like you'd expect passes this information onto java.  On the other hand, it does not have any option to modify the java temp directory except through putting it in the environment variable `JAVA_OPTS`, and hence the reason for utilizing that while ignoring `--memory-size`.

bio/qualimap/rnaseq/test/Snakefile
```diff
@@ -0,0 +1,14 @@
+rule qualimap:
+    input:
+        bam = "mapped/a.bam",
+        gtf = "annotation.gtf"
```
jafors, line 4:
```suggestion
        bam="mapped/a.bam",
        gtf="annotation.gtf"
```
If you format your Snakefiles with `snakefmt`, those things should be easy to catch. ;)

jafors: Also, it is always good to add some description for the `input`/`output`/`params` sections in the form of comments in the example Snakefile, even if they seem self-explanatory.

jafors (changes requested):

bio/qualimap/rnaseq/test/Snakefile
brcopeland: I've done that now.  I haven't heard of `snakefmt` previously, I'll take a look.

jafors (approved):
Thanks!

PR #363, opened 2021-05-21
Update wrapper.py for cutadapt/pe.

brcopeland:
The current version raises an exception when in line 20, neither extra nor adapters is specified.  Based on this logic, it should be the case that if either is specified the command should work, but if params.adapters is not specified, the shell command fails (because params.adapters does not exist in line 25).


jafors (approved):
Thanks for the catch. Makes sense to me.

PR #377, opened 2021-06-11
Liftoff

tdido:
### Description

Adding wrapper for liftoff (https://github.com/agshumate/Liftoff/tree/master/liftoff)

### QC

* [ x ] there is a test case which covers any introduced changes,
* [ x ] `input:` and `output:` file paths in the resulting rule can be changed arbitrarily,
* [ x ] all `environment.yaml` specifications follow [the respective best practices](https://stackoverflow.com/a/64594513/2352071),
* [ x ] wherever possible, command line arguments are inferred and set automatically (e.g. based on file extensions in `input:` or `output:`),
* [ x ] all fields of the example rules in the `Snakefile`s and their entries are explained via comments (`input:`/`output:`/`params:` etc.),
* [ x ] `stderr` and/or `stdout` are logged correctly (`log:`), depending on the wrapped tool,
* [ x ] the `meta.yaml` contains a link to the documentation of the respective tool or command,
* [ x ] `Snakefile`s pass the linting (`snakemake --lint`),
* [ x ] `Snakefile`s are formatted with [snakefmt](https://github.com/snakemake/snakefmt),
* [ x ] Python wrapper scripts are formatted with [black](https://black.readthedocs.io).


bio/liftoff/test/Snakefile
```diff
@@ -0,0 +1,18 @@
+rule liftoff:
+    input:
+        ref = "{ref}.fasta.gz",
+        tgt = "{ref}.fasta.gz",
```
jafors, line 4: This seems to be an unlucky choice for this example Snakefile, as both reference and target fasta are identical here (`{ref}` evaluates to the same wildcard). If you want to use wildcards, use `{ref}` and `{target}` or something like this.

Also, please format the Snakefile using `snakefmt` to get rid of the spaces around the `=` etc.

bio/liftoff/environment.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
+  - bioconda
+  - conda-forge
+dependencies:
+  - liftoff == 1.6.*
```
jafors, line 5:
```suggestion
  - liftoff =1.6
```

jafors (changes requested):
Thanks for this addition! Just a few comments.

bio/liftoff/environment.yaml
tdido: Thanks for the review! Changes applied.

jafors (approved):
Thanks for the changes! Looks good.

PR #386, opened 2021-06-30
Add metaspades assembler

SilasK:
### Description

Add metaspades assembler

### QC


SilasK:
@alienzj @asl
For your information. I create this snakemake wrapper for metaspades. Maybe the other spades later.

bio/spades/metaspades/wrapper.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+else:
+    # params.txt file exitst already I restart from previous run
+
+    shell("echo '\n\nRestart Spades \n' >> {log[0]}")
+
+    shell(
+        "spades.py --meta "
+        " --threads {threads} "
```
asl, line 71: You need --continue here

alienzj: Can we specify `--continue` and `--threads`, `--memory` at the same time?

asl: No, `--continue` uses the same amount of memory / CPU threads. If you need to change this, then you'd need to use `--restart-from last`.

bio/spades/metaspades/meta.yaml
```diff
@@ -1,5 +1,10 @@
...
-Input of metaspades should be at least one paired-end library (=2 fastq files) optionally merged reads as a third fastq file might be supplied and singleton reads as a 4th input file. Long reads can also be input as `pacbio` or `nanopore` input argument. To distinguish short from long reads. Use the `reads` as name for the short reads. 
+description: |
+  Assemble metagenome with metaspades. For more information see the `Spades documentation <https://cab.spbu.ru/software/spades/>`_.
+
+  Metagenome assembly uses a lot of computational resources. Spades is told to restart from a previous checkpont if the file `params.txt` exist in the output directory. In this way one can use snakemake with `--restart-times` to automatically restart the assembly. I read somewhere that spades doesn't scale very well to many cores therefore I suggest to keep the threads around 10.
```
SilasK, line 5: > I read somewhere that spades doesn't scale very well to many cores therefore I suggest to keep the threads around 10.
@asl Is this still true?

SilasK:
@johanneskoester @fgvieira Could you please review my PR?

bio/spades/metaspades/environment.yaml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
+dependencies:
+  - spades ==3.15
```
asl, line 6: Not sure if this is a correct specification. As it prohibits bug-fix versions, say, 3.15.2 or 3.15.3 that will be released soon

bio/spades/metaspades/meta.yaml
asl: No, it is not

bio/spades/metaspades/wrapper.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # parse long reads
+    for longread_name in ["pacbio", "nanopore"]:
+        if hasattr(snakemake.input, longread_name):
+            input += " --{name} {}".format(name=longread_name, **snakemake.input)
+
+    shell(
+        "spades.py --meta "
+        " --threads {threads} "
+        " --memory {resources.mem_mb}000 "
```
asl, line 69: memory is in gigabytes, so you need to divide

bio/spades/metaspades/wrapper.py
```diff
@@ -0,0 +1,104 @@
...
+else:
+    # params.txt file exitst already I restart from previous run
+
+    shell("echo '\n\nRestart Spades \n' >> {log[0]}")
+
+    shell(
+        "spades.py --meta "
+        " --restart-from last "
+        " --threads {threads} "
+        " --memory {resources.mem_mb}000 "
```
asl, line 87: ditto

bio/spades/metaspades/environment.yaml
SilasK: It is valuable for both
https://stackoverflow.com/questions/64594146/best-practices-for-bioconda-versions-in-snakemake-wrappers/64594513#64594513

asl: In fact, I'm slightly disagreeing with what is written. The issue is that all these pipelines and wrappers are rarely got updated. As a result the users still use old outdated versions of the tools (we're seeing 3-4-5-old versions with known issues still in use and users are unable to use newer version simply because the pipeline has the version hardcoded somewhere). Certainly, pinning of the version ensures that the wrapper will always work, so, it's all about pros and cons here.

From the SPAdes side we work hard on ensuring backward compatibility and given the extremely simplicity of the wrapper, may I suggest that the version specifier to be >= 3.15? :)

bio/spades/metaspades/wrapper.py
```diff
@@ -0,0 +1,110 @@
...
+__email__ = "silas.kieser@gmail.com"
+__license__ = "MIT"
+
+import os, shutil
+from snakemake.shell import shell
+
+
+# infer output directory
+
+if hasattr(snakemake.params, "output_dir"):
```
asl, line 14: Please, please save the assembly graph as well :)

bio/spades/metaspades/environment.yaml
SilasK: I adapted the version. 
I agree spades is very good compatible between the versions

alienzj (approved):

bio/spades/metaspades/test/Snakefile
```diff
@@ -0,0 +1,34 @@
+container: "docker://continuumio/miniconda3:4.4.10"
+
+
+rule run_metaspades:
+    input:
+        reads=["test_reads/sample1_R1.fastq.gz", "test_reads/ample1_R2.fastq.gz"],
```
jafors, line 6:
```suggestion
        reads=["test_reads/sample1_R1.fastq.gz", "test_reads/sample1_R2.fastq.gz"],
```
Typo!

bio/spades/metaspades/test/Snakefile
```diff
@@ -0,0 +1,34 @@
...
+        k="auto",
+        extra="--only-assembler",
+    log:
+        "log/spades.log",  #stdout/stderr is written in first log file
+    threads: 8
+    resources:
+        mem_mem=250000,
+        time=60 * 24,
+    wrapper:
+        "master/bio/bwa/aln"
```
jafors, line 24: This is likely a relict from the wrapper you used as a template. If you add a testcase, this should throw an error.

bio/spades/metaspades/test/Snakefile
```diff
@@ -0,0 +1,34 @@
...
+
+
+rule download_reads:
+    output:
+        ["test_reads/sample1_R1.fastq.gz", "test_reads/ample1_R2.fastq.gz"],
+    log:
+        "log/download.log",
+    shell:
+        " wget https://zenodo.org/record/3992790/files/test_reads.tar.gz >> {log} 2>&1"
+        " tar -xzf test_reads.tar.gz >> {log} 2>&1"
```
jafors, line 34: You are missing an `&&` between the `wget` and `tar`. Generally, it's not a bad idea to download the test data instead of adding it to the repo directly, but make sure that this is called correctly during your testcase.

bio/spades/metaspades/wrapper.py
```diff
@@ -0,0 +1,122 @@
...
+        output_file = snakemake.output.scaffolds
+    else:
+        output_file = snakemake.output[0]
+
+    output_dir = os.path.split(output_file)[0]
+
+
+# parse params
+extra = snakemake.params.get("extra", "")
+kmers = snakemake.params.get("k", "'auto'")
```
jafors, line 31: There is a default value for `-k` in spades, right? Why not use this value if `k` is not given in the params?

bio/spades/metaspades/test/Snakefile
jafors: Also, the same typo as above.

jafors (changes requested):
Hi and thanks for the addition. Some nice ideas! Yet you are lacking a testcase in the `test.py` and there are some other things that would have probably arisen in a test run.

bio/spades/metaspades/wrapper.py
SilasK: 'auto' is the default in spades, as far as I know.

SilasK:
Maybe it would be worth splitting up the test.py in per-repo test scripts and then run them all. 

Thank you for your comments @jafors. I made the modifications.

bio/spades/metaspades/wrapper.py
jafors: Ah, ok. I am not really familiar with the tool, probably read that wrong on the docs page.

SilasK:
@jafors Could you approve the review? and run the tests. From my side, I cannot proceed. 

Is there a way to test the wrapper locally?

jafors:
@SilasK , you can test locally following those instructions: https://snakemake-wrappers.readthedocs.io/en/stable/contributing.html#testing-locally

SilasK:
@jafors I successfully tested locally. GitHub tells me that you requested changes that are not yet approved. It links to the https://github.com/snakemake/snakemake-wrappers/pull/386#pullrequestreview-721092587

Could you approve this PR so it could be merged?

jafors (approved):

PR #390, opened 2021-07-19
Add GATK ValidateVariants

G-kodes:
### Description

An implementation of the GATK ValidateVariants tool

### QC

> Lord have mercy on my code. PR's are stressful lol


bio/gatk/validatevariants/wrapper.py
```diff
@@ -0,0 +1,27 @@
...
+from snakemake.shell import shell
+from snakemake_wrapper_utils.java import get_java_opts
+
+extra = snakemake.params.get("extra", "")
+java_opts = get_java_opts(snakemake)
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+extra = snakemake.params.get("extra", "")
+preloader = snakemake.params.get("preloader", "")
```
johanneskoester, line 15: What is this preloader meant for? Is it intended for setting some shell variables or loading env modules? Snakemake has built in mechanism for that (`shell.prefix` and the envmodule directive).

johanneskoester (changes requested):

bio/gatk/validatevariants/wrapper.py
G-kodes: NOted @johanneskoester I did not know that was a feature. The preloader is because all of my applications require the loading of modules as I work through a PBS/Torque batch schedule system. I will make the required changes.

G-kodes: @johanneskoester can you please maybe point me in the right direction for the proper usage of `prefix`. I have found an [FAQ](https://snakemake.readthedocs.io/en/stable/project_info/faq.html?highlight=shell.prefix#i-want-to-configure-the-behavior-of-my-shell-for-all-rules-how-can-that-be-achieved-with-snakemake) indicating it can be globally applied and found the [internal API reference](https://snakemake.readthedocs.io/en/stable/api_reference/internal/snakemake.html?highlight=envmodule#snakemake.shell.shell.prefix), however, the desired usage would be to apply this on a per-rule or per-shell basis. Do you know if this is currently supported?

bio/gatk/validatevariants/test/Snakefile
```diff
@@ -0,0 +1,13 @@
+rule vcf_spec_validation:
+    input:
+        vcf="sample.vcf",
+    output:
+        "results/sample_VALID.vcf",
+    log:
+        "results/sample_VALID.log",
+    params:
+        R="genome.fasta",
```
johanneskoester, line 9: I cannot find this to be used in the wrapper. On the other hand, extra is missing. Maybe you meant to write

```suggestion
        extra="", # optional extra arguments
```

bio/gatk/validatevariants/test/Snakefile
```diff
@@ -0,0 +1,13 @@
+rule vcf_spec_validation:
+    input:
+        vcf="sample.vcf",
+    output:
+        "results/sample_VALID.vcf",
```
johanneskoester, line 5: I think the output should just have a .txt suffix or so, since this tool will not create a vcf file, right?

johanneskoester (changes requested):

bio/gatk/validatevariants/test/Snakefile
G-kodes: It does create a vcf file. Only if it passes all tests.

fgvieira:
Maybe it would also be nice to set `tmpdir` with `tempfile`.

PR #394, opened 2021-07-26
add 'extra' in testing snakemake file

gbinux:
The wrapper uses named parameter 'extra' to get value from params but it is missing in the testing snakefile

### QC

* [ ] `Snakefile`s are formatted with [snakefmt](https://github.com/snakemake/snakefmt),
* [ ] Python wrapper scripts are formatted with [black](https://black.readthedocs.io).

### DIFF
--- a/bio/picard/markduplicates/test/Snakefile
+++ b/bio/picard/markduplicates/test/Snakefile
@@ -10,7 +10,7 @@ rule mark_duplicates:
     log:
         "logs/picard/dedup/{sample}.log"
     params:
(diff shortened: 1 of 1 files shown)


gbinux:
Also, when `--use-conda`, wrappers related to `picard` tend to have fatal errors as mentioned here: https://github.com/broadinstitute/picard/issues/1329

A simple fix is to add set `extra` to `USE_JDK_DEFLATER=true USE_JDK_INFLATER=true`

jafors (approved):
Thanks for the catch!

PR #397, opened 2021-08-10
Cutadapt/se: fixed bug in wrapper.py

litwinowicz:
Line 20 (and practical use of cutadapt) suggests, that defining only adapters or extra in params should work. However currently omitting either results in AttributeError.

### Description

Proposed change allows for defining only adapters or only extra, which much better suits practical use of cutadapt.
### QC


jafors (approved):
Good catch, thanks for this!

PR #401, opened 2021-09-09
Added wrapper for verifybamid2.

brcopeland:
### Description
This provides a wrapper for the tool `verifybamid2` which can be used to estimate intra-species contamination in a BAM file.

### QC


bio/verifybamid/verifybamid2/test/Snakefile
```diff
@@ -0,0 +1,17 @@
...
+        bam="a.bam",
+        ref="genome.fasta",
+    output:
+        selfsm="a.selfSM",
+        ancestry="a.ancestry",
+    params:
+        # optional - this can be used to specify custom resource files if
+        # necessary (if using GRCh37 or GRCh38 instead simply specify
+        # genome_build="38", for example
+        svd_prefix="ref.vcf",
```
johanneskoester, line 12: This should be an optional input file then, not a parameter.

johanneskoester (changes requested):

johanneskoester:
Thanks a lot, nice work! Just one comment above.

bio/verifybamid/verifybamid2/test/Snakefile
brcopeland: So the `svd_prefix` that is passed in to `verifybamid2` is not actually a file that exists, so that's why I didn't make it an input file.  I could change it to have as an input one of the files that do exist, strip off the file suffix, and pass that to `verifybamid2` if you would prefer.

johanneskoester: Mhm, would it be possible to specify all needed files under that prefix explicitly? E.g. using [multiext](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html?highlight=multiext#the-multiext-function)?

brcopeland: I changed `params.svd_prefix` to `input.svd_mu`.  This file name would be the same as the other with the suffix `.mu` appended.  `verifybamid2` requires other files with suffixes `.bed`, `.UD`, and `.V` to exist so the wrapper checks that.

Please note that the wrapper looks explicitly for `input.svd_mu` and not for the others as far as specified arguments go.  I just thought that approach would be simpler to work with.

johanneskoester (approved):

PR #411, opened 2021-10-22
Changed handling of temp directory to snakemake resource

fgvieira:
### Description

Use `snakemake` `tmpdir` resource, instead of defining `--tmp-dir` in the rule

### QC

### DIFF
--- a/bio/gatk/genotypegvcfs/test/Snakefile
+++ b/bio/gatk/genotypegvcfs/test/Snakefile
@@ -11,10 +11,6 @@ rule genotype_gvcfs:
     params:
         extra="",  # optional
         java_opts="", # optional
-    # optional specification of memory usage of the JVM that snakemake will respect with global
-    # resource restrictions (https://snakemake.readthedocs.io/en/latest/snakefiles/rules.html#resources)
-    # and which can be used to request RAM during cluster job submission as `{resources.mem_mb}`:
-    # https://snakemake.readthedocs.io/en/latest/executing/cluster.html#job-properties
     resources:
         mem_mb=1024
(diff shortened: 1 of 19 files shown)


bio/gatk/genotypegvcfs/test/Snakefile
```diff
@@ -11,11 +11,8 @@ rule genotype_gvcfs:
...
         extra="",  # optional
         java_opts="", # optional
-    # optional specification of memory usage of the JVM that snakemake will respect with global
-    # resource restrictions (https://snakemake.readthedocs.io/en/latest/snakefiles/rules.html#resources)
-    # and which can be used to request RAM during cluster job submission as `{resources.mem_mb}`:
-    # https://snakemake.readthedocs.io/en/latest/executing/cluster.html#job-properties
     resources:
-        mem_mb=1024
+        mem_mb=1024,
+        tmpdir="tmp/"
```
johanneskoester, line 16: No need to specify this here directly, this would just encourage anti-patterns. Rather, this should be handled by --default-resources, see [here](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html?highlight=tmpdir#standard-resources).

johanneskoester (changes requested):

bio/gatk/genotypegvcfs/test/Snakefile
fgvieira: Not sure what you mean by 
Ok, so I'll remove the explicit definition of `tmpdir` from here.

johanneskoester (approved):

PR #414, opened 2021-10-28
docs: instructions on limits setting for star wrapper

alienzj:
### Description

more friendly STAR
Fix this issue: https://github.com/alexdobin/STAR/issues/1344

### QC


bio/star/align/wrapper.py
```diff
@@ -38,6 +38,8 @@
 
 outprefix = os.path.dirname(snakemake.output[0]) + "/"
 
+shell("ulimit -n 10000")
```
johanneskoester, line 41: I am not sure whether this will always be allowed. There might be clusters where admins do not allow this. Hence, I'd rather let users or admins configure such things before a pipeline runs.

johanneskoester:
So, instead of setting the ulimit, I suggest to just link to the mentioned issue from the notes in the meta.yaml, mentioning that users should ensure their ulimits are configured according to the requirements when using the star wrapper. Maybe you would like to change this PR accordingly?

bio/star/align/wrapper.py
alienzj: Yes, if the user does not have permission to run this, the process runs an error. Setting it in advance is indeed a better way to deal with it.

alienzj:
OK, I will update meta.yaml.

PR #415, opened 2021-11-04
feat: wrapper for Manta SV caller

fgvieira:
### QC


fgvieira:
Not sure how to deal with the temp files, since `manta` creates a folder with everything (temp files included).
We could set the workdir to a temp folder (with python `tempfile`) and move the result files to their final destination at the end, but this would make it impossible to access all the logs and intermediate files `manta` creates (in case the user wants them).

bio/manta/test/Snakefile
```diff
@@ -0,0 +1,22 @@
+rule manta:
+    input:
+        ref="human_g1k_v37_decoy.small.fasta",
+        samples=["mapped/a.bam"],
+    output:
+        outdir=temp(directory("temp/")),
```
johanneskoester, line 6: Maybe leave a comment here explaining what this directory will contain.

johanneskoester (changes requested):
Nice work, just one thing.

bio/manta/test/Snakefile
fgvieira: But should we leave that directory for the user to define (and set as temp), or should we internally define it as `tmpdir` with python?

johanneskoester: Depends on whether you think the directory will contain useful output. The wrapper could alternatively also copy out certain files of it upon definition of them as log files or output files. That would probably be the nicer way.

fgvieira: Ok, in the end I think it is cleaner to just use `tempfile` and copy out important files.
If anyone needs any of the other files, we can just change the wrapper to also copy those files.

bio/manta/wrapper.py
```diff
@@ -0,0 +1,117 @@
...
+        shell(
+            "bcftools view --threads {snakemake.threads} --output {cand_sv_vcf:q} --output-type {cand_sv_vcf_format} {cand_sv_vcf_path:q} {log}"
+        )
+
+        cand_sv_idx = snakemake.output.get("cand_sv_idx", "")
+        cand_sv_idx_path = results_base / "candidateSV.vcf.gz.tbi"
+        if cand_sv_idx and cand_sv_idx != cand_sv_idx_path:
+            shell(
+                "bcftools index --threads {snakemake.threads} --output {cand_sv_idx:q} {cand_sv_vcf:q} {log}"
+            )
```
johanneskoester, line 117: This looks quite redundant to me. Can you turn it into a function that is just called 3 times?

johanneskoester (changes requested):

bio/manta/wrapper.py
fgvieira: You're right... it can definitely be made cleaner. :)

bio/manta/wrapper.py
```diff
@@ -0,0 +1,100 @@
...
+                shell(
+                    "bcftools index --threads {snakemake.threads} --output {dest_idx:q} {dest_vcf:q} {log}"
+                )
+
+    results_base = run_dir / "results" / "variants"
+
+    # Copy main VCF output
+    vcf_temp = results_base / "diploidSV.vcf.gz"
+    vcf_final = snakemake.output.get("vcf", "")
+    idx_final = snakemake.output.get("idx", "")
```
johanneskoester, line 87: No need to add "" as default case here and below. If you omit that, it will return None, which should be fine for the function as well.

johanneskoester (changes requested):

bio/manta/wrapper.py
fgvieira: Sure.

johanneskoester (approved):

Issue #420, opened 2021-11-13
Shared memory feature with Snakemake STAR wrapper


RuiyuRayWang:
This is a duplicate issue of
https://github.com/snakemake/snakemake/issues/1259#issue-1052036783

I'm writing a pipeline using STAR aligner. The task involves sequentially aligning fastq files to the same genome index. To speed up the alignment, I figured that the "shared memory" feature can be used to avoid repeatedly loading and un-loading the genome index.

I have tried and successfully achieved this with STAR pre-built on my machine or STAR built by the integrated package management in snakemake.
I reallly want to use the Snakemake wrapper to do this as it provides better transportability. However, the "STAR align" wrapper requires "fq1" as a minimum input which prevents the "un-load" step of STAR genome index. 

The shell command for STAR to unload genome doesn't have any input or output, which is why the current wrapper cannot be used:

`STAR --genomeLoad Remove --genomeDir {input.index}`

Would it be possible to modify the existing code to make this feature possible?

It may be even better to add a new wrapper sub-commands under star (i.e. `STAR GENOMELOAD`) that incorporates both `--genomeLoad LoadAndExit` and `--genomeLoad Remove` options.

RuiyuRayWang:
Created a pull request for this:
https://github.com/snakemake/snakemake-wrappers/pull/421#issue-1052559722

First time contributing to the repo, correct me if I made any mistakes. Thanks!

RuiyuRayWang:
Closing this issue as Snakemake `service` support has been explicitly developed to address similar jobs at backend. Thanks @johanneskoester!

Check out the relevant links:
[https://github.com/snakemake/snakemake-wrappers/pull/421](url)
[https://github.com/snakemake/snakemake/pull/1413](url)

TKMarkCheng:
Having the same struggle as @moxgreen here. @RuiyuRayWang please can you share your STAR rules for how did you load your genome? I find it awkward since `STAR --genomeLoad` doesn't take in input or outputs, and STAR index takes place in the same workflow which already occupies the genome directory in its output.

RuiyuRayWang:
@TKMarkCheng I have posted a workaround example here: [https://stackoverflow.com/questions/69894230/using-star-shared-memory-module-in-snakemake-for-sequential-alignment-tasks](url). Let me know if you have further questions.

fgvieira:
Thanks for posting it here also!

PR #421, opened 2021-11-13
STAR shared memory to load/unload genome for faster job chaining

RuiyuRayWang:
### Description
Relate to this issue:
https://github.com/snakemake/snakemake-wrappers/issues/420#issue-1052530618
STAR alignment jobs often runs over the same genome index for repeated times. To save time from repeated loading and unloading the genome index in individual runs, created a STAR wrapper sub-module "STAR genome" which explicitly take care of loading/unloading of genome index, using "shared memory" module from STAR.

### QC


johanneskoester (commented):
Thanks a lot for the proposal! While I like this a lot, I believe we need dedicated service job support in Snakemake first, in order to make it portable across backends. Your current proposal here would only work locally. I have taken this as an incentive to finally implement such support in Snakemake: https://github.com/snakemake/snakemake/pull/1413

RuiyuRayWang:
Thanks so much for the feedback! @johanneskoester 
I'll close this PR and update my pipeline with the Snakemake service support :-)

moxgreen:
Dear @RuiyuRayWang would you like to share the code to show how you managed to use services to deal with the shared momory in STAR?

PR #423, opened 2021-11-22
Add java_opts

FelixMoelder:
### Description

As `fgbio AnnotateBamWithUmis` is very memory consuming and tends to fail with an OutOfMemory-Error an optional `java_opts`-parameter has been added.
Also fgbio has been updated to its latest release.

### QC

### DIFF
--- a/bio/fgbio/annotatebamwithumis/test/Snakefile
+++ b/bio/fgbio/annotatebamwithumis/test/Snakefile
@@ -6,7 +6,7 @@ rule AnnotateBam:
         "mapped/{sample}.annotated.bam",
     params: ""
     resources:
-        mem_gb="4" # memory to be propagated to fgbio
+        mem_gb="4" # memory to be given to fgbio
     log:
         "logs/fgbio/annotate_bam/{sample}.log",
     wrapper:
--- a/bio/fgbio/annotatebamwithumis/test/Snakefile
+++ b/bio/fgbio/annotatebamwithumis/test/Snakefile
@@ -1,11 +1,13 @@
 rule AnnotateBam:
     input:
         bam="mapped/{sample}.bam",
-        umi="umi/{sample}.fastq"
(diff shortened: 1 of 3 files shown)


bio/fgbio/annotatebamwithumis/test/Snakefile
```diff
@@ -1,11 +1,13 @@
...
         bam="mapped/{sample}.bam",
-        umi="umi/{sample}.fastq"
+        umi="umi/{sample}.fastq",
     output:
-        "mapped/{sample}.annotated.bam"
-    params: ""
+        "mapped/{sample}.annotated.bam",
+    params:
+        extra="",  #optional
+        java_opts="",  #optional
```
johanneskoester, line 9: Rather propose to use resources with a mem_mb statement, and add a comment that this will be propagated to fgbio (see [here](https://github.com/snakemake/snakemake-wrapper-utils/blob/master/snakemake_wrapper_utils/java.py)).

johanneskoester (changes requested):

bio/fgbio/annotatebamwithumis/test/Snakefile
```diff
@@ -1,11 +1,13 @@
...
     input:
         bam="mapped/{sample}.bam",
-        umi="umi/{sample}.fastq"
+        umi="umi/{sample}.fastq",
     output:
-        "mapped/{sample}.annotated.bam"
+        "mapped/{sample}.annotated.bam",
     params: ""
+    resources:
+        mem_gb="4" # memory to be propagated to fgbio
```
johanneskoester, line 9:
```suggestion
        mem_gb="4" # memory to be given to fgbio
```

johanneskoester (changes requested):

johanneskoester (approved):

PR #448, opened 2022-01-27
feat: explicitly specify bwa index in bwa samxe

fgvieira:
### Description

Following #232, doing the same for the `smaxe` wrapper. 

### QC

### DIFF
--- a/bio/bwa/samxe/test/Snakefile_samtools
+++ b/bio/bwa/samxe/test/Snakefile_samtools
@@ -2,7 +2,6 @@ rule bwa_sam_pe:
     input:
         fastq=["reads/{sample}.1.fastq", "reads/{sample}.2.fastq"],
         sai=["sai/{sample}.1.sai", "sai/{sample}.2.sai"],
-        # Index can be a list of (all) files created by bwa, or one of them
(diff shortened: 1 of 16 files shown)


fgvieira:
@johanneskoester this one has the same problem as #446, so it seems it is an issue with `snakemake`. Can you take a look, when you have time?

thanks,

bio/bwa/samxe/test/Snakefile_samtools
```diff
@@ -56,10 +59,11 @@ rule bwa_bam_se:
     input:
         fastq="reads/{sample}.1.fastq",
         sai="sai/{sample}.1.sai",
+        # Index can be a list of (all) files created by bwa, or one of them
```
johanneskoester, line 62: While one of them would suffice in the local case, you need all in case there is no shared file system (i.e. in the cloud). Hence, it is best practice to have all of them mentioned here.

johanneskoester (changes requested):

bio/bwa/samxe/test/Snakefile_samtools
fgvieira: It is just that a similar comment was added on the other `bwa` wrappers recently. But just removed it and also in the other wrappers.

johanneskoester (approved):

PR #460, opened 2022-03-01
fix: update multiqc version to 1.12

vinisalazar:
Hi,

I've been having constant problems with this wrapper. My Snakemake execution almost always hangs in the "Solving environment step". I was supportive of merging #422 but considering MultiQC is on version 1.12, I believe it would be better to drop this pinned version. I tried creating the environment from the file as it is in this PR and it correctly installs v1.12 of MultiQC.

Ping @fgvieira @thomasbtf, would you agree?

Thank you for any assistance you can provide.

Cheers,
Vini

Summary of changes:

  - Remove '= 1.11' from multiqc environment yaml

MultiQC has been updated to 1.12 and pinning v1.11 is hampering environment solving.

@thomasbtf @fgvieira would you be okay with this?

### DIFF
--- a/bio/biobambam2/bamsormadup/wrapper.py
+++ b/bio/biobambam2/bamsormadup/wrapper.py
@@ -34,7 +34,7 @@ if metrics:
 with tempfile.TemporaryDirectory() as tmpdir:
     # This folder must not exist; it is created by BamSorMaDup
     tmpdir_bamsormadup = Path(tmpdir) / "bamsormadup_{:06d}".format(
-        random.randrange(10 ** 6)
+        random.randrange(10**6)
     )
 
     shell(
--- a/bio/gatk/applybqsrspark/wrapper.py
+++ b/bio/gatk/applybqsrspark/wrapper.py
@@ -21,7 +21,7 @@ log = snakemake.log_fmt_shell(stdout=True, stderr=True)
 
 with tempfile.TemporaryDirectory() as tmpdir:
     # This folder must not exist; it is created by GATK
-    tmpdir_shards = Path(tmpdir) / "shards_{:06d}".format(random.randrange(10 ** 6))
+    tmpdir_shards = Path(tmpdir) / "shards_{:06d}".format(random.randrange(10**6))
 
     shell(
         "gatk --java-options '{java_opts}' ApplyBQSRSpark"
--- a/bio/multiqc/environment.yaml
+++ b/bio/multiqc/environment.yaml
@@ -3,4 +3,4 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - multiqc =1.11
+  - multiqc =1.12


fgvieira:
I agree about updating `multiqc` version, but not sure about completely unpinning its version. Why not just update to `1.12`?

vinisalazar:
Hi,

@fgvieira thank you for the reply. I assumed that most wrappers did not specify a version, instead letting conda solve for the latest version available in the channel. Upon a second look I now see that most if not all wrappers set a specific version. I will update the PR accordingly.

However, this did get me thinking that all wrappers must be manually updated every time a new version is out? This seems very laborious. Is there a reason to avoid not pinning a version (i.e. attempting to solve for the latest one)? If there is, I believe this repository would benefit from some sort of automation, such as a GitHub Action or Bot, to update the version based on the Bioconda repository? Or is manual updating a deliberate practice? 

Apologies if I'm confused about any aspects of best practices regarding the usage of wrappers, and thank you for the support.

Cheers,
V

fgvieira:
As far as I know, all wrappers set a specific version so the results of a workflow are reproducible; this way every wrapper version is associated with a specific program version.

As far as implementing a GitHub Action or Bot, it could be a good idea (at least for most wrappers), but maybe we should have some feedback from @johanneskoester.

For this PR, can you fix the two failed tests?

vinisalazar:
Hi @fgvieira thank you for your response.

I fixed the title of the PR. The Code quality / formatting check however fails on code that is unrelated to this PR. I even tried running black on the files indicated in the test but they were left unchanged.

This is what I get when running Black:
```
$ black --check */*/*/wrapper.py */*/wrapper.py
All done! ✨ 🍰 ✨
288 files would be left unchanged.
```

Would you have any advice or would it be okay to merge as is?

Thank you,
V

thomasbtf:
It's also totally fine with me to update the multiQC version 👍 

Regarding the [code quality issue](https://github.com/snakemake/snakemake-wrappers/runs/5386871614?check_suite_focus=true#step:4:6), it seems that some other files would be changed:
```python
would reformat bio/biobambam2/bamsormadup/wrapper.py
would reformat bio/gatk/applybqsrspark/wrapper.py
```

Maybe it helps to update this PR with the current master of the snakemake-wrappers repo?

vinisalazar:
Hi @thomasbtf, thank you for your response.

I just double-checked and the branch is up-to-date with master (at d66c264).

Cheers,
V

thomasbtf:
Okay, no luck there.

Another idea: maybe your local black version doesn't match the one from the GitHub checks? On the GitHub Runners, black is used in version `22.1.0`.

vinisalazar:
@thomasbtf, thanks, that nailed it. Just pushed c9e489e that fixes the tests.

fgvieira (approved):

PR #476, opened 2022-04-19
feat: make it possible to set output type, vcf or gvcf.

Smeds:
### Description

Make it possible to set if a vcf or gvcf will be outputted from the wrapper

### QC

* [ ] either the wrapper can only use a single core, or the example rule contains a `threads: x` statement with `x` being a reasonable default,


Smeds:
conde refused to setup if I didn't lock mamba version to 0.15.3

bio/gatk/haplotypecaller/wrapper.py
```diff
@@ -27,9 +27,25 @@
...
+        raise ValueError(
+            "please set vcf or gvcf as output, not both! It's not supported by gatk"
+        )
+    else:
+        raise ValueError("please set one of vcf or gvcf as output (not both)!")
+
 bam_output = snakemake.output.get("bam", "")
 if bam_output:
-    bam_output = "--bam-output " + str(bam_output)
+    output = " --bam-output " + str(bam_output)
```
fgvieira, line 48: This is an extra output, compatible with both `vcf` anf `g.vcf` ouput.

As it is, it overwrites the previous output options

johanneskoester:
Can you revert your hotfix for the setup? I have just merged fixes to the CI config.

Smeds:
> Can you revert your hotfix for the setup? I have just merged fixes to the CI config.

should only be changes to the wrapper left

fgvieira (approved):

PR #482, opened 2022-04-29
feat: Salmon update

tdayris:
### Description

This PR brings an update of Salmon to 1.8

Decoys, pseudo-bams and gene-map files are handled, `resources`:`tempdir` is used to set temporary directory in salmon index. Dedicated errors are raised in salmon quant as well as deprecation warning.

Main behavior is changed in salmon index, causing a change break, since multiple input are possible now. Salmon quant interface is not changed and the update should be silent, as long as indexes are re-built by users.

You may notice the test dataset in salmon quant has changed: this is due to the new indexing method used after salmon 1.XXX. It required for unit-testing.

Also, manual decompression is not needed anymore.

### QC

* [X] either the wrapper can only use a single core, or the example rule contains a `threads: x` statement with `x` being a reasonable default,


bio/salmon/index/wrapper.py
```diff
@@ -6,11 +6,24 @@
...
-shell(
-    "salmon index -t {snakemake.input} -i {snakemake.output} "
-    " --threads {snakemake.threads} {extra} {log}"
-)
+decoys = ""
+if "decoys" in snakemake.input.keys():
+    decoys = "--decoys {}".format(snakemake.input["decoys"])
+
+resources_tmp = snakemake.resources.get("tmpdir", None)
+with TemporaryDirectory(dir=resources_tmp) as tempdir:
```
fgvieira, line 19: No need to specifiy `resources_tmp`, since it is automatically set through an environmental variable.
You can do just like in (e.g.):
https://github.com/snakemake/snakemake-wrappers/blob/dd8c0662a39850b62b5758f7ecf291a5337c86b8/bio/gatk/haplotypecaller/wrapper.py#L36

bio/salmon/index/wrapper.py
```diff
@@ -6,11 +6,24 @@
...
 log = snakemake.log_fmt_shell(stdout=True, stderr=True)
 extra = snakemake.params.get("extra", "")
 
-shell(
-    "salmon index -t {snakemake.input} -i {snakemake.output} "
-    " --threads {snakemake.threads} {extra} {log}"
-)
+decoys = ""
+if "decoys" in snakemake.input.keys():
+    decoys = "--decoys {}".format(snakemake.input["decoys"])
```
fgvieira, line 16: Maybe a bit cleaner:

```suggestion
decoys = snakemake.input.get("decoys", "")
if "decoys":
    decoys = f"--decoys {decoys}"
```

bio/salmon/index/test/Snakefile
```diff
@@ -1,13 +1,13 @@
 rule salmon_index:
     input:
-        "assembly/transcriptome.fasta"
+        sequences="assembly/transcriptome.fasta",
     output:
-        directory("salmon/transcriptome_index")
+        directory("salmon/transcriptome_index"),
```
fgvieira, line 5: Would it make sense to explicitely have all files created?
For eaxmple:

      multiext("salmon/transcriptome_index/", "complete_ref_lens.bin", "ctable.bin", "ctg_offsets.bin", "..."),

This way `snakemake` checks all output files created and not just the directory.

bio/salmon/index/wrapper.py
tdayris: I removed it, thank you

bio/salmon/index/wrapper.py
tdayris: f-strings used here and in salmon-quant wrapper as you suggested, thank you

bio/salmon/index/test/Snakefile
tdayris: multiext has be included, with a new test case in salmon quant.

tdayris:
Hi, I solve new errors and come back here asap

bio/salmon/quant/test/Snakefile_index_list
```diff
@@ -0,0 +1,37 @@
...
+        ),
+    output:
+        quant="salmon/{sample}/quant.sf",
+        lib="salmon/{sample}/lib_format_counts.json",
+    log:
+        "logs/salmon/{sample}.log",
+    params:
+        # optional parameters
+        libtype="A",
+        #zip_ext = bz2 # req'd for bz2 files ('bz2'); optional for gz files('gz')
```
fgvieira, line 33: This parameter does not seemt to be in use anymore.
Does salmon automatically infer input format now?

tdayris: You are right, the snakefiles and meta.yaml have been updated

fgvieira: And does the wrapper still support `bz2` and `gz` as input?

tdayris: Test cases are already gzipped. Let me add a bz2 test case.

tdayris: You were right about bz2, the last commit restores the bz2 support.

fgvieira: Shoule `gzip` and `bzip2` alse be included in `environment.yaml`? Not sure if they are included byu default...

tdayris: bzip2 is not included, as Salmon explicitly raises errors on non gz/fastq/fq files. None were not included in previous wrapper version, but let's add bzip2 1.0.8 and gzip 1.11 from conda-forge.

We are talking about `salmon/quant/environment.yaml`. Should I also add it to `salmon/index/environment.yaml` to let both subcommands have the same conda environment ?

fgvieira (approved):

bio/salmon/index/wrapper.py
```diff
@@ -5,12 +5,31 @@
...
+decoys = snakemake.input.get("decoys", "")
+if decoys:
+    decoys = f"--decoys {decoys}"
+
+print(type(snakemake.output))
+output = snakemake.output
+if len(output) > 1:
+    output = dirname(snakemake.output[0])
+
+resources_tmp = snakemake.resources.get("tmpdir", None)
```
fgvieira, line 24:
```suggestion
resources_tmp = snakemake.resources.get("tmpdir", None)
```

bio/salmon/quant/test/Snakefile_index_list
fgvieira: It doesn't seem like gzip or bzip2 are needed for `salmon/index`, right? If so , no need to add it there.

bio/salmon/index/wrapper.py
tdayris: Removed in last commit. Thanks

bio/salmon/index/wrapper.py
```diff
@@ -5,12 +5,30 @@
...
 
-shell(
-    "salmon index -t {snakemake.input} -i {snakemake.output} "
-    " --threads {snakemake.threads} {extra} {log}"
-)
+decoys = snakemake.input.get("decoys", "")
+if decoys:
+    decoys = f"--decoys {decoys}"
+
+print(type(snakemake.output))
```
fgvieira, line 19: Is this print intended?

tdayris: Of course not. I removed it. Thank you for you vigilance.

fgvieira (approved):

PR #485, opened 2022-05-06
feat: Mashmap

tdayris:
### Description

This PR adds [MashMap](https://github.com/marbl/MashMap), a long read aligner.

### QC

* [X] either the wrapper can only use a single core, or the example rule contains a `threads: x` statement with `x` being a reasonable default,


bio/mashmap/environment.yaml
```diff
@@ -0,0 +1,7 @@
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
+dependencies:
+  - mashmap ==2.0
+  - gsl ==2.7
```
fgvieira, line 7:
```suggestion
  - mashmap =2.0
  - gsl =2.7

```

bio/mashmap/meta.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - query: Path to query file (fasta, fastq)
+output:
+  - Path to the alignment file
+params:
+  - extra: Optional parameters for MashMap
+notes: |
+  `input.ref` may be either a path to a file containing sequences (fasta) or a text file containing a list
+  paths to fasta-formatted sequence files.
+  
+  `input.query` may be either a fastq file or a fasta file.
```
fgvieira, line 17:
```suggestion
  * `input.ref` may be either a path to a fasta file or a text file containing a list of paths to several fasta files.
  * `input.query` may be either a fastq file or a fasta file.
```

bio/mashmap/wrapper.py
```diff
@@ -0,0 +1,31 @@
...
+__license__ = "MIT"
+
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+extra = snakemake.params.get("extra", "")
+
+# Handling input file types (either a fasta file, or a text file with a list of paths to fasta files)
+ref = snakemake.input["ref"]
+if ref.endswith((".fa", ".fna", ".fasta")):
```
fgvieira, line 18: Does `MashMap` support gzipped reference genomes?

fgvieira:
It seems that `MashMap` has its own output format, "space-delimited with each line consisting of query name, length, 0-based start, end, strand, target name, length, start, end and mapping nucleotide identity".

Just for the sake com consistency with other read mappers, is there a way to convert this output to a SAM/BAM/CRAM?

bio/mashmap/wrapper.py
tdayris: Unfortunately not. Do you think this wrapper should support on-the-fly decompression of reference fasta files ?

tdayris:
> Just for the sake com consistency with other read mappers, is there a way to convert this output to a SAM/BAM/CRAM?

Mashmap format to SAM/BAM/CRAM is not supported by MashMap. There is an open issue about that on their git. I'm wrapping this tool to follow Salmon guideline to create a gentrome file with decoy sequences.

> Also, does MasMap uses any temp folder?

There is no official variable or command line argument to overload any default temporary directory.

bio/mashmap/meta.yaml
```diff
@@ -11,7 +11,5 @@ output:
 params:
   - extra: Optional parameters for MashMap
 notes: |
-  `input.ref` may be either a path to a file containing sequences (fasta) or a text file containing a list
-  paths to fasta-formatted sequence files.
-  
-  `input.query` may be either a fastq file or a fasta file.
\ No newline at end of file
+  * `input.ref` may be either a path to a fasta file or a text file containing a list of paths to several fasta files.
+  * `input.query` may be either a fastq file or a fasta file.
```
fgvieira, line 15: According to your latest changes, it seems this one can also be a text file with a list of FASTA/Q files, no?

bio/mashmap/wrapper.py
tdayris: Okay, I'm still on this. Using `<(gzip ...)` syntax leads to a 127 error. Command not found. But gzip is here, mashmap and gsl are here ... The error only occurs with gzipped fasta files as reference.

tdayris: Sooo ... It works here ... It does not locally.

fgvieira (approved):

PR #488, opened 2022-05-13
feat: make it possible to set input as files or dirs to multiqc

Smeds:
make it possible to use the actual input from the rule as input to multiqc, instead of converting it to dirs that are then searched.

### Description
If multiqc search directories it's possible that unwanted files are included to the reports. This will make it possible to only use the provided files, when wanted.

### DIFF
--- a/bio/multiqc/meta.yaml
+++ b/bio/multiqc/meta.yaml
@@ -1,6 +1,7 @@
 name: multiqc
 description: |
   Generate qc report using multiqc.
+url: https://multiqc.info/
 authors:
   - Julian de Ruiter
 input:
--- a/bio/multiqc/meta.yaml
+++ b/bio/multiqc/meta.yaml
@@ -4,6 +4,8 @@ description: |
 authors:
   - Julian de Ruiter
 input:
-  - input directory containing qc files
+  - input directory containing qc files, default behaviour is to extract folder path from the provided files or parent folder if a folder is provided.
+params:
+  - use_input_files_only: if this variable is set to True input will be used as it is, i.e no folder will be extract from provided file names
 output:
   - qc report (html)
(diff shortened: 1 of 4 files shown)


bio/multiqc/test/Snakefile
```diff
@@ -1,10 +1,23 @@
-rule multiqc:
+rule multiqc_dir:
     input:
         expand("samtools_stats/{sample}.txt", sample=["a", "b"])
     output:
         "qc/multiqc.html"
     params:
-        ""  # Optional: extra parameters for multiqc.
+        extrra=""  # Optional: extra parameters for multiqc.
```
fgvieira, line 7:
```suggestion
        extra=""  # Optional: extra parameters for multiqc.
```

fgvieira:
I find the variable name `convert_file_input_to_dirs` a bit confusing. Could we have something a bit simpler, like (e.g.) `use_input_files_only`, `use_input_dirs`, `scan_input_dirs`?

Smeds:
> I find the variable name `convert_file_input_to_dirs` a bit confusing. Could we have something a bit simpler, like (e.g.) `use_input_files_only`, `use_input_dirs`, `scan_input_dirs`?

good suggestion! I have updated the pull-request with variable name use_input_files_only instead of convert_file_input_to_dirs

fgvieira:
I think it looks goos, but could you also describe this new parameter in the `meta.yaml` file?
Just so it is explained in the documentation.

bio/multiqc/meta.yaml
```diff
@@ -4,6 +4,8 @@ description: |
 authors:
```
fgvieira, line 4:
```suggestion
url: https://multiqc.info/
authors:
```

fgvieira (approved):

PR #500, opened 2022-06-20
fix: allow for arbitrary filename prefixes in the optitype wrapper

dlaehnemann:
### Description

This is a first attempt, will update with a tmp-folder based solution in a bit.

### QC

* [ ] either the wrapper can only use a single core, or the example rule contains a `threads: x` statement with `x` being a reasonable default,
* [ ] `Snakefile`s are formatted with [snakefmt](https://github.com/snakemake/snakefmt),
* [ ] Python wrapper scripts are formatted with [black](https://black.readthedocs.io).


bio/optitype/wrapper.py
```diff
@@ -20,13 +19,16 @@
...
+    shell(
+        "(OptiTypePipeline.py"
+        "  --input {snakemake.input.reads}"
+        "  --outdir {tempdir}"
+        "  --prefix tmp_prefix"
+        "  {seq_type}"
+        "  {config}"
+        "  {extra}; "
+        " mv {tmpdir}/tmp_prefix_coverage_plot.pdf {snakemake.output.pdf:q} ;"
+        " mv {tmpdir}/tmp_prefix_result.tsv {snakemake.output.tsv:q} )"
```
fgvieira, line 32: Would this also work in case `output.tsv` (or `output.pdf`) is a pipe?

dlaehnemann: Probably not, although I'm not sure. Do you have that use case somewhere? Did this work with the previous setup? Or is this just a general question?

fgvieira: Yes, I guess in your case it should be safe to assume that the outputs won't be piped, but I still think it would be nice to have a general solution (since we never know if someone might want to pipe the `TSV` file).

Why did you decide to change the wrapper from the previous behaviour to this new "output-to-temp-and-move" behaviour?

From the top of my head, when we have programs that require an `output directory` and an `output prefix`, the advantages of "output-to-temp-and-move" are:

1. if programs creates lots of temp files and user usually only interested in a few (easier to clean-up than having to specify all temp output files/folders in the rule)
2. allows to specify different paths for each output file
3. allows to specify custom names for output files

In theory it looks quite a clean solution but, as stated before, it either creates extra I/O or breaks snakemake `pipe` functionality. For this reason, I've been thinking that it might not be the best solution after all, specially because 1) can (in theory) also be done by using a [shadow rule](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#shadow-rules), and 2 and 3 are quite minor issues.

However, I am not completely sure if I am missing any other advatntages/drawbacks...

dlaehnemann: The most important reason for allowing arbitrary output file names. Relying on the specific names of wildcards in a snakemake wrapper is not optimal, I ran into that with this wrapper when trying to use it with a more flexible set of wildcards. Which is why we try to enforce that wrappers can have arbitrary output file names, to make them more reusable. So I would argue that that is more important than ensuring that a rule is always pipeable.

fgvieira: I am mostly concerned in cases where it causes extra I/O, since that can considerably slowdown execution.

But maybe it's ok in some circumstances, and we should just avoid it in cases where the output is expected to be large (e.g. STAR)?

tedil (approved):

PR #494, opened 2022-06-08
feat: Add wrapper for GATK DepthOfCoverage

ljmesi:
### Description

This PR adds a wrapper for GATK DepthOfCoverage tool.

### QC

For all wrappers added by this PR, I made sure that

* [x] there is a test case which covers any introduced changes,
* [x] `input:` and `output:` file paths in the resulting rule can be changed arbitrarily,
* [x] either the wrapper can only use a single core, or the example rule contains a `threads: x` statement with `x` being a reasonable default,
* [x] rule names in the test case are in [snake_case](https://en.wikipedia.org/wiki/Snake_case) and somehow tell what the rule is about or match the tools purpose or name (e.g., `map_reads` for a step that maps reads),
* [x] all `environment.yaml` specifications follow [the respective best practices](https://stackoverflow.com/a/64594513/2352071),
* [x] wherever possible, command line arguments are inferred and set automatically (e.g. based on file extensions in `input:` or `output:`),
* [x] all fields of the example rules in the `Snakefile`s and their entries are explained via comments (`input:`/`output:`/`params:` etc.),
* [x] `stderr` and/or `stdout` are logged correctly (`log:`), depending on the wrapped tool,
* [x] temporary files are either written to a unique hidden folder in the working directory, or (better) stored where the Python function `tempfile.gettempdir()` points to (see [here](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir); this also means that using any Python `tempfile` default behavior works),
* [x] the `meta.yaml` contains a link to the documentation of the respective tool or command,
* [x] `Snakefile`s pass the linting (`snakemake --lint`),
* [x] `Snakefile`s are formatted with [snakefmt](https://github.com/snakemake/snakefmt),
* [x] Python wrapper scripts are formatted with [black](https://black.readthedocs.io).

PR #515, opened 2022-07-06
feat: BWA mem2 index - remove prefix parameter and determine prefix from output

christopher-schroeder:
### Description

Remove the param prefix and detect prefix by output

### QC


bio/bwa-mem2/index/wrapper.py
```diff
@@ -16,13 +16,6 @@
...
 # Prefix that should be used for the database
-prefix = None
-if "prefix" in snakemake.params.keys():
-    prefix = snakemake.params["prefix"]
-else:
-    prefix = splitext(snakemake.output[0])[0]
-
-if len(prefix) > 0:
-    prefix = "-p " + prefix
+prefix = "-p" + path.splitext(snakemake.output[0])[0]
```
tedil, line 19: `splitext` will only remove the very last extension, e.g. 
```python
splitext("genome.fasta.gz") == ("genome.fasta", ".gz")
splitext("genome.fasta") == ("genome", ".fasta")
```
Is this the desired behaviour?

christopher-schroeder: The output[0] gets split, not the input. But you are right anyway, because there is an output file with suffix ".bwt.2bit.64". The case that that this file is the first declared output file is now covered.

tedil (approved):
LGTM!

fgvieira:
Wouldn't it be cleaner to infer the prefix with `os.path.commonprefix()`?

bio/bwa-mem2/index/wrapper.py
```diff
@@ -15,14 +15,19 @@
...
+    prefixes.append(s[: -len(suffix)])
 
-if len(prefix) > 0:
-    prefix = "-p " + prefix
+if len(set(prefixes)) != 1:
+    raise ValueError("Output files must share common prefix up to their file endings.")
+prefix = prefixes[0]
 
-shell("bwa-mem2 index" " {prefix}" " {snakemake.input[0]}" " {log}")
+shell(f"bwa-mem2 index -p {prefix} {snakemake.input[0]} {log}")
```
fgvieira, line 33:
```suggestion
shell("bwa-mem2 index -p {prefix} {snakemake.input[0]} {log}")
```

fgvieira:
Like mentioned on issue #494, maybe we can define a function to infer the prefixes and add it to `snakemake-wrapper-utils`. Someting like:

```
def infer_prefix(files, suffixes, strict = True):
    prefixes = []
    suffixes = set(suffixes)
    
    for file in files:
        for suffix in suffixes:
            if file.endswith(suffix):
                prefixes.append(file[:-len(suffix)])

    if len(prefixes) != len(files) and strict:
        raise ValueError("All files must have a valid suffix.")
    if len(set(prefixes)) != 1:
        raise ValueError("All files must share common prefix.")

    return prefixes[0]
```

bio/bwa-mem2/index/wrapper.py
```diff
@@ -15,14 +15,19 @@
 elif len(snakemake.input) > 1:
     raise ValueError("Please provide exactly one reference genome as input.")
 
-# Prefix that should be used for the database
-prefix = None
-if "prefix" in snakemake.params.keys():
-    prefix = snakemake.params["prefix"]
-else:
-    prefix = splitext(snakemake.output[0])[0]
+valid_suffixes = [".0123", ".amb", ".ann", ".bwt.2bit.64", ".pac"]
```
johanneskoester, line 18:
```suggestion
valid_suffixes = {".0123", ".amb", ".ann", ".bwt.2bit.64", ".pac"}
def get_valid_suffix(path):
    for suffix in valid_suffixes:
        if path.endswith(suffix):
            return suffix
```

bio/bwa-mem2/index/wrapper.py
```diff
@@ -15,14 +15,19 @@
...
+prefixes = []
+for s in snakemake.output:
+    if s.endswith(".bwt.2bit.64"):
+        suffix = ".bwt.2bit.64"
+    else:
+        suffix = s[:-4]
+    if not suffix in valid_suffixes:
```
johanneskoester, line 25:
```suggestion
    suffix = get_valid_suffix(s)
    if suffix is None:
```

johanneskoester (changes requested):
Nice work, some suggestions below. In general, I like the idea of @fgvieira to move the suffix handling code into snakemake-wrapper-utils, but to get things going I am fine with committing it here for now and moving it there later.

johanneskoester (approved):

PR #533, opened 2022-08-08
feat: add "qualimap bamqc" wrapper

mobilegenome:
### Description

`qualimap bamqc` is a tool to provide comprehensive quality control reports for alignments in BAM format. 

The wrapper repository already contains a wrapper for `qualimap rnaseq`, this PR migrates this wrapper for `bamqc`. 

### QC

* [x] I confirm that:


bio/qualimap/bamqc/environment.yaml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - bioconda
+  - conda-forge
+  - defaults
+dependencies:
+  - qualimap ==2.2.2d
```
fgvieira, line 6:
```suggestion
  - qualimap =2.2
```

bio/qualimap/bamqc/wrapper.py
```diff
@@ -0,0 +1,29 @@
...
+
+import os
+
+from snakemake.shell import shell
+
+java_opts = snakemake.params.get("java_opts", "")
+if java_opts:
+    java_opts_str = f'JAVA_OPTS="{java_opts}"'
+else:
+    java_opts_str = ""
```
fgvieira, line 16: You can use `snakemake-wrapper-utils`:
https://github.com/snakemake/snakemake-wrapper-utils/blob/master/snakemake_wrapper_utils/java.py

test.py
```diff
@@ -4288,6 +4288,12 @@ def test_qualimaprnaseq():
         ["snakemake", "--cores", "1", "--use-conda", "-F"],
     )
 
+@skip_if_not_modified
+def test_qualimapbamqc():
+    run(
+        "bio/qualimap/bamqc",
+        ["snakemake", "--cores", "1", "--use-conda", "-F"],
```
fgvieira, line 4295:
```suggestion
        ["snakemake", "--cores", "1", "qc/a", "--use-conda", "-F"],
```

bio/qualimap/bamqc/test/Snakefile
```diff
@@ -0,0 +1,14 @@
+rule qualimap:
+    input:
+        # BAM aligned, splicing-aware, to reference genome
+        bam="mapped/a.bam",
+    output:
+        directory("qc/a"),
```
fgvieira, line 6: IS the output just a HTML/PDF report? If so, it would be better to specifically name the output file, instead for the output folder.

bio/qualimap/bamqc/environment.yaml
mobilegenome: Thanks for suggestions! The latest version in bioconda is specifically 2.2.2d, why would you only specify the minor version?   Not opposing it, but would like to understand :)

fgvieira: It is just so we don't have to update the wrapper every time there is a small bugfix.
Assuming patch/micro versions are just small bugfixes that do not affect the overall program execution/output, it should be safe to always use the latest one.

bio/qualimap/bamqc/test/Snakefile
mobilegenome: I looked into this. The output file is always named `genome_results.txt` inside the directory currently specified. I think it would be confusing if the output file would be specified in the rule. This would create the impression the filename could be changed, and lead to errors if a different name would be specified. Thus, I would suggest to keep the output as directory.

test.py
mobilegenome: done

bio/qualimap/bamqc/wrapper.py
mobilegenome: done

bio/qualimap/bamqc/meta.yaml
```diff
@@ -0,0 +1,13 @@
+name: qualimap rnaseq
+description: |
+        Run qualimap bamqc to create a QC report for aligned NGS data data. It can be
+        used for WGS, WES, RNA, ChIP-Seq, etc.
```
fgvieira, line 4:
```suggestion
        used for WGS, WES, RNA, ChIP-Seq, etc.
url: http://qualimap.conesalab.org/doc_html/analysis.html#bam-qc
```

bio/qualimap/bamqc/test/Snakefile
fgvieira: But is it the only output file?
Alternatively, we could set the output folder to a temp folder (using `tempfile`) and at the end, depending on which output files were specified, copy them to their final destination (e.g. like in [star](https://github.com/snakemake/snakemake-wrappers/blob/master/bio/star/align/wrapper.py)).

bio/qualimap/bamqc/meta.yaml
```diff
@@ -0,0 +1,14 @@
...
+url: http://qualimap.conesalab.org/doc_html/analysis.html#bam-qc
+authors:
+  - Fritjof Lammers
+  - Brett Copeland
+input:
+  - BAM file of data aligned to genome
+output:
+  - QC report in TXT format (genome_results.txt)
+notes: |
+  * For more information, see http://qualimap.conesalab.org/doc_html/analysis.html#bam-qc.
```
fgvieira, line 14:
```suggestion
  * The `extra` param allows for additional program arguments.
```

bio/qualimap/bamqc/test/Snakefile
mobilegenome: This is actually a very good suggestion, but unfortunately it breaks the MultiQC compatibility that is based on exact file namings of the TXT and HTML reports based on the programs default. As I assume MultiQC is a major use-case of bamqc, it is definitly the better option to use the `directory` directive instead of specifying exact filenames.

I've add a parameter to create HTML reports by defaults, as these are also parsed by MultiQC and documented its use.

fgvieira: But that then would be up to the user to properly define the output file name, no?
I was just wondering if, under some circumstances, the user might want to have different files placed in different places.

mobilegenome: In my view, the wrapper should execute the tools' defaults parameters. In case of `qualimap` this is handling the output via a directory using fixed filenames. For the HTML output its even more complicated, because the output consists of a bunch of folders of files, containing all CSS, HTML and images. 

I would know of any use-case right know, for which renaming the file brings a benefit. Instead of providing a complicated do-it-all solution, I would stick with the most-common use-case, that is also consistent with the `qualimap/rnaseq` wrapper. If users want to rename the output, this could be also easily implemented with an additional rule that renames the file, but again, I hardly see a use-case for that.

mobilegenome: Hi @fgvieira, I don't mean to push this review - but while my mind is still fresh on the topic - wanted to ask if there are any more concerns or if I can provide more information/changes to make it pass. :)

fgvieira: I tend to always avoid `directory()` and refer to the output files directly, since it allows to reference specific files in downstream rules and mark unnecessary files as `temp()`.

Also in the PR checklist, it states that:
- `input:` and `output:` file paths in the resulting rule can be changed arbitrarily,

but not sure if it refers to single files or also to folders, or how much @johanneskoester wants to enforce it.

As for the rest, I think it looks great! 😄

johanneskoester (approved):

bio/qualimap/bamqc/test/Snakefile
johanneskoester: Sorry folks. I merged this accidentally before these have been fixed. I'll fix them myself now since it was my fault.

johanneskoester: Thinking about it, nowadays people might rather use it via multiqc. Hence, I will close my PR and leave it as it is.

PR #535, opened 2022-08-12
feat: add datavzrd wrapper

FelixMoelder:
### Description

A wrapper for datavzrd

### QC

* [x] I confirm that:


utils/datavzrd/wrapper.py
```diff
@@ -0,0 +1,10 @@
+__author__ = "Johannes Köster"
+__copyright__ = "Copyright 2017, Johannes Köster"
+__email__ = "johannes.koester@protonmail.com"
+__license__ = "MIT"
+
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+shell("datavzrd {snakemake.input.config} --output {snakemake.output[0]} {log}")
```
fgvieira, line 10: Is `input.table` not used?

utils/datavzrd/meta.yaml
```diff
@@ -0,0 +1,7 @@
+name: datavzrd
+description: |
+  datavzrd allows to render tables by providing a configuration file.
+  All other input variables are optional but the user has to ensure that files linked
+  within the configuration file are present. 
+authors:
```
fgvieira, line 6:
```suggestion
url: https://github.com/koesterlab/datavzrd
authors:
```

utils/datavzrd/wrapper.py
FelixMoelder: datavzrd only takes a config file as input without any option for extra parameters.
Therefore the only mandatory input field is `input.config`. Still, datavzrd utilizes tables and json-files defined within the config-file.
In the testcase I added `input.table` as an example as it appears reasonable to me defining supplementary files being processed by datavzrd.
I admit that this is not intuitive which is why I added a small description in the `meta.yaml`.
In general the wrapper is kind of a pseudo wrapper as it does not do more than calling `datavzrd` without any further preprocessing.

Also the report-directive appears to be unnecessary for a minimal example but datavzrd reports are often embedded into snakemake reports which is why this is also added to the testcase.

The reason for making it a wrapper is that datavzrd is required for several workflows and it at least provides us a template rule which can be customized with low effort.

fgvieira: Is the config file pretty standard or are there a lot of options to tweak? Because it could also be generated by the wrapper itself, from user-specified parameters. It might be overkill, though...

FelixMoelder: The config file is pretty custom and describes the layout of the final html-report.
In our current setup we also perform parsing of a template config file using `yte`.
Still it might be possible to add this to the wrapper as `yte` offers a python api.
@johanneskoester What's your opinion on that?

johanneskoester: Automatic inference of the config file makes no sense here, because it really lives from the semantic added to it beyond just the file paths. Above the rule, there should however be a link to the section on template rendering, because most of the time people will do that when using this wrapper: https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#template-rendering-integration

utils/datavzrd/environment.yaml
```diff
@@ -0,0 +1,4 @@
+channels:
+  - conda-forge
```
johanneskoester, line 2:
```suggestion
  - conda-forge
  - nodefaults
```

utils/datavzrd/meta.yaml
```diff
@@ -0,0 +1,10 @@
+name: datavzrd
+description: |
+  datavzrd allows to render tables by providing a configuration file. 
+  Configuration templates can be dynamically customized by utilizing the 
+  `rendering integration <https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#template-rendering-integration>`_.
+  Additional input variables are optional but recommended to ensure that files linked 
+  within the configuration file are present.
```
johanneskoester, line 7:
```suggestion
  Any files specified in the configuration file have to be also specified as additional input files in the
  datavzrd rule.
```

utils/datavzrd/test/Snakefile
```diff
@@ -0,0 +1,23 @@
+rule all:  # [hide]
+    input:  # [hide]
+        "results/datavzrd-report/A",  # [hide]
+
+
+rule datavzrd:
+    input:
+        config="resources/{sample}.datavzrd.yaml",
+        # optional files required for rendering
```
johanneskoester, line 9:
```suggestion
        # optional files required for rendering the given config
```

johanneskoester (changes requested):
Some small suggestions below.

utils/datavzrd/test/Snakefile
```diff
@@ -0,0 +1,23 @@
...
+        table="data/A.tsv",
+    output:
+        report(
+            directory("results/datavzrd-report/{sample}"),
+            htmlindex="index.html",
+            caption="../report/calls.rst",
+            category="Variant calls",
+            labels=["Fancy label"],
+            subcategory="somatic variants",
```
johanneskoester, line 18:
```suggestion
            # see https://snakemake.readthedocs.io/en/stable/snakefiles/reporting.html
            # for additional options like caption, categories and labels
```

johanneskoester (approved):

PR #540, opened 2022-08-16
feat: Merqury

fgvieira:
### QC

* [x] I confirm that:


bio/merqury/test/Snakefile
```diff
@@ -0,0 +1,61 @@
...
+        qv="results/haploid/out.qv",
+        spectra_asm_hist="results/haploid/out.spectra_asm.hist",
+        spectra_asm_ln_png="results/haploid/out.spectra_asm.png",
+        fas1_only_bed="results/haploid/hap1.bed",
+        fas1_only_wig="results/haploid/hap1.wig",
+        fas1_only_hist="results/haploid/hap1.hist",
+        fas1_qv="results/haploid/hap1.qv",
+        fas1_spectra_hist="results/haploid/hap1.spectra.hist",
+        fas1_spectra_ln_png="results/haploid/hap1.spectra.png",
+        logs=directory("logs/haploid/"),
```
johanneskoester, line 20: Is this really an output or rather should be put into log?

bio/merqury/wrapper.py
```diff
@@ -0,0 +1,104 @@
...
+        shell("cat out.spectra-cn.ln.png > {snakemake.output.spectra_cn_ln_png}")
+    if snakemake.output.get("spectra_cn_fl_png"):
+        shell("cat out.spectra-cn.fl.png > {snakemake.output.spectra_cn_fl_png}")
+    if snakemake.output.get("spectra_cn_st_png"):
+        shell("cat out.spectra-cn.st.png > {snakemake.output.spectra_cn_st_png}")
+
+    if snakemake.output.get("hapmers_count"):
+        shell("cat out.hapmers.count > {snakemake.output.hapmers_count}")
+    if snakemake.output.get("hapmers_png"):
+        shell("cat out.hapmers.blob.png > {snakemake.output.hapmers_png}")
```
johanneskoester, line 65: Maybe use mv for all of these? That would be faster than cat and pipe.

johanneskoester (changes requested):

bio/merqury/wrapper.py
fgvieira: I switched PNG output files to `mv`. As for the others, what if the user wants to pipe the outputs?

bio/merqury/test/Snakefile
fgvieira: This program is a bit of a mess... saves all outputs and logs to the `cwd`, and most output names are hard coded (that is why I recommend running it as a shadow rule). Among those, it creates a `log` folder with a couple of log files inside. I guess I could specify several log files... Do you think that would be cleaner?

bio/merqury/wrapper.py
johanneskoester: Good point.

bio/merqury/test/Snakefile
johanneskoester: Or add that folder to the log directive.

bio/merqury/test/Snakefile
```diff
@@ -0,0 +1,65 @@
...
+    log:
+        "logs/haploid.log",
+    threads: 1
+    shadow:
+        "minimal"
```
johanneskoester, line 25: Instead of shadow, I'd rather let the tool itself run in a temp dir as cwd. This way, the complexity is hidden from the user.

bio/merqury/test/Snakefile
fgvieira: But then cannot use `log_shell_fmt`, right?

bio/merqury/test/Snakefile
fgvieira: It got a bit messy, but it should work. :smile:

bio/merqury/test/Snakefile
johanneskoester: yes, atm that is true

johanneskoester: but I think that is worth it.

fgvieira: Ok, just pushed new changes.

bio/merqury/wrapper.py
```diff
@@ -0,0 +1,138 @@
...
+            out = cwd / out
+            shell("mv {out_prefix}.{prefix}.spectra-cn.ln.png {out}")
+        out = snakemake.output.get(f"fas{fas}_spectra_fl_png")
+        if out:
+            out = cwd / out
+            shell("mv {out_prefix}.{prefix}.spectra-cn.fl.png {out}")
+        out = snakemake.output.get(f"fas{fas}_spectra_st_png")
+        if out:
+            out = cwd / out
+            shell("mv {out_prefix}.{prefix}.spectra-cn.st.png {out}")
```
johanneskoester, line 138: Sorry, another request, I just noticed this. This is all very similar code. I'd say just turn this into one or a few functions and call them in a loop over the file suffixes.

johanneskoester (changes requested):

bio/merqury/wrapper.py
fgvieira: Done

johanneskoester (approved):

PR #544, opened 2022-08-18
feat: Wrapper and meta-wrapper for rbt consensus reads

FelixMoelder:
### Description

This adds two wrappers.
First a wrapper for rust-bio-tools consensus read calculation and second a meta wrapper which adds best practice post processing steps for consensus read calculation.

### QC

* [x] I confirm that:


meta/bio/rbt_calc_consensus/meta.yaml
```diff
@@ -0,0 +1,5 @@
+name: calc_consensus_reads
+description: |
+    Performs consensus read calculation on a marked bam records and merges single,
```
johanneskoester, line 3:
```suggestion
    Performs consensus read calculation on a marked bam and merges single, 
```

meta/bio/rbt_calc_consensus/test/common.smk
```diff
@@ -0,0 +1,7 @@
+def get_processed_consensus_input(wildcards):
```
johanneskoester, line 1: Please put this into the main Snakefile. People have to see the entire meta-wrapper code, so that they case decide how to use it (copy-paste or as a module).

meta/bio/rbt_calc_consensus/test/Snakefile
```diff
@@ -0,0 +1,104 @@
...
+        "master/bio/picard/markduplicates"
+
+
+rule merge_consensus_reads:
+    input:
+        "results/consensus/{sample}.skipped.marked.bam",
+        "results/consensus/{sample}.se.mapped.bam",
+        "results/consensus/{sample}.pe.mapped.bam",
+    output:
+        "results/consensus/{sample}.merged.bam",
```
johanneskoester, line 99:
```suggestion
        "results/consensus/{sample}.bam",
```

johanneskoester: Is this a sorted bam then? That would be ideal. Maybe mention this as a comment, so that people don't have to wonder about it.

meta/bio/rbt_calc_consensus/test/Snakefile
```diff
@@ -0,0 +1,104 @@
...
+    input:
+        "resources/chr21.fa",
+    output:
+        idx=multiext("resources/chr21.fa", ".amb", ".ann", ".bwt", ".pac", ".sa"),
+    log:
+        "logs/bwa_index.log",
+    resources:
+        mem_mb=369000,
+    wrapper:
+        "master/bio/bwa/index"
```
johanneskoester, line 19:
```suggestion
```

Remove the index and consider it to be given. When people include this in their workflow, they will already have one because they have already generated a bam.

meta/bio/rbt_calc_consensus/test/Snakefile
```diff
@@ -0,0 +1,104 @@
+include: "common.smk"
+
+
+rule all:  # [hide]
+    input:  # [hide]
+        "results/consensus/sampleA.merged.bam",  # [hide]
```
johanneskoester, line 6:
```suggestion
```

Should also be removed. meta-wrappers should not be used standalone but inside of a larger workflow.

meta/bio/rbt_calc_consensus/test/Snakefile
```diff
@@ -0,0 +1,104 @@
...
+rule calc_consensus_reads:
+    input:
+        "mapped/{sample}.marked.sorted.bam",
+    output:
+        consensus_r1=temp("results/consensus/{sample}.1.fq"),
+        consensus_r2=temp("results/consensus/{sample}.2.fq"),
+        consensus_se=temp("results/consensus/{sample}.se.fq"),
+        skipped=temp("results/consensus/{sample}.skipped.bam"),
+    params:
+        extra="--verbose-read-names",
```
johanneskoester, line 47: Maybe add a comment with an explanation.

johanneskoester (changes requested):

meta/bio/rbt_calc_consensus/test/common.smk
FelixMoelder: I did so, but in that case the linting test failed.

meta/bio/rbt_calc_consensus/test/Snakefile
FelixMoelder: Yes, I added this in the description of the meta-wrapper.

FelixMoelder: Removing the `.merged` suffix lets snakemake run endless as the `{sample}` wildcard would be valid for most of the other rules.
An option would be to exclude a `.` in the wildcard_constraints but this could cause issues if the wrapper is added to a workflow where a `.` is part of a later rule.

johanneskoester (changes requested):

meta/bio/rbt_calc_consensus/test/Snakefile
```diff
@@ -0,0 +1,90 @@
...
+        temp("{path}/{sample}.sorted.bam"),
+    params:
+        extra="-m 4G",
+        tmp_dir="/tmp/",
+    log:
+        "{path}_{sample}.log",
+    # Samtools takes additional threads through its option -@
+    threads: 8  # This value - 1 will be sent to -@.
+    wrapper:
+        "master/bio/samtools/sort"
```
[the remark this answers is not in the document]
FelixMoelder: I changed it to a fixed directory and set an appropriate rule name.
Previously, that rule was used two times. First ensuring that the initial input of the workflow gets sorted and another time later within the workflow of meta wrapper.
Now, I changed the wrapper in a way that the initial bam file is expected to be presorted as an unsorted bam-file would require the sorting rule to be generic.

johanneskoester (approved):

PR #547, opened 2022-08-29
fix: upgrade vep wrappers to latest version, fixing issues with strict channel priorities.

johanneskoester:
### QC

* [x] I confirm that:

### DIFF
--- a/bio/vep/annotate/environment.yaml
+++ b/bio/vep/annotate/environment.yaml
@@ -5,4 +5,4 @@ channels:
 dependencies:
   - ensembl-vep =107
   - bcftools =1.15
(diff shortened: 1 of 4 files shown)


dlaehnemann:
I think this actually needs a fix in the conda recipe of ensembl-vep, see here:
https://github.com/bioconda/bioconda-recipes/pull/36693

johanneskoester:
@fgvieira in case you have any idea how to resolve this error, that would be highly appreciated. Somehow perl can't find certain modules (e.g. lwp-simple, encode-locale) although they should be there actually (and reportedly are when trying the env locally).

PR #580, opened 2022-10-12
feat: bazam wrapper

christopher-schroeder:
### QC

* [x] I confirm that:


bio/bazam/wrapper.py
```diff
@@ -0,0 +1,31 @@
...
+
+# Extract arguments.
+if reads := snakemake.output.get("reads", ""):
+    out_cmd = f"-o {reads}"
+elif (r1 := snakemake.output.get("r1", "")) and (r2 := snakemake.output.get("r2", "")):
+    out_cmd = f"-r1 {r1} -r2 {r2}"
+else:
+    raise ValueError("either 'reads' or 'r1' and 'r2' must be specified in output")
+
+shell("(bazam -Xmx12g {reference_cmd} {extra} -bam {bam} {out_cmd}) {log}")
```
johanneskoester, line 31: Instead of hardcoding the memory here, it is better to use our get_java_opts helper, which infers such arguments from the resources definition of the rule.
See [here](https://github.com/snakemake/snakemake-wrappers/blob/a67765ce5fe2e97685072b1d30d84fcd461d507a/bio/picard/markduplicates/wrapper.py#L14) for usage, and [here](https://github.com/snakemake/snakemake-wrappers/blob/a67765ce5fe2e97685072b1d30d84fcd461d507a/bio/picard/markduplicates/test/Snakefile#L19) for how to encode the mem in the resources section of the example rule.

johanneskoester (changes requested):
Excellent, just one tiny thing below.

johanneskoester (approved):

PR #594, opened 2022-10-13
feat: move pindel region bed files to input

maehler:
### Description

Recently I discovered that if you pass a non-existent bed file as a value for the command line parameters `-j/--include` or `-J/--exclude` in pindel, the program behaves as if empty files are being passed. Since these are supplied in `params.extra`, no checks are done on them.

I realise that this is not an issue with the wrapper, per se, but the pindel repo has been archived, and I think that modifying the wrapper might be a good way to catch potential mistakes.

Not sure whether my implementation is the best way to handle things, so thoughts and opinions are welcome.

### QC

* [x] I confirm that:


fgvieira:
Why not just have the BED file as an input file? This way `snakemake` will check if the file exists.

maehler:
I was thinking to have the bed files as inputs, but that more or less also makes them required, which I kind of wanted to avoid. If I recall correctly, it is possible to set an input to an empty list for it to not fail, and that could then be handled by the wrapper, but that's also kind of ugly. What do you think?

Good point about the prefix. I just followed the existing wrapper blindly. Having the prefix managed by the wrapper would be much cleaner.

maehler:
> I was thinking to have the bed files as inputs, but that more or less also makes them required, which I kind of wanted to avoid. If I recall correctly, it is possible to set an input to an empty list for it to not fail, and that could then be handled by the wrapper, but that's also kind of ugly.

Or no, I'm being a square. Whether the argument would be optional or not would be handled by the wrapper. Much cleaner.

fgvieira:
Just modify this one...

bio/pindel/call/wrapper.py
```diff
@@ -3,13 +3,34 @@
...
+if include_bed is not None and exclude_bed is not None:
+    raise Exception("supply either include_bed or exclude_bed, not both")
+
+include_arg = ""
+if include_bed is not None:
+    include_arg = "-j {}".format(include_bed)
+
+exclude_arg = ""
+if exclude_bed is not None:
+    exclude_arg = "-J {}".format(exclude_bed)
```
fgvieira, line 23:
```suggestion
    exclude_arg = f"-J {exclude_bed}"
```

bio/pindel/call/wrapper.py
```diff
@@ -3,13 +3,34 @@
...
 
+include_bed = snakemake.input.get("include_bed", None)
+exclude_bed = snakemake.input.get("exclude_bed", None)
+
+if include_bed is not None and exclude_bed is not None:
+    raise Exception("supply either include_bed or exclude_bed, not both")
+
+include_arg = ""
+if include_bed is not None:
+    include_arg = "-j {}".format(include_bed)
```
fgvieira, line 19:
```suggestion
    include_arg = f"-j {include_bed}"
```

fgvieira:
It looks great! Just some small cosmetic suggestions. :smile:

fgvieira (approved):

maehler:
> It looks great! Just some small cosmetic suggestions. 😄

Great, thanks for the input! This turned out a lot better.

fgvieira:
It seems there is a conflict on `test.py`. Can you take a look at it?

maehler:
Should be good to go now.

fgvieira (approved):

PR #772, opened 2022-11-03
feat: gbasf2_wrapper

casschmitt:
### Description

Adding a wrapper called gbasf2, enabeling snakemake to submit rules as jobs to the Belle II grid, similar to the LHC Computing Grid. Intended to be used for workflows on the KEKcc computing cluster only.

PS: Input file paths to the Belle II grid are always absolute, which gives an unavoidable warning in snakemake linting.

### QC

* [x] I confirm that:


fgvieira:
Is this wrapper intended to submit jobs to a specific cluster?
if so, why not use snakemake profiles (e.g. [here](https://github.com/orgs/Snakemake-Profiles/repositories))?

casschmitt:
As indicated above:
"Input file paths to the Belle II grid are always absolute, which gives an unavoidable warning in snakemake linting."
Is there a way to circumvent the failed linting check above? Thank you.

casschmitt:
> Is this wrapper intended to submit jobs to a specific cluster? if so, why not use snakemake profiles (e.g. [here](https://github.com/orgs/Snakemake-Profiles/repositories))?

I am not using a profile, since only some rules need to be submitted to the cluster and not all rules of a typical Belle II workflow. With a wrapper, I can specify per rule what goes on the cluster.

fgvieira:
The rules that are not submitted to the cluster, are they intended to be run locally? Because if so, they could be defined as [local rules](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#local-rules).

casschmitt:
> The rules that are not submitted to the cluster, are they intended to be run locally? Because if so, they could be defined as [local rules](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#local-rules).

In a typical Belle II workflow, we have 3 types of rules. The ones that run on our dedicated cluster (with the wrapper), the ones that run on a qsub cluster (--cluster qsub) and the ones that run locally (localrules). This is the only way I was able to implement it in a snakemake workflow and it works very nicely.

PS: To circumvent the linting error, I could read in the absolute path from a separate file. Would this be preferred? Thank you

fgvieira:
From what you are saying, it seems like you'd like to be able to specify rule-specific profiles. If so, then I think the best way would be to do it through `snakemake`, and not through a wrapper. This would be quite neat, as it would allow `snakemake` to work with an arbitrary number (and type) of clusters.

casschmitt:
> From what you are saying, it seems like you'd like to be able to specify rule-specific profiles. If so, then I think the best way would be to do it through `snakemake`, and not through a wrapper. This would be quite neat, as it would allow `snakemake` to work with an arbitrary number (and type) of clusters.
> 
> @johanneskoester what do you think?
 
Rule-specific profiles would be very nice. For now though, we might use the wrapper and make it publicly available for analysts? Once rule-specific profiles are available in snakemake, we could transfer to a profile.

casschmitt:
Dear all, I moved the absolute path into a separate file. In this way snakemake linting throws no warnings. Is there any updated opinion on this merge? Thank you

casschmitt:
It's been two weeks and I am trying to provide this wrapper to collaborators. Are there any updates on how to proceed @johanneskoester @fgvieira? Thank you.

casschmitt:
Dear all, as far as I know the snakemake wrapper repository does not come with GPL3 compatible license. However, it turns out that some collaborators require this. Therefore I will close this pull request and publish the wrapper elsewhere with GPL3 compatible license. Thank you for your help nevertheless!

PR #964, opened 2022-12-09
feat: improve star_arriba metawrapper

FelixMoelder:
### Description

When creating the star index the path of the annotation file is currently hardcoded. As it is defined in the input I would suggest to make it a variable.

In addition I would like to discuss if the Test-Snakefile should be more generic.
The `arriba`-rule currently definies a blacklist-file as parameter while it is not part of the input.
In addition the contigs are also limited to chromosome 1 and 2.
As I would expect that a meta wrapper should work out of the box with default parameters when imported as a module it appears reasonable to me to remove this parameters (if this does not make the CI run forever). 

### QC

* [x] I confirm that:

### DIFF
--- a/meta/bio/star_arriba/test/Snakefile
+++ b/meta/bio/star_arriba/test/Snakefile
@@ -6,7 +6,7 @@ rule star_index:
         directory("resources/star_genome"),
     threads: 4
     params:
-        extra=lambda wc, input: "--sjdbGTFfile {input.annotation} --sjdbOverhang 100",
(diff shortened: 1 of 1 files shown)


fgvieira:
This will also give errors once the version is updated, no?
```
extra="-T -P -i 1,2",
```

FelixMoelder:
> This will also give errors once the version is updated, no?
> 
> ```
> extra="-T -P -i 1,2",
> ```

Yes, this test is already running when updating arriba.
I therefore decided to add this into PR https://github.com/snakemake/snakemake-wrappers/pull/963 as it depends on each other.

PR #970, opened 2023-01-06
fix: RSEM-calculate-expression wrapper with FASTQ-files as input + Integration of test case

KuechlerO:
### Description

- Replaced deprecated argument for alignment inputs with appropriate argument (`--bam`-> `--alignments`)
- Fixed wrong argument for FASTQ-input: Deleted `input_bam = False`
- Integrated test-case for FASTQ-input 
- Fixed environment-yaml-file: bowtie is needed for FASTQ-file input
- updated meta.yaml

### Notes
The original provided test-case from the authors for BAM-file inputs includes only an empty BAM-file (except of the header). My test case also includes reads, which do not map to the reference sequence (since the provided reference sequence is too short anyways). So as I understand, the purpose of these tests is just to ensure, whether the pipeline is able to run through. However, to make it more robust additional tests with real working examples would be needed.


johanneskoester (approved):

bio/rsem/calculate-expression/meta.yaml
```diff
@@ -8,6 +8,8 @@ input:
   - bam: BAM file with reads aligned to transcriptome
   - fq_one: FASTQ file of reads (read_1 for paired-end sequencing)
   - fq_two: Optional second FASTQ file of reads (read_2 for paired-end sequencing)
+  - reference: One of the index files created by rsem-prepare-reference; the file suffix is stripped and passed on to rsem
+  - reference_bowtie: Only needed for FASTQ input; One of the bowtie index files created by bowtie-build
```
johanneskoester, line 12: We should rather require all index files. multiext can be used for that. Otherwise, the wrapper only works on shared filesystems.

johanneskoester (changes requested):

bio/rsem/calculate-expression/meta.yaml
KuechlerO: Ok, I've now employed multiext as requested.

johanneskoester (approved):

PR #974, opened 2023-01-11
fix: Unpin minor version for cooltools wrappers

Phlya:


Phlya:
I can't find it in the docs, is it possible for wrappers to share an environment? If multiple wrappers use the same tool, it would make sense...

Phlya:
How is it possible that the environment for linting fails when it works for all other tests?..

fgvieira:
Hi @Phlya, these changes might not be needed since, I think, there are some new GitHub actions that check for newer packages and update the environment.
Am I right @johanneskoester?

Phlya:
I don't know about that, to be honest I encountered a bug that was fixed in the latest update, so wanted to avoid it by unpinning...

Phlya:
Any more thoughts about this change?.. @fgvieira @johanneskoester

johanneskoester:
Yes, unpinning is not the right way, wrappers are updated automatically by a bot. However, the cooltools wrappers all had wrong channel orders, which perhaps prevented the bot from determining how to update them. I've fixed that in PR https://github.com/snakemake/snakemake-wrappers/pull/980 which I would like to merge instead of this one. Nevertheless, thanks a lot for pointing to the issue and the first approach here.

Phlya:
Oh I see, thank you for looking into it and fixing!

PR #984, opened 2023-01-30
debug autobump

johanneskoester:
trying to find out what problem currently occurs in the autobumping job


johanneskoester:
Can be closed because the problem has been fixed in snakedeploy itself.

PR #1149, opened 2023-03-17
feat: added vsearch wrapper

fgvieira:
### Description

Added wrapper for VSEARCH.

### QC

* [x] I confirm that:


johanneskoester (changes requested):
I think we should avoid the out_arg necessity and just infer the required args from the output file extension. And then, I would just have an extra param for additional args instead of fasta_arg and fastq_arg.

fgvieira:
Added automated output inference.

johanneskoester:
What about the fasta_arg and fastq_arg? Can't people just use extra for that?

fgvieira:
I think the way I did for `outputs` could also work for `inputs`... what do you think?

dlaehnemann (approved):
Looks good to me, very thorough work!

johanneskoester (approved):

PR #1169, opened 2023-03-27
feat: automate inference of index name

fgvieira:
### Description

Added automated inference of index name and to call `MarkDuplicatesWithMateCigar` instead of the default, since both tools are quite similar.
These changes render the wrapper `bio/picard/markduplicateswithmatecigar` redundant,

### QC

* [x] I confirm that:

### DIFF
--- a/bio/picard/markduplicates/meta.yaml
+++ b/bio/picard/markduplicates/meta.yaml
@@ -1,15 +1,19 @@
 name: picard MarkDuplicates
 description: |
-  Mark PCR and optical duplicates with picard tools. For more information about MarkDuplicates see `picard documentation <https://broadinstitute.github.io/picard/command-line-overview.html#MarkDuplicates>`_.
+  Mark PCR and optical duplicates with picard tools.
+url: https://broadinstitute.github.io/picard/command-line-overview.html#MarkDuplicates
 authors:
   - Johannes Köster
   - Christopher Schröder
+  - Filipe G. Vieira
 input:
   - bam/cram file(s)
 output:
   - bam/cram file with marked or removed duplicates
+params:
+  - `java_opts` allows for additional arguments to be passed to the java compiler, e.g. "-XX:ParallelGCThreads=10" (not for `-XmX` or `-Djava.io.tmpdir`, since they are handled automatically).
+  - `extra` allows for additional program arguments.
+  - `embed_ref` allows to embed the fasta reference into the cram
+  - `withmatecigar` allows to run `MarkDuplicatesWithMateCigar <https://broadinstitute.github.io/picard/command-line-overview.html#MarkDuplicatesWithMateCigar>`_ instead.
 notes: |
-  * The `java_opts` param allows for additional arguments to be passed to the java compiler, e.g. "-XX:ParallelGCThreads=10" (not for `-XmX` or `-Djava.io.tmpdir`, since they are handled automatically).
-  * The `extra` param allows for additional program arguments.
   * `--TMP_DIR` is automatically set by `resources.tmpdir`
-  * For more information see, https://broadinstitute.github.io/picard/command-line-overview.html#MarkDuplicates
--- a/bio/picard/markduplicates/wrapper.py
+++ b/bio/picard/markduplicates/wrapper.py
@@ -7,38 +7,41 @@ __license__ = "MIT"
 import tempfile
 from snakemake.shell import shell
 from snakemake_wrapper_utils.java import get_java_opts
+from snakemake_wrapper_utils.samtools import get_samtools_opts
+
 
 log = snakemake.log_fmt_shell()
-
 extra = snakemake.params.get("extra", "")
 # the --SORTING_COLLECTION_SIZE_RATIO default of 0.25 might
 # indicate a JVM memory overhead of that proportion
 java_opts = get_java_opts(snakemake, java_mem_overhead_factor=0.3)
+samtools_opts = get_samtools_opts(snakemake)
+
+
+tool = "MarkDuplicates"
+if snakemake.params.get("withmatecigar", False):
+    tool = "MarkDuplicatesWithMateCigar"
+
 
 bams = snakemake.input.bams
 if isinstance(bams, str):
     bams = [bams]
 bams = list(map("--INPUT {}".format, bams))
 
-if snakemake.output.bam.endswith(".cram"):
-    output = "/dev/stdout"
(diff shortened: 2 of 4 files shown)


bio/picard/markduplicates/meta.yaml
```diff
@@ -1,15 +1,19 @@
...
 authors:
   - Johannes Köster
   - Christopher Schröder
+  - Filipe G. Vieira
 input:
   - bam/cram file(s)
 output:
   - bam/cram file with marked or removed duplicates
+params:
+  - java_opts allows for additional arguments to be passed to the java compiler, e.g. "-XX:ParallelGCThreads=10" (not for `-XmX` or `-Djava.io.tmpdir`, since they are handled automatically).
```
dlaehnemann, line 14:
```suggestion
  - java_opts: allows for additional arguments to be passed to the java compiler, e.g. "-XX:ParallelGCThreads=10" (not for `-XmX` or `-Djava.io.tmpdir`, since they are handled automatically).
```

bio/picard/markduplicates/environment.yaml
```diff
@@ -3,6 +3,6 @@ channels:
   - bioconda
   - nodefaults
 dependencies:
-  - picard =2.27.4
+  - picard =3.0.0
   - samtools =1.16.1
   - snakemake-wrapper-utils =0.5.2
```
dlaehnemann, line 8: Just out now, so we avoid a further auto-bump in the future:
```suggestion
  - snakemake-wrapper-utils =0.5.3
```

dlaehnemann (changes requested):
Nice work!

Some questions/suggestions are in the code, and additionally these two:
1. Users might not want the index, and maybe we can simply leave one of the example rules without an `idx=` specified, to showcase that it is not necessary to specify it.
2. If the wrapper `MarkDuplicatesWithMateCigar` becomes obsolete with these changes, I'd simply remove it with the same pull request. Or do you see any unforeseen side-effects of this?

bio/picard/markduplicates/wrapper.py
```diff
@@ -7,38 +7,41 @@
...
+        "(picard {tool}"  # Tool and its subcommand
         " {java_opts}"  # Automatic java option
         " {extra}"  # User defined parmeters
         " {bams}"  # Input bam(s)
         " --TMP_DIR {tmpdir}"
-        " --OUTPUT {output}"  # Output bam
+        " --OUTPUT /dev/stdout"  # Output bam
         " --METRICS_FILE {snakemake.output.metrics}"  # Output metrics
-        " {convert} ) {log}"  # Logging
+        " | samtools view {samtools_opts}) {log}"  # Logging
```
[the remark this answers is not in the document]
fgvieira: I think that `embed_ref` should be dealt on the `wrapper_utils`, just like `bcftools` uncompressed `bcf`.

Also thought about it, but `picard` does not allow to define the index name, so it requires the wrapper to check if the flag was used, check if the `output.idx` is present, check what output format was requested, check if `output.idx` is different from `.bam.bai` (for BAM) or `.cram.crai` (for CRAM), and then rename. And then we still have the `.csi` indexes...

Not sure how much of a performance hit, but the code is much cleaner if we always pipe it to `samtools`.
But if you think it is a big hit, I can change it...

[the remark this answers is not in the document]
dlaehnemann: Yeah, I know the code in the wrapper will become much more convoluted, but I think it is worth doing this for any performance gain -- we regularly run this wrapper on huge data, so even a small performance gain will be helpful. I think the pattern to go for is to:
1. Check whether the output is CRAM, then you need to do the piping to samtools. If an `idx=` is specified in the output, have samtools produce it.
2. If the output is BAM or SAM, check whether an `idx=` is specified in the output, insert the `--CREATE_INDEX` flag into the command and add a command that `mv`s the default name index (from the `tmpdir`) to the file name specified in `idx=`.

Does this make sense?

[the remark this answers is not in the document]
dlaehnemann: Ah, and yes, the `--CREATE_INDEX` parsing is something that could potentially be done by a `get_picard_opts()` function in a `picard.py` file over at `snakemake-wrapper-utils`, as this is a common `picard` option. But maybe we go for implementing this here, to start with. And once this is tested, we can generalize the code over at `snakemake-wrapper-utils`.

bio/picard/markduplicates/wrapper.py
```diff
@@ -41,7 +52,13 @@
...
         " --METRICS_FILE {snakemake.output.metrics}"  # Output metrics
-        " | samtools view {samtools_opts}) {log}"  # Logging
+        " {convert}) {log}"  # Logging
     )
+
+
+output_prefix = Path(snakemake.output.bam).with_suffix("")
+if snakemake.output.get("idx"):
+    if (output_fmt == "BAM" and snakemake.output.idx != str(output_prefix) + ".bai"):
+        shell("mv {output_prefix}.bai {snakemake.output.idx}")
```
dlaehnemann: I think this `mv` will need to happen under the `with tempfile` clause above, as the index file will only (be guaranteed to) exist for the duration of that clause.

bio/picard/markduplicates/wrapper.py
```diff
@@ -5,40 +5,60 @@
...
         " --METRICS_FILE {snakemake.output.metrics}"  # Output metrics
-        " {convert} ) {log}"  # Logging
+        " {convert}) {log}"  # Logging
     )
+
+
+output_prefix = Path(snakemake.output.bam).with_suffix("")
+if snakemake.output.get("idx"):
+    if output_fmt == "BAM" and snakemake.output.idx != str(output_prefix) + ".bai":
+        shell("mv {output_prefix}.bai {snakemake.output.idx}")
```
dlaehnemann, line 64: I don't think the second check is necessary. If the `mv` doesn't do anything, that's no problem.
```suggestion
if snakemake.output.get("idx") and output_fmt == "BAM":
        shell("mv {output_prefix}.bai {snakemake.output.idx}")
```

fgvieira: I don` think you can move a file to itself. Destination needs to be different, no?

dlaehnemann: You are right. :see_no_evil: Feel free to either move all conditions into one row (with another `and`) or just merge as is. Whichever you like better.

PR #1172, opened 2023-03-28
fix: Allow concurrent and large refgenie downloads

Austin-s-h:
### Description

After running into a few instances of the refgenie wrapper downloading individual assets concurrently, I was met with a lock-timeout error when attempting to pull two large assets (I'm looking at **you** hg38_cdna/salmon_index at ~25Gb). The default timeout is 60 seconds, but I wanted to attempt to handle this error.

So, I slightly modified the refgenie wrapper to attempt to handle the `RefgenconfError` generated when the lock is unable to be obtained to skip the lock requirement. This may or may not be desirable behavior across all pipelines, but it did resolve the issues with mine as well as pass the testing requirements. 

I added a rule that mimics obtaining a large asset, but I am not familiar enough yet with the wrapper system to know if simply adding a rule means that it is tested.

Additionally, the inclusion of `force_large=True` was necessary to download assets larger than the `large` threshold of 5Gb

### QC

* [X] I confirm that:


bio/refgenie/wrapper.py
```diff
@@ -5,17 +5,27 @@
...
+    rgc = refgenconf.RefGenConf(conf_path, writable=True)
+except RefgenconfError:
+    # If read lock timeout, attempt to skip the read lock
+    rgc = refgenconf.RefGenConf(
+        conf_path, writable=True, skip_read_lock=True, genome_exact=False
+    )
 # pull asset if necessary
-gat, archive_data, server_url = rgc.pull(genome, asset, tag, force=False)
+gat, archive_data, server_url = rgc.pull(
+    genome, asset, tag, force=False, force_large=True
```
dlaehnemann, line 27: Is `force_large=True` a good general default, or would it make more sense to make this settable via the `params:` keyword in the rule definition? I am assuming their default of prompting has a reason, to avoid accidental downloads of huge reference data, and having to explicitly specify this via `params:` would at least be a minimal sanity check that the user knows what they are doing.

dlaehnemann (commented):
Thanks for looking into this!

To get your new rule tested, have a look at the existing `refgenie` test (in `test.py`):

https://github.com/snakemake/snakemake-wrappers/blob/0d2c92a5be480dd60cdb1c422282b9812aa01e2f/test.py#L4322-L4330

You would need to include another test that requires your new rule's output. However, before doing so, it might make sense to change this to an example that is only slighlty over the threshold of 5GB, because otherwise the test would do an excessive download (for example the 25GB you mentioned) and fill up the GitHub Action VM's disk space every time it is run (even though it is only run if anything changes in this wrapper).

bio/refgenie/wrapper.py
Austin-s-h: I think that is a good alternative to implement. As is, there is no way to override this while using the wrapper.

bio/refgenie/wrapper.py
```diff
@@ -5,17 +5,27 @@
...
-rgc = refgenconf.RefGenConf(conf_path, writable=True)
-
+# BUG If there are multiple concurrent refgenie commands, this will fail due to
+# unable to acquire lock of the config file.
+try:
+    rgc = refgenconf.RefGenConf(conf_path, writable=True)
+except RefgenconfError:
+    # If read lock timeout, attempt to skip the read lock
+    rgc = refgenconf.RefGenConf(
+        conf_path, writable=True, skip_read_lock=True, genome_exact=False
```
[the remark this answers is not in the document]
Austin-s-h: I didn't attempt to, but I suspect that this might not be a great choice either. If someone is downloading an asset over a slow connection, even setting wait_max from its default of 60 to 600 might not make a difference and result in a hard-to-diagnose timeout error.

I'm not sure if this was some sort of conflict with the snakemake locking system as well. If we rely on that to protect other files, then the result of the wrapper is it either produces the output file, or the rule fails with a RefgenconfError error and recommends setting the skip_read_lock=TRUE param to try to fix the issue.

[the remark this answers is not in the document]
dlaehnemann: From what I gathered by poking around a little, I think that the lock only happens while something is written to the conf file. So I would think that this lock is not in place the whole time you are doing the download and that the `wait_max=` should already help. But the documentation on this is not very clear and I didn't immediately find the mechanism in the code, so I might be misunderstanding this lock.

Do you have the possibility to try `wait_max=` in your use case and test whether this actually helps?

bio/refgenie/wrapper.py
dlaehnemann: Do you feel comfortable implementing this?

I'd introduce an (optional) `params: force_large=True` in the (one of the) examples, and parse this here in the `wrapper.py` with `force_large=snakemake.params.get("force_large", None)`, so defaulting to what the default in the original function is, only changing it if this is a deliberate choice by the user.

fgvieira:
@Austin-s-h can you fix the conflicts?

PR #1173, opened 2023-03-28
feat: also define overhang on params.extra

fgvieira:
### QC

* [x] I confirm that:

### DIFF
--- a/bio/star/index/wrapper.py
+++ b/bio/star/index/wrapper.py
@@ -10,18 +10,16 @@ from snakemake.shell import shell
 from snakemake.utils import makedirs
 
 log = snakemake.log_fmt_shell(stdout=True, stderr=True)
-
 extra = snakemake.params.get("extra", "")
-sjdb_overhang = snakemake.params.get("sjdbOverhang", "100")
 
-gtf = snakemake.input.get("gtf")
-if gtf is not None:
-    gtf = f"--sjdbGTFfile {gtf}"
+sjdb_overhang = snakemake.params.get("sjdbOverhang", "")
+if sjdb_overhang:
     sjdb_overhang = f"--sjdbOverhang {sjdb_overhang}"
-else:
-    gtf = sjdb_overhang = ""
 
-makedirs(snakemake.output)
+gtf = snakemake.input.get("gtf", "")
+if gtf:
+    gtf = f"--sjdbGTFfile {gtf}"
+
 
 with tempfile.TemporaryDirectory() as tmpdir:
     shell(
--- a/meta/bio/star_arriba/test/Snakefile
+++ b/meta/bio/star_arriba/test/Snakefile
@@ -1,12 +1,13 @@
 rule star_index:
     input:
         fasta="resources/genome.fasta",
-        annotation="resources/genome.gtf",
+        gtf="resources/genome.gtf",
     output:
         directory("resources/star_genome"),
     threads: 4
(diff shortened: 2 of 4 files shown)


bio/star/index/wrapper.py
```diff
@@ -10,18 +10,16 @@
...
-
 extra = snakemake.params.get("extra", "")
-sjdb_overhang = snakemake.params.get("sjdbOverhang", "100")
 
-gtf = snakemake.input.get("gtf")
-if gtf is not None:
-    gtf = f"--sjdbGTFfile {gtf}"
+sjdb_overhang = snakemake.params.get("sjdbOverhang", "")
+if sjdb_overhang:
     sjdb_overhang = f"--sjdbOverhang {sjdb_overhang}"
```
dlaehnemann, line 17: I know there was code in this wrapper for parsing `snakemake.params.sjdbOverhang` already, but does anything speak against simply removing this special handling? This option can easily be set in the standard `params.extra:` entry, with `--sdbOverhang 160`, for example.

dlaehnemann (approved):
This generally looks good to me, one question/suggestion.

bio/star/index/wrapper.py
fgvieira: Don't really see the point of it, but just trying to keep backwards compatibility.

dlaehnemann: Ah, yes, you're right. Good point. So maybe just leave it as is, for now.

meta/bio/star_arriba/test/Snakefile
```diff
@@ -1,12 +1,13 @@
...
         fasta="resources/genome.fasta",
-        annotation="resources/genome.gtf",
+        gtf="resources/genome.gtf",
     output:
         directory("resources/star_genome"),
     threads: 4
     params:
-        extra=lambda wc, input: f"--sjdbGTFfile {input.annotation} --sjdbOverhang 100",
+        sjdbOverhang=100,
+        extra="--genomeSAindexNbases 2",
```
dlaehnemann, line 10: @fgvieira Any idea why you added this here? Because this implicitly ends up being copy-pasted when people use the meta-wrapper, and I have just found it in one of our major workflows and think this might actually mess up the index creation. Here's what the STAR docs say on this:
> 2.2.5 Very small genome.
For small genomes, the parameter `--genomeSAindexNbases` must to be scaled down, with a typical
value of `min(14, log2(GenomeLength)/2 - 1)`. For example, for 1 megaBase genome, this is equal
to 9, for 100 kiloBase genome, this is equal to 7.

And here's what the `STAR --help` message reads:
> genomeSAindexNbases         14
    int: length (bases) of the SA pre-indexing string. Typically between 10 and 15. Longer strings will use much more memory, but allow faster searches. For small genomes, the parameter --genomeSAindexNbases must be scaled down to min(14, log2(GenomeLength)/2 - 1).

So I would just opt for not setting this in the meta-wrapper, at all.

fgvieira: That was just to have an example `params.extra` but, if you think it is confusing, feel free to remove it (or replace it by something else).

dlaehnemann: Thanks for the background. Then I'll just remove it, see here:
https://github.com/snakemake/snakemake-wrappers/pull/4904

PR #1209, opened 2023-04-05
feat: add wrapper for rbt vcf_fix_iupac_alleles

fgvieira:
### QC

* [x] I confirm that:

## Summary by CodeRabbit

* **New Features**
  * Introduced a new tool for converting IUPAC codes in VCF files to Ns, ensuring compatibility with VCF version 4 specifications.
  * Added support for processing both VCF and BCF formats via Snakemake rules.
  * Provided a dedicated environment configuration for reproducibility.

* **Tests**
  * Added automated tests to verify correct processing of VCF and BCF files with the new tool.


bio/rbt/vcf_fix_iupac_alleles/test/Snakefile
```diff
@@ -0,0 +1,16 @@
+
+rule vcf_fix_iupac_alleles:
+    input:
+        "{file}.vcf.gz",
+    output:
+        vcf="results/{file}.vcf.gz",
+        idx="results/{file}.vcf.gz.tbi",
```
johanneskoester, line 7: I think the index here is a bit of overkill, isn't it? As we already have a tabix wrapper. Also, consider mentioning the other possible input and output formats (bcf, vcf,)

johanneskoester (changes requested):

bio/rbt/vcf_fix_iupac_alleles/wrapper.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+bcftools_opts = get_bcftools_opts(snakemake, parse_ref=False)
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+extra = snakemake.params.get("extra", "")
+
+
+with tempfile.TemporaryDirectory() as tmpdir:
+    shell(
+        "(rbt vcf-fix-iupac-alleles {extra} < {snakemake.input[0]} | bcftools sort --temp-dir {tmpdir} {bcftools_opts}) {log}"
```
johanneskoester, line 17: Why the sort? rbt vcf-fix-iupac-alleles does not change the order of the records.

fgvieira: The main idea was to allow for automatic output on any format (`bcf`, `vcf`, `vcf.gz`).
Since I was using it to standardize resources (known variation from ensembl), I thought that `sort` would be a nice-to-have (just to be sure the output was sorted) and the extra runtime would be ok since it would only be run once.

But I can also remove it (or switch to `view`), if you prefer to keep the wrapper more general.

fgvieira: Ping @johanneskoester

Issue #1238, opened 2023-04-07
MACS2: support --extsize with fragment size input file


ftabaro:
Hello, 

I am developing a pipeline for ChIP-seq. I have to support single-end reads. I need to estimate fragment size before doing peak calling. Currently, I am doing a autocorrelation analysis writing the estimated value to a txt file upstream of MACS2.

I would like to use MACS2 callpeak  with `--nomodel --extsize $FRAGSIZE` where `FRAGSIZE=$(cat {input.fragsize}`. I tried to implement it with the shell directive but there I cannot have optional input files. Therefore, I thought to add an optional `fragment_size` input file to the MACS2 callpeak wrapper.  This optional input file should then add `--extsize {fragsize}` to the MACS2 callpeak command, in a similar fashion to the optional `control` input file.

PR #1239, opened 2023-04-07
feat: implement fragment size as optional input file

ftabaro:
### Description

Addresses #1238 

### QC

* [x ] I confirm that:


fgvieira:
It seems `extsize` (or `fragsize`) is a single integer, no? If so, I feel if would make more sense to be a `param`.
Would it make sense to define it as a param and define a function that reads the input file? For example, define a function that returns `extsize`:
```
def get_frag_size(in_fragsize):
    with open(in_fragsize) as handler:
        return handler.readline().strip()
```
and call it in `params` as:
```
rule foo:
[...]
params:
    extsize = lambda w, input: get_frag_size(input.fragsize)
```
or even:
```
rule foo:
[...]
params:
    extra = lambda w, input: "--extsize {}".format(get_frag_size(input.fragsize))
```

ftabaro:
Hi, thanks! Yes that would make sense. But how to make the callpeak rule depend on upstream rule(s) that compute this integer? For instance, using the `csaw` R package, one could print the value to stdout. However, currently we save it to a file and require such file as input for downstream rules. Will reading the file in a functions within  params make it required for the target rule (here macs2 callpeak)?

fgvieira:
> Will reading the file in a functions within params make it required for the target rule (here macs2 callpeak)?

No, but you can add the file to the rule's input function.

You can keep the rules as you have now: first rule estimates parameter and writes into file, while second rule takes output of previous rule as input and calls peaks:
```
def get_frag_size(in_fragsize):
    with open(in_fragsize) as handler:
        return handler.readline().strip()

rule csaw:
    input:
        [...]
    output:
        "csaw.out"

rule call_peak:
    input:
        fragsize = "csaw.out"
    output:
        "peaks.out"
    params:
        extra = lambda w, input: "--extsize {}".format(get_frag_size(input.fragsize))
```

ftabaro:
Do you think it would make sense to have this implemented in the wrapper?

Issue #1074, opened 2023-02-18
RSEM Calculate Expression bug for paired-end bam file input


isaacvock:
In the documentation for the [rsem-calculate-expression wrapper](https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/rsem/calculate-expression.html), the params block includes `paired_end=True` as an example of what needs to be included if passing a bam file from aligned paired end reads as input. In the actual wrapper script though, this parameter is fetched as `paired_end=snakemake.params.get("paired-end", False)`. This is both a problem because it clashes with the parameter name used in the documentation, but also I believe parameter names can't have hyphens, so changing the documentation to `paired-end=True` is probably fraught. Thus, I think the wrapper code needs to be modified to `paired_end=snakemake.params.get("paired_end", False).

fgvieira:
Hi @isaacvock, thanks for reporting the bug. Do you think you could do a PR?

isaacvock:
Will do!

fgvieira:
thanks!

isaacvock:
I got it to run with bam files from paired-end data aligned with STAR in a separate, upstream rule. I have not tried with fastqs as input.

It is probably worth including more details about the errors you are getting and/or problems that you are running into with this particular use-case. In the meantime, I'll try and run the wrapper with paired-end fastqs as input and see if I run into any similar problems.

KuechlerO:
Hi @isaacvock,
thx for the quick answer, I see.

Kay, actually I was able to make it run.
However, when using the wrapper one needs to take into consideration:
When fastq-files are used as input, one needs to provide a bowtie-index. I suspect that in that case the creation of the RSEM-index with the "RSEM prepare reference"-wrapper is not necessary.

But I need to take a closer look into that.
Maybe the wrapper examples could be split into 2 cases then: One using pre-aligned results, the other one not.

PR #1289, opened 2023-04-23
refactor: Improving RSEM wrappers' code and documenation

KuechlerO:
### Description

This PR includes:
- RSEM-calculate-expression: 
    - Integrate separate examples for aligned and non-aligned inputs
- RSEM-prepare-reference:
    -  Renamed the variable `reference_genome` to `reference_fasta` to make it more clear that either transcriptomic or genomic sequences can be passed
    - Improved documentation in `meta.yaml`
- test.py
    - Renamed `test_generate_data_matrix` to `test_rsem_generate_data_matrix`, so that a clear association with the RSEM wrapper is given.

### DIFF
--- a/bio/umis/bamtag/environment.yaml
+++ b/bio/umis/bamtag/environment.yaml
@@ -4,4 +4,4 @@ channels:
   - nodefaults
 dependencies:
   - umis =1.0.9
-  - samtools =1.17
+  - samtools =1.18


KuechlerO:
Problems have been mentioned [in issue 1074](https://github.com/snakemake/snakemake-wrappers/issues/1074).

bio/rsem/prepare-reference/meta.yaml
```diff
@@ -4,9 +4,10 @@ description: |
...
-  - reference genome
-  - additional optional arguments
+  - reference_fasta: reference FASTA with either the entire genome or transcript sequences. 
+      If the entire genome is used, the --gtf option must be specified under params.
+  - extra: additional optional arguments, e.g. --gtf
 output:
-  - index files for downstream use with rsem
+  - index files for downstream usage with RSEM
 notes: |
   * For more information, see https://github.com/deweylab/RSEM.
```
fgvieira, line 13:
```suggestion
```

KuechlerO:
@fgvieira Are not all rules that are included in the Snakefile also executed when the Snakefile is passed to the test?
Also `test_rsem_calculate_expression()` in test.py includes a separate Snakefile for testing the wrapper with non-aligned inputs (I pushed it last year already...). Having both, one test in a separate Snakefile and the two versions in the main Snakefile makes on redundant though.
So I would suggest to delete the additional test case in the separate Snakefi.e.

fgvieira:
You need to specify a `target` (either a `rule` or an output file) for the tests to run.

I agree, I also think it is better to have a single `Snakefile` per wrapper with all the tests.

KuechlerO:
Ok, I see. I followed your advices and modified the Snakefile and tests accordingly :)

bio/rsem/prepare-reference/meta.yaml
```diff
@@ -1,12 +1,12 @@
...
         Run rsem-prepare-reference to create index files for downstream analysis with rsem.
+url: https://github.com/deweylab/RSEM
 authors:
   - Brett Copeland
 input:
-  - reference genome
-  - additional optional arguments
+  - ref: reference FASTA with either the entire genome or transcript sequences. 
+      If the entire genome is used, the --gtf option must be specified under params.
+  - extra: additional optional arguments, e.g. --gtf
```
fgvieira, line 10: This should be after `output`, under `params`, no?

KuechlerO: True, I was not sure, whether a certain structure is required.
Now I have updated the meta.yaml file accordingly.

bio/rsem/prepare-reference/meta.yaml
```diff
@@ -1,12 +1,13 @@
...
 input:
-  - reference genome
-  - additional optional arguments
+  - ref: reference FASTA with either the entire genome or transcript sequences. 
+      If the entire genome is used, the --gtf option must be specified under params.
 output:
-  - index files for downstream use with rsem
-notes: |
-  * For more information, see https://github.com/deweylab/RSEM.
+  - index files for downstream usage with RSEM
```
fgvieira, line 11: The name of the output file has to be `seq`, right?
```suggestion
  - seq: index files for downstream usage with RSEM
```

bio/rsem/prepare-reference/meta.yaml
```diff
@@ -1,12 +1,13 @@
...
+params:
+  - extra: additional optional arguments, e.g. "--gtf <path-to-gtf>" or "--transcript-to-gene-map <path-to-map>"
```
fgvieira, line 13: It sounds like the `--gtf` and `--transcript-to-gene-map` are input files, no?
If so, could they be automatically inferred from the rule `input`?

bio/rsem/prepare-reference/meta.yaml
KuechlerO: I am not one of the RSEM developers, but in the example code (Snakefile) they had just written, that 
```
# one of the index files created and used by RSEM (required)
seq="index/reference.seq",
```

So, in general all the listed output files are part of the file set that is produced for indexing purpose.

bio/rsem/prepare-reference/meta.yaml
KuechlerO: Mhm, but these arguments are completely optional.
I am not sure what the common practice for setting up Snakemake wrappers in such cases is... So you would suggest to have every possible additional argument with its corresponding file to be listed in the input?

fgvieira: The ideal scenario, yes, all optional inputs should be properly dealt by the wrapper so that the user only has to specify them as `input:` (see examples [here](https://github.com/snakemake/snakemake-wrappers/blob/master/bio/gatk/denoisereadcounts/wrapper.py#L12-L19) or [here](https://github.com/snakemake/snakemake-wrappers/blob/master/bio/deeptools/alignmentsieve/wrapper.py#L12-L18)).

It is not that complex (e.g.):
```
gtf = snakemake.input.get("gtf", ""):
if gtf:
    gtf = f"--gtf {gtf}"

transc2gene_map = snakemake.input.get("transc2gene_map", ""):
if transc2gene_map:
    gc_intervals = f"--transcript-to-gene-map {transc2gene_map}"
```

and then add `gtf` and `transc2gene_map` to the command.

PR #1291, opened 2023-04-25
feat: Gffread

tdayris:
### Description

This PR adds [gffread](https://github.com/gpertea/gffread) to the list of available wrappers.

### QC

* [X] I confirm that:


bio/gffread/environment.yaml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - bioconda
+  - conda-forge
```
fgvieira, line 3:
```suggestion
  - conda-forge
  - bioconda
```

bio/gffread/test/Snakefile
```diff
@@ -0,0 +1,20 @@
+rule test_gffread:
+    input:
+        fasta="genome.fasta",
+        annotation="annotation.gtf",
+        # ids="",  # Optional path to records to kkep
```
fgvieira, line 5:
```suggestion
        # ids="",  # Optional path to records to keep
```

bio/gffread/wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+__license__ = "MIT"
+
+
+from snakemake.shell import shell
+
+extra = snakemake.params.get("extra", "")
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+
+annotation = snakemake.input["annotation"]
+records = snakemake.output["records"]
```
fgvieira, line 13:
```suggestion
annotation = snakemake.input.annotation
records = snakemake.output.records
```

bio/gffread/environment.yaml
```diff
@@ -0,0 +1,6 @@
...
+  - nodefaults
+dependencies:
+  - gffread
```
fgvieira, line 6:
```suggestion
  - gffread =0.12.7
```

bio/gffread/wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+    extra += " --in-tlf "
+
+
+# Output format control
+if records.endswith(".gtf") and (" -T " not in extra):
+    extra += " -T "
+elif (records.endswith(".bed")) and (" --bed " not in extra):
+    extra += " --bed "
+elif (records.endswith(".tlf")) and (" --tlf " not in extra):
+    extra += " --tlf "
```
fgvieira, line 28: Are the outputs mutually exclusive, or can it output (e.g.) both a `gtf` and a `bed`?

bio/gffread/wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+from snakemake.shell import shell
+
+extra = snakemake.params.get("extra", "")
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+
+annotation = snakemake.input["annotation"]
+records = snakemake.output["records"]
+
+# Input format control
+if (annotation.endswith(".bed")) and (" --in-bed " not in extra):
```
fgvieira, line 16: The check `" --in-bed " not in extra` won't work if `extra` starts with `--in-bed` (`extra = "--in-bed some/file.bed"`).

Is this check necessary? You can just state in the `meta.yaml` that those options are automatically inferred.

bio/gffread/wrapper.py
tdayris: There is only one output file, the format may be changed with these flags

bio/gffread/test/Snakefile
```diff
@@ -0,0 +1,20 @@
+rule test_gffread:
+    input:
+        fasta="genome.fasta",
+        annotation="annotation.gtf",
+        # ids="",  # Optional path to records to keep
+        # nids="",  # Optional path to records to drop
+        # seq_info="",  # Optional path to sequence information
+        # sort_by="",  # Optional path to the ordered list of reference sequences
+        # attr="",  # Optional annotation attributes to keep.
+        # chr_replace="",  # Optional path to <original_ref_ID> <new_ref_ID>
```
fgvieira, line 10: Not sure if it is feasible but, since there seems to be some many extra inputs, it would be nice to also have tests for some of them. Is it too much of a hassle?

johanneskoester (approved):

PR #1297, opened 2023-04-27
feat: remove existing run directory option for strelka

christopher-schroeder:
### Description
When strelka reruns it doesn't like if the run directory exists. Add an option to delete it before.

### QC

* [x] I confirm that:


fgvieira:
Wouldn't it be better to output to a temp folder and then copy relevant results to their final destination?
For example, like it was done for [STAR](https://github.com/snakemake/snakemake-wrappers/blob/master/bio/star/align/wrapper.py).

christopher-schroeder:
> Wouldn't it be better to output to a temp folder and then copy relevant results to their final destination? For example, like it was done for [STAR](https://github.com/snakemake/snakemake-wrappers/blob/master/bio/star/align/wrapper.py).

Yes, ok - why not.

bio/strelka/germline/test/Snakefile
```diff
@@ -7,13 +7,13 @@ rule strelka_germline:
         fasta_index="genome.fasta.fai"
     output:
         # Strelka results - either use directory or complete file path
-        directory("strelka/{sample}")
+        variants="strelka/{sample}.vcf.gz"
```
fgvieira, line 10: Maybe add explicit names, in case in the future someone wants to implement automatic conversion to other formats (e.g. `bcf`).
```suggestion
        vcf="strelka/{sample}.vcf.gz"
        tbi="strelka/{sample}.vcf.gz.tbi"
```

christopher-schroeder: I thought about that but wouldn't it be better to call it variants and detect the ending of the filename to automatically convert?

fgvieira: I (personally) find it easier (and more flexible) to check the format by the name of the output, rather than the actual extension of the file.

christopher-schroeder: But this is the normal behaviour of most tools. An example

bcftools view -o test.bcf
bcftools view -o test.vcf
bcftools view -o test.vcf.gz

I very must prefer this over the potential

bcftools view -obcf test.bcf
bcftools view -ovcf test.vcf
bcftools view -ovcfgz test.vcf.gz

johanneskoester: I agree with @christopher-schroeder , the wrapper should react on the extension, just like we do it with other wrappers as well.

johanneskoester (changes requested):

bio/strelka/germline/test/Snakefile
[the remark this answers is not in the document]
christopher-schroeder: yes, the wrapper is not done yet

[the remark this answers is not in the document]
johanneskoester: I unresolved this just so that I don't forget the request until it is done.

johanneskoester (approved):

PR #1321, opened 2023-05-02
feat: gatk modelsegments wrapper

Smeds:
### Description

gatk ModelSegments wrapper

### QC

* [ ] I confirm that:

### DIFF
--- a/bio/gatk/modelsegments/meta.yaml
+++ b/bio/gatk/modelsegments/meta.yaml
@@ -10,7 +10,6 @@ input:
   - matched_normal allelic-counts, Optional
   - segments picard interval-list file containing a multisample segmentation output by a previous run, optional
 params:
-  - prefix for output filenames, if not specified the wrapper will try to extract it from the output files
   - The `java_opts` param allows for additional arguments to be passed to the java compiler, e.g. "-XX:ParallelGCThreads=10" (not for `-XmX` or `-Djava.io.tmpdir`, since they are handled automatically).
   - The `extra` param allows for additional program arguments.
 output:
--- a/bio/gatk/modelsegments/meta.yaml
+++ b/bio/gatk/modelsegments/meta.yaml
@@ -5,15 +5,15 @@ description: |
 authors:
   - Patrik Smeds
 input:
-  - denoised_copy_ratios file, optional 
-  - allelic_counts file, Optional
-  - matched_normal allelic-counts, Optional
-  - segments picard interval-list file containing a multisample segmentation output by a previous run, optional
-params:
-  - The `java_opts` param allows for additional arguments to be passed to the java compiler, e.g. "-XX:ParallelGCThreads=10" (not for `-XmX` or `-Djava.io.tmpdir`, since they are handled automatically).
-  - The `extra` param allows for additional program arguments.
+  - denoised_copy_ratios: denoised_copy_ratios file (optional)
+  - allelic_counts: allelic_counts file (optional)
+  - normal_allelic_counts: matched_normal allelic-counts (optional)
+  - segments: segments Picard interval-list file containing a multisample segmentation output by a previous run (optional)
 output:
-  - list of files, one ore more files ending with '.modelFinal.seq', '.cr.seg', '.af.igv.seg', '.cr.igv.seg', '.hets.tsv', '.modelBegin.cr.param', '.modelBegin.af.param', '.modelBegin.seg', '.modelFinal.af.param' and '.modelFinal.cr.param'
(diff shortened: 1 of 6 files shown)


bio/gatk/modelsegments/wrapper.py
```diff
@@ -0,0 +1,105 @@
...
+            if output_file.endswith(output_extensions):
+                output_prefix = output_file.replace(output_extensions, "")
+                break
+        if output_prefix:
+            break
+    if not output_prefix:
+        raise Exception(
+            "Unable to extract prefix from listed files, expecting file(s) ending "
+            f"with at least one of {expected_output_file_endings}"
+        )
```
fgvieira, line 79: Not sure if this would do the same:
```
prefix = os.path.commonprefix(output)
output_prefix = os.path.basename(prefix)
```

It only works with, at least, 2 files; and does not check the extensions, but this would be done by the rule when it finishes.

Smeds: Your solution is much nicer. But it could be that the user only wants one output. I could handle that by making a check for at least 2 files and if only one is listed throw an exception saying that prefix needs to be specified when only one file is request.

fgvieira: It was really just a suggestion to make the code more readable, so completely up to you! :smile: 

But looking into it again, maybe the files don't even need to have the same prefix. Since you are copying them from the `temp` folder at the end, you can use any prefix you want, no?

You just need to have a dict with the correspondence between output names and extensions.

Smeds: I have updated the code doing the copy. The user can now:
- have different prefix for each file, generated output will be matched with listed output
- the wrapper will not assume that specif filed are generate, i.e gatk can update output from the function and this should not break the wrapper
- prefix has been removed from params since the output list will set name of files

bio/gatk/modelsegments/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+input:
+  - denoised_copy_ratios file, optional 
+  - allelic_counts file, Optional
+  - matched_normal allelic-counts, Optional
+  - segments picard interval-list file containing a multisample segmentation output by a previous run, optional
+params:
+  - The `java_opts` param allows for additional arguments to be passed to the java compiler, e.g. "-XX:ParallelGCThreads=10" (not for `-XmX` or `-Djava.io.tmpdir`, since they are handled automatically).
+  - The `extra` param allows for additional program arguments.
+output:
+  - list of files, one ore more files ending with '.modelFinal.seq', '.cr.seg', '.af.igv.seg', '.cr.igv.seg', '.hets.tsv', '.modelBegin.cr.param', '.modelBegin.af.param', '.modelBegin.seg', '.modelFinal.af.param' and '.modelFinal.cr.param'
```
fgvieira, line 16:
```suggestion
  - denoised_copy_ratios: denoised_copy_ratios file (optional)
  - allelic_counts: allelic_counts file (optional)
  - normal_allelic_counts: matched_normal allelic-counts (optional)
  - segments: segments Picard interval-list file containing a multisample segmentation output by a previous run (optional)
output:
  - list of files ending with either '.modelFinal.seq', '.cr.seg', '.af.igv.seg', '.cr.igv.seg', '.hets.tsv', '.modelBegin.cr.param', '.modelBegin.af.param', '.modelBegin.seg', '.modelFinal.af.param' or '.modelFinal.cr.param'
params:
  - java_opts: additional arguments to be passed to the java compiler, e.g. "-XX:ParallelGCThreads=10" (not for `-XmX` or `-Djava.io.tmpdir`, since they are handled automatically).
  - extra: additional program arguments.
```

bio/gatk/modelsegments/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - normal_allelic_counts: matched_normal allelic-counts (optional)
+  - segments: segments Picard interval-list file containing a multisample segmentation output by a previous run (optional)
+output:
+  - list of files ending with either '.modelFinal.seq', '.cr.seg', '.af.igv.seg', '.cr.igv.seg', '.hets.tsv', '.modelBegin.cr.param', '.modelBegin.af.param', '.modelBegin.seg', '.modelFinal.af.param' or '.modelFinal.cr.param'
+params:
+  - java_opts: additional arguments to be passed to the java compiler, e.g. "-XX:ParallelGCThreads=10" (not for `-XmX` or `-Djava.io.tmpdir`, since they are handled automatically).
+  - extra: additional program arguments.
+note: |
+  * Expected input is either denoised_copy_ratios or allelic_counts
+  * matched_normal must provided together with allelic_counts
```
fgvieira, line 19:
```suggestion
  * Expected input is either `denoised_copy_ratios` or `allelic_counts`
  * `normal_allelic_counts` must be provided together with `allelic_counts`

```

fgvieira (approved):

Issue #885, opened 2022-11-17
Error in snakemake wrapper goes not to log file


SilasK:
**Snakemake version**
Note the Snakemake **and** snakemake-wrappers version for which you experience the bug.
Always make sure to try out the latest version of both before creating a new bug report.

7.15

**Describe the bug**

Error in snakemake wrapper goes not to log file

**Logs**
```
rule mapping_stats_genomes:
    input: genomes/alignments/bams/sample1.bam
    output: genomes/alignments/stats/sample1.stats
    log: logs/genomes/alignments/sample1_stats.log
    jobid: 0
    reason: Forced execution
    wildcards: sample=sample1
    resources: mem_mb=10000, disk_mb=1000, tmpdir=/tmp, mem=10, time=1, time_min=60

python -c "import sys; print('.'.join(map(str, sys.version_info[:2])))"
Activating conda environment: ../../../../gpfs/nihs/microbiome/Software/anaconda/Anaconda3-ATLAS/ATLAS_DATABASES/conda_envs/559a96fc72675fbc18f6663be49b4d36_
python /scratch/rdkiesersi1/Debug_atlas/WD/.snakemake/scripts/tmphrhrjlo3.wrapper.py
Activating conda environment: ../../../../gpfs/nihs/microbiome/Software/anaconda/Anaconda3-ATLAS/ATLAS_DATABASES/conda_envs/559a96fc72675fbc18f6663be49b4d36_
Traceback (most recent call last):
  File "/scratch/rdkiesersi1/Debug_atlas/WD/.snakemake/scripts/tmphrhrjlo3.wrapper.py", line 31, in <module>
    shell(
  File "/gpfs/home/rdkiesersi1/miniconda3/envs/atlas-dev/lib/python3.10/site-packages/snakemake/shell.py", line 139, in __new__
    cmd = format(cmd, *args, stepout=2, **kwargs)
  File "/gpfs/home/rdkiesersi1/miniconda3/envs/atlas-dev/lib/python3.10/site-packages/snakemake/utils.py", line 431, in format
    return fmt.format(_pattern, *args, **variables)
  File "/gpfs/nihs/microbiome/Software/anaconda/Anaconda3-ATLAS/ATLAS_DATABASES/conda_envs/559a96fc72675fbc18f6663be49b4d36_/lib/python3.10/string.py", line 161, in format
    return self.vformat(format_string, args, kwargs)
[...]

PR #1322, opened 2023-05-02
feat!: replace all bbtool wrappers with one bbtool wrapper for all subcommands

SilasK:
### Description

I add a generic bbtool wrapper. 
It is based on #2 

BBtools hase many bash shcripts. This wrapper should allow to run them all. 

### QC

* [x] I confirm that:


bio/bbtools/generic/wrapper.py
```diff
@@ -0,0 +1,275 @@
...
+    logfile = get_stderr_file(snakemake)
+
+    # clear stderr file
+    open(logfile, "w").close()
+
+except:
+    print("No log file provided, logging to stdout")
+    logfile = None
+
+logging.basicConfig(
```
SilasK, line 83: That's my way to enshure that all errors in the wrapper are written to the log file

bio/bbtools/generic/wrapper.py
```diff
@@ -0,0 +1,275 @@
...
+    )
+
+
+# Install exception handler
+sys.excepthook = handle_exception
+
+logger = logging.getLogger(__file__)
+
+##############################################################
+## helper functions they should go in snakemake_wrapper_utils
```
SilasK, line 112: helper functions they should go in snakemake_wrapper_utils

bio/bbtools/generic/wrapper.py
SilasK: #885

SilasK:
Can I ask @fgvieira  and or @dlaehnemann for a review. 

@fgvieira You did the other bbtools snakemake wrappers, isn't it?

bio/bbtools/generic/wrapper.py
SilasK: @dlaehnemann Do I need to do this first?

bio/bbtools/generic/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+input:
+  - input: All keywords in input are passed to the bbtool as key=value pairs. A special case is 'input' see in the notes. 
+  - flag: The keyword 'flag' is ignored and allows to make input files that are not processed by the bbtool.
+output:
+  - out: All keywords in output are passed to the bbtool as key=value pairs. A special case is 'out' see in the notes. 
+params:
+  - command: Required parameter defining the command to be used e.g. 'bbmap.sh'
+  - extra: additional program arguments
+  - other: All other parameters are passed to the bbtool as key=value pairs
+  - define_threads: By default the the threads argument of snakemake is passed to the bbtool. For bbtools that do not support this argument, set this to False.
```
fgvieira, line 16: Instead of this `param`, what if there was a list of all the BBduk tools that support threads in the wrapper?

bio/bbtools/generic/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - Silas Kieser
+input:
+  - input: All keywords in input are passed to the bbtool as key=value pairs. A special case is 'input' see in the notes. 
+  - flag: The keyword 'flag' is ignored and allows to make input files that are not processed by the bbtool.
+output:
+  - out: All keywords in output are passed to the bbtool as key=value pairs. A special case is 'out' see in the notes. 
+params:
+  - command: Required parameter defining the command to be used e.g. 'bbmap.sh'
+  - extra: additional program arguments
+  - other: All other parameters are passed to the bbtool as key=value pairs
```
fgvieira, line 15: It seems this `param` is not used.

fgvieira:
Nice PR! Can you also add tests for the other rules on the `test/Snakefile`?

And maybe even add the tests from the other `bbduk` wrappers (`bbmap`, `tadpole`, and `loglog`) since, if this wrapper works with all, I'd even say to delete the other wrappers and have this one as the single wrapper for `bbduk`.

bio/bbtools/generic/environment.yaml
```diff
@@ -0,0 +1,9 @@
+channels:
+  - conda-forge
+  - bioconda
+  - nodefaults
+dependencies:
+  - bbmap =39.01
+  - snakemake-wrapper-utils =0.5
+  - python =3.11
```
fgvieira, line 8: Is python really necessary?

fgvieira: I tested the other wrappers (and made #1366), and they seem to run fine without it.

johanneskoester: Yes, if there is no python in the env, Snakemake uses the python from its main process.

johanneskoester (changes requested):
I also think the other bbtool wrappers should be deleted in order to avoid duplication. One should probably just call the wrapper bbtools, without the "generic" then.

fgvieira:
And all add tests for other bbtool tools, e.g. `clumpify`.

bio/bbtools/generic/wrapper.py
SilasK: Finally now as this will be the only wrapper that uses this

SilasK (commented):
I moved them

SilasK (commented):
I moved the wrapper to the base dir `bio/bbtools` and removed the others. I combined all the tests

bio/bbtools/environment.yaml
```diff
@@ -4,4 +4,5 @@ channels:
   - nodefaults
 dependencies:
   - bbmap =39.01
-  - snakemake-wrapper-utils =0.5.3
+  - snakemake-wrapper-utils =0.5
```
fgvieira, line 7:
```suggestion
  - snakemake-wrapper-utils =0.6.2
```

bio/bbtools/meta.yaml
```diff
@@ -0,0 +1,133 @@
...
+input:
+  - input: All keywords in input are passed to the bbtool as key=value pairs. A special case is 'input' see in the notes.
+  - flag: The keyword 'flag' is ignored and allows to make input files that are not processed by the bbtool. All other input are passed to the bbtool as key=value pairs
+output:
+  - out: All keywords in output are passed to the bbtool as key=value pairs. A special case is 'out' see in the notes. All other outputs are passed to the bbtool as key=value pairs
+params:
+  - command: Required parameter defining the command to be used e.g. 'bbmap.sh'
+  - extra: additional program arguments. All other parameters are passed to the bbtool as key=value pairs
+notes: |
+  This wrapper allows to run any of the bbtools. The command is defined by the 'command' parameter, which required.
```
fgvieira, line 16:
```suggestion
  This wrapper allows to run any of the bbtools. The command is defined by the 'command' parameter, which is required.
```

bio/bbtools/meta.yaml
```diff
@@ -0,0 +1,133 @@
+name: Generic
+url: https://jgi.doe.gov/data-and-tools/bbtools/bb-tools-user-guide
```
fgvieira, line 2:
```suggestion
url: https://jgi.doe.gov/data-and-tools/software-tools/bbtools/bb-tools-user-guide/
```

bio/bbtools/meta.yaml
```diff
@@ -0,0 +1,133 @@
...
+  **All keywords in input, output, params are passed as key value pairs to the command!**
+  Take care that they are valid for the bbtool.
+
+  As it it not possible to define 'in' as a keyword, the keyword 'input' is used instead. Allowed aliases are 'sample' and 'reads', but we recommend to use 'input'.
+  If exactly two files are provides for 'input' or 'out' the wrapper parses them as in1, in2 or out1, out2, respectively.
+
+  For all parameters, if more than two files are provided, the wrapper parses them as key=value1,value2,value3... 
+
+  Limitations:
+    - The wrapper cannot run bbwarap with paired + singleton files. But this is a rare use case.
```
fgvieira, line 26: What is the `bbwarap` tool? I could not find it on the documentation.

fgvieira:
Left some minor comments but, otherwise, great work!

bio/bbtools/meta.yaml
SilasK: bbwrap can process single + paired end reads, which uses some special input format for in1, in2.

SilasK:
@fgvieira  Do we need to wait for @johanneskoester  or can you accept the PR.

There will be probably some edge cases in the future. but I think it's quite neat for now.

bio/bbtools/test/Snakefile
```diff
@@ -0,0 +1,288 @@
...
+rule index_genome:
+    input:
+        ref="genome.fasta",
+    output:
+        flag=touch("genome.indexed"),
+    params:
+        command="randomreads.sh",
+        build= 1
+    log:
+        "logs/index_genome/{sample}.log",
```
fgvieira, line 56:
```suggestion
        "logs/index/genome.log",
```

SilasK:
Cool all checks have passed!

SilasK:
I think @johanneskoester You need to approve your requested changes

bio/bbtools/environment.yaml
fgvieira: Any reason for not using the latest `snakemake_wrapper_utils` version?

fgvieira:
It seems there is an issue with the test files.

SilasK:
It worked before 😠 .
It seems a problem with, writing and reading to the index.


BBmap stores index files in `ref/{buildnr}` it allows to change the build number but not really to change the path.
One can ignore the path and set different build numbers if needed. 
But I now set the index directory as an output, which makes things more explicit.

But now, it doesn't really correspond with the wrapper's guideline that input and output can be freely set.

I hope you grant this exception.

SilasK:
It is passing the checks now @fgvieira Thank you for your patience..

bio/bbtools/test/Snakefile
```diff
@@ -44,14 +46,15 @@ rule pileup:  # all rules use the same wrapper!!
...
 
 
+
 rule index_genome:
     input:
         ref="genome.fasta",
     output:
-        flag=touch("genome.indexed"),
+        # Unfortunately bbmap requires an index to be stored in this folder
+        flag=temp(directory("ref/genome/1"))
```
fgvieira: Should it be a flag, or can it be a `multiext`?

fgvieira: Just made a quick test, and it seems the `bbmap` index is a variable number of output files (with variable names). Isn't that so?

dlaehnemann:
> It worked before 😠 . It seems a problem with, writing and reading to the index.
> 
> BBmap stores index files in `ref/{buildnr}` it allows to change the build number but not really to change the path. One can ignore the path and set different build numbers if needed. But I now set the index directory as an output, which makes things more explicit.
> 
> But now, it doesn't really correspond with the wrapper's guideline that input and output can be freely set.
> 
> I hope you grant this exception.

I think you can actually give it an alternative path using the `path=/another/location/` command line argument, which would then be the way to go here. See point `8.` for the index creation here:
https://jgi.doe.gov/data-and-tools/software-tools/bbtools/bb-tools-user-guide/bbmap-guide/

SilasK:
so bbmap index, creates files in `ref/genome/{build_nr}` and `ref/index/{buildnr}`
with `path=path/to /folder` we can give a folder where the ref folder is placed. But, it doesn't correspond to the goal of freely setting an output path.

dlaehnemann:
But it seems like you are already trying things along those lines... :D

SilasK:
Thank you for the suggestion. It should work now.

fgvieira:
Can you bump the `snakemake-wrappers` version?

dlaehnemann:
> Thank you for the suggestion. It should work now.

And it was easier than I thought. Cool!

SilasK:
version updated

SilasK:
But apparently, bbmap's jave doesn't understand the `-Djava.io.tmpdir=/tmp` 

Should I deowngread the snakemake_wrapper_utils

fgvieira:
Damn... :disappointed:
Not ideal, but I guess we can pin `snakemake-wrapper-utils` to version `0.5.3` until `bbtools` supports it.
@dlaehnemann @johanneskoester What do you think?

dlaehnemann:
Arrgh, that's ugly. None of the BBTools shell scripts passes anything except for the memory specification on to Java. I guess the thorough (but laborious) fix would be to switch to the actual java script syntax, namely this:
```


 Maybe we should add an argument to `get_java_opts(adjust_tmpdir=True)`, which you can then unset in your `get_java_opts()` call.

**Comment by dlaehnemann** [comment]
I think you could parse the file `ToolDescriptions.txt` from the folder `bbmap/docs/` of the [downloadable `BBTools` source code from sourceforge](https://sourceforge.net/projects/bbmap/files/latest/download) to get a mapping from shell script to java tool. But then I also just saw that they seem to pass some hard-coded default values to some tools. So switching from the shell scripts will loose default parameter settings... :exploding_head:

**Comment by dlaehnemann** [comment]
Hmm, but if we look at the actual java command that is executed, the option is still in there. So it is simply the sanity checking of BBTools that is off here. Maybe passing `-da` will turn this off? But then, this might turn other useful checks off...

**Comment by dlaehnemann** [comment]
Mhmmm, no. It seems like `-da` will not turn off the failure we get...

**Comment by SilasK** [comment]
Alternatively, we could re-implement the get_java_opts inside the wrapper?

**Comment by dlaehnemann** [comment]
Or remove the respective part from the resulting `java_opts` string via a regex. :see_no_evil:

**Comment by dlaehnemann** [comment]
Or maybe we can change the `get_java_opts` to instead set all of the general option in `os.environ["JDK_JAVA_OPTIONS"] = "-Djava.io.tmpdir={}"`. This sounds like this should then automatically use those options in any following `java` call:
https://docs.oracle.com/en/java/javase/17/docs/specs/man/java.html#using-the-jdk_java_options-launcher-environment-variable

Why are Java tools always such a headache... :disappointed:

**Comment by SilasK** [comment]
Change impelemented.

**Comment by dlaehnemann** [comment]
Nice. But yes, (not) knowing Java and all, I am also a bit concerned this might not work as expected. And have no idea how to easily test this.

Maybe one could try to locally (and manually) run the tests on this pull request and on some other wrapper(s) that use the `get_java_opts()` function, while having manually modified the respective conda environments' `snakemake-wrapper-utils` to include the change...?!

**Comment by dlaehnemann** [comment]
I think you might have to split the `'--default-resources mem_mb=200'` into two separate arguments, so `'--default-resources', 'mem_mb=200'`.

**Comment by SilasK** [comment]
@dlaehnemann  To follow up on 
> But this is something that the user might want to set via snakemake command-line options or via the `resources: tempdir=` directive in a rule using the wrapper, so it would then be good if the wrapper did pass this on to the actual execution of the tool.
https://github.com/snakemake/snakemake-wrapper-utils/pull/37#issuecomment-1805909767

I don't think bbmap uses tempdirs a lot. But I found that one command can use it.

> The location of temp files can be specified with the “tmpdir” flag; it defaults to the environment variable $TMPDIR, which on Genepool points to local disk when available. Temp files will be cleaned up once BBNorm finishes.

`$TMPDIR` will already be set by snakemake using the `resources.tempdir` isn't it. So I would say the specification of tempdir should be ok, fo rbbtools.

File: bio/bbtools/wrapper.py [hunk 12]
```diff
@@ -0,0 +1,444 @@
...
+    # Getting memory.
+    if "-Xmx" in extra:
+        sys.exit(java_mem_xmx_error("extra"))
+
+    mem_mb = get_mem(snakemake)
+    logger.debug(
+        f"Memory specified {mem_mb} MiB, use {(1.0 - java_mem_overhead_factor)*100}% of it for Java heap."
+    )
+
+    return " -Xmx{}M".format(round(mem_mb * (1.0 - java_mem_overhead_factor)))
```
dlaehnemann, line 221: Sorry to have one more suggestion here: wouldn't it be a bit more future-proof to simply use the snakemake-wrapper-utils `get_java_opts` and simply extract only the `-Xmx` option with a regex? That way, we don't reimplement it here and if something gets fixed in the utils function, we also benefit here.

dlaehnemann (commented):
It seems like I had some comments a while ago, that are still lurking around. I think only the latest one is relevant and will try to resolve all the others. Sorry for the noise.

And many thanks for following up on the `$TMPDIR` question. If `bbtools` don't seem to use it and if they do, to respect the environment variable, then I think you are right -- we can ignore this for the wrapper...

SilasK:
I now extract the mem using a regex.

I also made a major update to the Snakefile test. It now represents a good usage of the wrapper and the bbtools functions. 
e.g. previously all parameters were input as a comma-separated string. 
With the same rules, it works for paired or single-end reads.

What do you think of my updated workflow?

fgvieira:
I think it looks good but, since it will introduce breaking changes, not sure if @johanneskoester want to take a look first.

dlaehnemann (approved):
This generally looks good to me. But I haven't looked at all the wrapper.py code in detail -- @johanneskoester has requested changes anyway, and will thus have to have another look. I'll ping him...

johanneskoester (approved):

bio/bbtools/test/Snakefile
```diff
@@ -0,0 +1,188 @@
+include: "input_functions.smk"
```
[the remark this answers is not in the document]
SilasK: I had it this way. But then the lining didn't pass.

bio/bbtools/test/Snakefile
```diff
@@ -0,0 +1,188 @@
...
+        expand("logs/loglog/{sample}.log", sample=["SampleA"]),
+        expand("covstats/{sample}.tsv", sample=["SampleA"]),
+
+
+rule index_genome:
+    input:
+        ref="genome.fasta",
+    output:
+        # path to the index base folder, The index is stored in {path}/ref/genome/{build} and {path}/ref/index/{build}
+        # With build by defualt 1 and path by default the working directory
```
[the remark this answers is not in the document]
SilasK: Maybe it's not clear, The comment explains the workings of path and build. 
But maybe we don't need to bother the users with that.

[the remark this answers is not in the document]
SilasK: I adapted the commend and moved it down in the Snakefile as it is less important.

bio/bbtools/wrapper.py
```diff
@@ -0,0 +1,456 @@
...
+
+# global flags to check if input and output are parsed multiple times
+parsed_input, parsed_output = False, False
+
+import warnings
+from snakemake_wrapper_utils import java
+
+
+import sys
+import warnings
```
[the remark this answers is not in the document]
SilasK: imported at the beginning now.

I have a second section of imports after all the logging is set up.
So that errors in imports would also be logged to the log file.

SilasK:
Now we are almost done!

I had two problems with the linter:

1. It didn't want me to have functions in the main Snakemake, therefore the import. Even if I think it's less ideal.

2. It interpreted a `ktrim="r"` as a path, and didn't accept it. I found a solution, which is less optimal.

SilasK:
In the wrapper I put the links to other PRs in snakemake that will incorporate these functions. But I guess we should not wait until then to release the next version of this wrapper.

bio/bbtools/test/Snakefile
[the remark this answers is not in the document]
dlaehnemann: Ah, I see. That's a bit annoying in this setup. But I guess there's no way around it for now. Maybe add a quick comment above the include, which rule this is for and the function name? Although the file name is rather descriptive already... :D

bio/bbtools/wrapper.py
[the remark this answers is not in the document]
dlaehnemann: Ah, OK. That makes sense. Maybe add a quick comment at the start, that there is another import block below, at the `start of the wrapper`?

dlaehnemann:
> Now we are almost done!
Yes, definitely!

> I had two problems with the linter: 
>     1. It didn't want me to have functions in the main Snakemake, therefore the import. Even if I think it's less ideal.
Then I guess we'll leave it as is. Maybe add a comment as suggested, if you think this is useful.
 
>     2. It interpreted a `ktrim="r"` as a path, and didn't accept it. I found a solution, which is less optimal.
Is that under params? Because I would find that weird, and possibly a linter bug. But if it enforces this for input and output, this makes sense. Then, I would suggest simply specifying this as a param. And if that's what you did, did you try something like either of the following, to make it work with the current linter:
```
params:
    ktrim="'r'",
    ktrim="'r' ",
    ktrim="r ",
```

As for the links to PRs, I think it is perfectly fine to simply leave those as TODOs. Maybe make it a bit clearer still, that once these are merged, the respective functions can simply go away and use snakemake internal functionality.

SilasK:
It looks like a linter bug:
[Code](https://github.com/metagenome-atlas/snakemake-wrappers/blob/37d0bf2dfcfbde20269090c687e4184136033779/bio/bbtools/test/Snakefile#L88C1-L92C19)
[Error](https://github.com/snakemake/snakemake-wrappers/actions/runs/6849925541/job/18623086042)

dlaehnemann:
Ah, thanks for the pointers to the linter. Regarding the (input) function, you could indeed make this an inline lambda expression like this:

```
rule loglog:
    input:
        expand(
            "reads/raw/{{sample}}_{fraction}.fastq.gz",
            fraction=lambda wc: ["R1", "R2"] if config.get("reads_are_paired", True) else ["se"]
        ),
```

It has the advantage of being right there in the input of the rule. It has the disadvantage of being slightly less readable than the actual function with newlines. And you'll have to repeat that code wherever you provide such a read `fraction` specification. It seems like there's no perfect solution with the current linter.

And from what I understand about the `ktrim="r"`, this recognizes that `r` is a prefix of the input/output path, which starts with `reads/...`. That's a very short prefix, but it is correct. The linter could probably be amended to ignore such single letter cases or make it more of a suggestion, then a must (not sure if that exists, though). And I think you should get around this by including some random whitespace at the end, so:
`ktrim="r "`
Then, it shouldn't be a prefix any more...

SilasK:
The whitespace fixed the linter.

I find the lambda version less clear.

dlaehnemann:
Agreed, I'll merge! :tada:

SilasK:
Thank you very much for your review and your suggestions!

dlaehnemann:
Thanks for keeping at it! Getting things right for reproducible and reusable data analyses can be exhausting, especially when you're battling with multiple levels of the workflow / analysis stack and several workflows at the same time.

So take a moment and celebrate! :rocket:

fgvieira:
Nice work!!

PR #1397, opened 2023-05-26
feat: add ragtag wrapper

currocam:
### Description

Add RagTag wrappers. 
https://github.com/malonge/RagTag/tree/master

Some test files are not small, but the tool requires alignments of a certain length not to fail. It seems more reasonable to me not to change the test case to use unrealistic parameters and add some test files as small as possible. 

### QC

* [x] I confirm that:


bio/ragtag/correction/wrapper.py
```diff
@@ -0,0 +1,40 @@
...
+import os.path
+
+
+log = snakemake.log_fmt_shell()
+
+# Check that two input files were supplied
+n = len(snakemake.input)
+assert n == 2, "Input must contain 2 files. Given: %r." % n
+query = snakemake.input.query
+reference = snakemake.input.reference
```
fgvieira, line 19:
```suggestion

```

bio/ragtag/correction/wrapper.py
```diff
@@ -0,0 +1,40 @@
...
+out_dir = os.path.dirname(snakemake.output[0])
+for file_path in snakemake.output[1:]:
+    assert out_dir == os.path.dirname(file_path), (
+        "ragtag can only output files to a single directory."
+        " Please indicate only one directory for the output files."
+    )
+
+shell(
+    "ragtag.py correct"
+    " {snakemake.input.reference}"
```
fgvieira, line 35: just for the sake of consistency with other wrappers:

```suggestion
    " {snakemake.input.ref}"

```

bio/ragtag/correction/test/Snakefile
```diff
@@ -0,0 +1,14 @@
+rule correction:
+    input:
+        query="fasta/{query}.fasta",
+        reference="fasta/{reference}.fasta",
```
fgvieira, line 4: For consisency

```suggestion
        ref="fasta/{reference}.fasta",
```

bio/ragtag/correction/meta.yaml
```diff
@@ -0,0 +1,14 @@
+name: ragtag-correction
+description: |
+  Homology-based misassembly correction.
+authors:
+  - Curro Campuzano Jiménez
+input:
+  - reference fasta file (uncompressed or bgzipped)
+  - query fasta file (uncompressed or bgzipped)
```
fgvieira, line 8:
```suggestion
  - ref: reference fasta file (uncompressed or bgzipped)
  - query: query fasta file (uncompressed or bgzipped)

```

bio/ragtag/correction/meta.yaml
```diff
@@ -0,0 +1,14 @@
...
+  - Curro Campuzano Jiménez
+input:
+  - reference fasta file (uncompressed or bgzipped)
+  - query fasta file (uncompressed or bgzipped)
+output:
+  - The corrected query assembly in FASTA format.
+  - The AGP file defining the exact coordinates of query sequence breaks.
+params:
+  - extra: additional parameters
+notes: https://github.com/malonge/RagTag/wiki/correct
```
fgvieira, line 14: Can you add the URL on its own field? Just after `description`, add `url: https://github.com/malonge/RagTag/wiki/correct`

currocam: Thanks!

fgvieira:
It seems that `ragtag` outputs files to the same output folder. However, as stated in the PR QC, "file paths in the resulting rule can be changed arbitrarily".

One way to do it is to output everything to a temp folder, and then copy the relevant files to their final destinations (e.g. see [STAR](https://github.com/snakemake/snakemake-wrappers/blob/master/bio/star/align/wrapper.py) wrapper).

bio/ragtag/correction/meta.yaml
```diff
@@ -0,0 +1,14 @@
...
+description: Homology-based misassembly correction.
+url: https://github.com/malonge/RagTag/wiki/correct
+authors:
+  - Curro Campuzano Jiménez
+input:
+  - ref: reference fasta file (uncompressed or bgzipped)
+  - query: query fasta file (uncompressed or bgzipped)
+output:
+  - The corrected query assembly in FASTA format.
+  - The AGP file defining the exact coordinates of query sequence breaks.
```
fgvieira, line 11:
```suggestion
  - fasta: The corrected query assembly in FASTA format.
  - agp: The AGP file defining the exact coordinates of query sequence breaks.

```

currocam: Done! Thank you so much for being so patient, you are very kind :)

bio/ragtag/patch/meta.yaml
```diff
@@ -0,0 +1,20 @@
+name: ragtag-path
+description: Homology-based assembly patching.
+url: https://github.com/malonge/RagTag/wiki/patch
+authors:
+  - Curro Campuzano Jiménez
+input:
+  - reference fasta file (uncompressed or bgzipped)
+  - query fasta file (uncompressed or bgzipped)
```
fgvieira, line 8:
```suggestion
  - ref: reference fasta file (uncompressed or bgzipped)
  - query: query fasta file (uncompressed or bgzipped)

```

bio/ragtag/patch/test/Snakefile
```diff
@@ -0,0 +1,20 @@
+rule patch:
+    input:
+        query="fasta/{query}.fasta",
+        reference="fasta/{reference}.fasta",
```
fgvieira, line 4:
```suggestion
        ref="fasta/{reference}.fasta",
```

bio/ragtag/patch/wrapper.py
```diff
@@ -0,0 +1,54 @@
...
+]
+for key in snakemake.output.keys():
+    assert (
+        key in valid_keys
+    ), "Invalid key in output. Valid keys are: %r. Given: %r." % (valid_keys, key)
+
+with tempfile.TemporaryDirectory() as tmpdir:
+    shell(
+        "ragtag.py patch"
+        " {snakemake.input.reference}"
```
fgvieira, line 40:
```suggestion
        " {snakemake.input.ref}"

```

bio/ragtag/patch/wrapper.py
```diff
@@ -0,0 +1,54 @@
...
+from snakemake.shell import shell
+
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+# Check that two input files were supplied
+n = len(snakemake.input)
+assert n == 2, "Input must contain 2 files. Given: %r." % n
+query = snakemake.input.query
+reference = snakemake.input.reference
```
fgvieira, line 18:
```suggestion

```

bio/ragtag/scaffold/wrapper.py
```diff
@@ -0,0 +1,41 @@
...
+from snakemake.shell import shell
+import tempfile
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+# Check that two input files were supplied
+n = len(snakemake.input)
+assert n == 2, "Input must contain 2 files. Given: %r." % n
+query = snakemake.input.query
+reference = snakemake.input.reference
```
fgvieira, line 18:
```suggestion
```

bio/ragtag/scaffold/wrapper.py
```diff
@@ -0,0 +1,41 @@
...
+valid_keys = ["agp", "fasta", "stats"]
+for key in snakemake.output.keys():
+    assert (
+        key in valid_keys
+    ), "Invalid key in output. Valid keys are: %r. Given: %r." % (valid_keys, key)
+
+with tempfile.TemporaryDirectory() as tmpdir:
+    shell(
+        "ragtag.py scaffold"
+        " {snakemake.input.reference}"
```
fgvieira, line 31:
```suggestion
        " {snakemake.input.ref}"

```

bio/ragtag/scaffold/test/Snakefile
```diff
@@ -0,0 +1,15 @@
+rule scaffold:
+    input:
+        query="fasta/{query}.fasta",
+        reference="fasta/{reference}.fasta",
```
fgvieira, line 4:
```suggestion
        ref="fasta/{reference}.fasta",
```

bio/ragtag/scaffold/meta.yaml
```diff
@@ -0,0 +1,15 @@
+name: ragtag-scaffold
+description: Homology-based assembly scaffolding.
+url: https://github.com/malonge/RagTag/wiki/scaffold
+authors:
+  - Curro Campuzano Jiménez
+input:
+  - reference fasta file (uncompressed or bgzipped)
+  - query fasta file (uncompressed or bgzipped)
```
fgvieira, line 8:
```suggestion
  - ref: reference fasta file (uncompressed or bgzipped)
  - query: query fasta file (uncompressed or bgzipped)

```

bio/ragtag/merge/test/Snakefile
```diff
@@ -0,0 +1,15 @@
+rule merge:
+    input:
+        "input/{assembly}.fasta",
+        "input/{scaffolds1}.agp",
+        "input/{scaffolds2}.agp",
```
fgvieira, line 5: Wouldn't it be easier to separate the fasta and AGP files?

```suggestion
        ref=["input/{assembly}.fasta"],
        #bam=["input/{assembly}.bam"],
        agp=["input/{scaffolds1}.agp", "input/{scaffolds2}.agp"],
```

currocam: Yes, it makes much more sense. I didn't know you could have as a named input a list of files (because the wrapper should accept an arbitrary number of agp files). Thanks!

bio/ragtag/patch/meta.yaml
```diff
@@ -0,0 +1,21 @@
+name: ragtag-path
+description: Homology-based assembly patching.
+url: https://github.com/malonge/RagTag/wiki/patch
+authors:
+  - Curro Campuzano Jiménez
+input:
+  - ref: reference fasta file (uncompressed or bgzipped)
+  - query: query fasta file (uncompressed or bgzipped)
+
```
fgvieira, line 9:
```suggestion
```

bio/ragtag/scaffold/wrapper.py
```diff
@@ -0,0 +1,39 @@
...
+
+from snakemake.shell import shell
+import tempfile
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+# Check that two input files were supplied
+n = len(snakemake.input)
+assert n == 2, "Input must contain 2 files. Given: %r." % n
+
```
fgvieira, line 17:
```suggestion
```

bio/ragtag/patch/wrapper.py
```diff
@@ -0,0 +1,52 @@
...
+import tempfile
+from snakemake.shell import shell
+
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+# Check that two input files were supplied
+n = len(snakemake.input)
+assert n == 2, "Input must contain 2 files. Given: %r." % n
+
```
fgvieira, line 17:
```suggestion
```

bio/ragtag/merge/wrapper.py
```diff
@@ -0,0 +1,55 @@
...
+__author__ = "Curro Campuzano Jiménez"
+__copyright__ = "Copyright 2023, Curro Campuzano Jiménez"
+__email__ = "campuzanocurro@gmail.com"
+__license__ = "MIT"
+
+
+from snakemake.shell import shell
+import tempfile
+
+
```
fgvieira, line 12: I just noticed that the wrapper uses `snakemake.params.extra` directly. This will fail if the rule does no define `params.extra`. To avoid that, it is better to check and `extra`:

```suggestion
extra = snakemake.params.get("extra", "")
```

And can you also change the other wrappers?

fgvieira (approved):

bio/ragtag/merge/wrapper.py
```diff
@@ -0,0 +1,57 @@
...
+assert len(agp_files) >= 2, "Input must contain at least 2 agp files. Given: %r." % len(
+    agp_files
+)
+
+bam_file = snakemake.input.get("bam")
+
+# Add Hi-C BAM file to params if present
+if bam_file:
+    snakemake.params.extra += f" -b {bam_file}"
```
fgvieira, line 29: This will fail if `params.extra` is not defined. Usually the best is to declare `extra = snakemake.params.get("extra", "")` at the top (e.g. right after `log`) and use `extra` from then on.

bio/ragtag/patch/wrapper.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    for key in valid_keys[:-1]:
+        outfile = snakemake.output.get(key)
+        if outfile:
+            extension = key.replace("_", ".")
+            shell("mv {tmpdir}/ragtag.patch.{extension} {outfile}")
+    outdir = snakemake.output.get("asm_dir")
+    if outdir:
+        # Move files into directory outdir
+        shell("mkdir -p {outdir} && mv {tmpdir}/ragtag.patch.* {outdir}")
```
fgvieira, line 50: It seems the directory `asm_dir` will contain the same files as specified in the rule, no?

bio/ragtag/merge/wrapper.py
currocam: Got it!

bio/ragtag/patch/wrapper.py
currocam: Yes, you are absolutely right. Trying it locally probably worked because they were the last files left. I've just changed to move only 'ragtag.patch.asm.*', aka assembly files.

fgvieira: Do these `ragtag.patch.asm.*` files always have the same name? Or you can get different number of files with variable names?

currocam: According to the documentation, I believe you  can get different number of files with variable names depending on your input and the aligner.

fgvieira: Ok, then I guess it will have to be a `directory`.

fgvieira:
Thanks for your work on this!
I think it is fine now and have approved it but won't merge it now, just in case @johanneskoester has some more comments.

PR #1430, opened 2023-06-14
feat: Xsv (complete subcomand suite)

tdayris:
### Description

This PR adds [xsv](https://github.com/BurntSushi/xsv) select, to the list of available wrappers.

### QC

* [X] I confirm that:


fgvieira:
I think it looks good, but I was hoping that it would be a bit more straightforward. 😄 
I would try to make it as simple as possible and leave it up to the user to make sure of provided params/options.

For example, you assign `--delimiter`, `--out-delimiter` and `--jobs` depending on each command. What happens if you (e.g.) provide `--jobs` to `xsv headers`? I assume it is not threaded, but does it fail? Or does it just ignore the option? If the latter, you could just specify `--jobs` for all commands.
As for `--delimiter` and `--out-delimiter`, the user could specify it on `params.extra`. Maybe it would not match the extension, but it would be more flexible and allow to specify other delimiters.

The same with line 71, checking `command == cat`; if the user provides just `cat` I guess the program will fail and the rule will also fail, no? So snakemake will already give an error if the user only specifies `cat`. Also with checking the number of input and output files; I'd leave it up to the user to make sure the correct number of files are provided. @johanneskoester what do you think.

tdayris:
Unfortunately, providing unexpected parameters throws exceptions. Example, `xsv headers` does not accept neither `--jobs` nor `--output`, and `xsv cat rows` does not accept `--out-delimiter`, etc. They almost all accept input delimiters, most of them also accept `--no-headers`.

The subcommand `cat` expects a positional parameter `row` or `column` as first parameter. The rest of the command line interface is identical for both sub-sub-commands.

I can leave the number of input / output  file for the end-user. With the exception of `xsv join` which especially expects two, and only two input files in a given order. Originally, I was thinking about `left` / `right` parameters as defined in `bedtools intersect`, but not with a common wrapper.

fgvieira:
Then I guess `--jobs` needs to be command-specific. But I am inclined to just leave (e.g.) `--no-headers`, `--delimiter` and `--out-delimiter` for the user to specify in `params.extra`.

On `cat`, if the user user does not specify `row` or `column`, it will throw an error, right? If so, there I think there is no need for the wrapper to check that.

tdayris:
My bad, `--delimiter` is always accepted.

I left almost all parameters to the user, as you advised. Would you mind checking my approach with the `join` subcommand ? I can't see how to make it consistent with other subcommands.

bio/xsv/test/Snakefile
```diff
@@ -304,17 +288,46 @@ rule test_xsv_sort:
...
+        table="table.csv",
     output:
-        directory("split.csv"),
+        directory("split"),
     threads: 1
     log:
         "xsv/split.log",
     params:
         subcommand="split",
+        extra=lambda wildcards, output: f"-s 2 {output}",
```
fgvieira, line 299: Do you think the wrapper can deal with this `{output}`? You already have a special case for `join`, so it would be cleaner to also deal with `output`.

bio/xsv/wrapper.py
```diff
@@ -0,0 +1,48 @@
...
+# Command line building
+if subcommand == "join":
+    shell(
+        "xsv {subcommand} {extra} "
+        "{snakemake.params.col1} {snakemake.input.table[0]} "
+        "{snakemake.params.col2} {snakemake.input.table[1]} "
+        "> {snakemake.output} {log}"
+    )
+elif subcommand == "index":
+    log = snakemake.log_fmt_shell(stdout=True, stderr=True)
```
fgvieira, line 36: Is anything actually printed to `stdout`?

fgvieira (approved):

tdayris:
I fixed an issue that occurred when a list of output files was provided to `xsv split` instead of a directory.

PR #1448, opened 2023-06-17
feat: barrnap-wrapper

currocam:
### Description
A wrapper for barrnap, a  bacterial ribosomal RNA predictor.

### QC

* [ ] I confirm that:


bio/barrnap/wrapper.py
```diff
@@ -0,0 +1,44 @@
...
+
+valid_keys = ["gff", "fasta"]
+for key in snakemake.output.keys():
+    assert (
+        key in valid_keys
+    ), "Invalid key in output. Valid keys are: %r. Given: %r." % (valid_keys, key)
+
+gff_out = snakemake.output.get("gff")
+assert gff_out, "Output must contain at least a gff file."
+
```
fgvieira, line 31: This part is not really necessary.

```suggestion
```

bio/barrnap/wrapper.py
```diff
@@ -0,0 +1,44 @@
...
+if fasta_out:
+    extra += f" -o {fasta_out}"
+
+shell(
+    "barrnap"
+    " --threads {snakemake.threads}"
+    " -k {kingdom}"
+    " {extra}"
+    " < {snakemake.input}"
+    " > {gff_out}"
```
fgvieira, line 42:
```suggestion
    " > {snakemake.output.gff}"
```

bio/barrnap/wrapper.py
```diff
@@ -0,0 +1,44 @@
...
+fasta_out = snakemake.output.get("fasta")
+if fasta_out:
+    extra += f" -o {fasta_out}"
+
+shell(
+    "barrnap"
+    " --threads {snakemake.threads}"
+    " -k {kingdom}"
+    " {extra}"
+    " < {snakemake.input}"
```
fgvieira, line 41:
```suggestion
    " < {snakemake.input.fasta}"
```

bio/barrnap/wrapper.py
```diff
@@ -0,0 +1,44 @@
...
+assert gff_out, "Output must contain at least a gff file."
+
+fasta_out = snakemake.output.get("fasta")
+if fasta_out:
+    extra += f" -o {fasta_out}"
+
+shell(
+    "barrnap"
+    " --threads {snakemake.threads}"
+    " -k {kingdom}"
```
fgvieira, line 39:
```suggestion
    " -k {snakemake.params.kingdom}"

```

bio/barrnap/wrapper.py
```diff
@@ -0,0 +1,35 @@
...
+extra = snakemake.params.get("extra", "")
+kingdom = snakemake.params.get("kingdom", "")
+assert kingdom in [
+    "bac",
+    "arc",
+    "euk",
+    "mito",
+], "Invalid kingdom. Valid kingdoms are: 'bac', 'arc', 'euk', 'mito'. Given: %r." % (
+    kingdom
+)
```
fgvieira, line 21: The program will fail if the kingdom is not one of those, right?
If so, this part is not necessary.

currocam: Yes, you're right

currocam: I just added the default kingdom (the one you use if there's no --kingdom flag)

bio/barrnap/wrapper.py
```diff
@@ -0,0 +1,27 @@
...
+__copyright__ = "Copyright 2023, Curro Campuzano Jiménez"
+__email__ = "campuzanocurro@gmail.com"
+__license__ = "MIT"
+
+
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+extra = snakemake.params.get("extra", "")
+kingdom = snakemake.params.get("kingdom", "bac")
```
fgvieira, line 13:
```suggestion
```

currocam: Actually, it was a typo. I was trying to use it in the shell command with --kingdom. With the default "bac", if no kingdom parameter is given in the params, the program runs as if no --kingdom paramater was given (so as expected)

fgvieira (approved):

Issue #624, opened 2022-10-25
BUG:  BWA-MEM wrapper - Passing command line argument -O in extra triggers wrapper cancel


KuechlerO:
Hello guys,
I stumbled over an unexpected behaviour for the wrapper [BWA MEM](https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/bwa/mem.html) - `v1.17.3/bio/bwa/mem`:

**Snakemake version**
7.12.1

**Describe the bug**
Including the command line argument `-O` with any INT as passed value is causing the cancelation of the called wrapper. It seems that `-O` is protected somehow. However, it is a valid argument for the "BWA MEM" algorithm.

Print-out on command line: `You have specified output format (-O/--output-fmt) in params.extra; this is automatically infered from output file extension.`

**Logs**
```
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 50
Rules claiming more threads will be scaled down.
Provided resources: mem_mb=204800, disk_mb=15169, cpus=50, time_min=360
Select jobs to execute...

[Tue Oct 25 20:39:25 2022]
rule bwa_mem_custom2:
    input: /sc-scratch/sc-scratch-btg/olik_splicing_project/splice-prediction/snakemake_workflows/Snakemake_Main/output/module1_qc_preproc_and_alignment/output/trimmomatic/76664_R1.fastq.gz, /sc-scratch/sc-scratch-btg/olik_splicing_project/splice-prediction/snakemake_workflows/Snakemake_Main/output/module1_qc_preproc_and_alignment/output/trimmomatic/76664_R2.fastq.gz, output/bwa/ref_genome/hg19.amb, output/bwa/ref_genome/hg19.ann, output/bwa/ref_genome/hg19.bwt, output/bwa/ref_genome/hg19.pac, output/bwa/ref_genome/hg19.sa
    output: bwa/mapped_custom2/76664.bam
    log: logs/bwa_mem_custom2/76664.log
    jobid: 0
    reason: Forced execution
    wildcards: sample=76664
    threads: 50
    resources: mem_mb=204800, disk_mb=15169, tmpdir=/tmp, cpus=50, time_min=360

python -c "import sys; print('.'.join(map(str, sys.version_info[:2])))"
Activating conda environment: .snakemake/conda/06e9e9e3769739946610da9d3cba085a
[...]

PR #1450, opened 2023-06-20
fix: bug with -O being used for gap opening penalty and samtools output directory

huzuner:
### Description

Gap penalty of minimap2 is done via "-O" option and the very same option also exists for samtools, but to indicate output directory. Using "-O" in the "extra" leads to the following error because of using snakemake utils for samtools (https://github.com/snakemake/snakemake-wrapper-utils/blob/f9cba17cb380dc7e6729a845ae37f26972c3d1dd/snakemake_wrapper_utils/samtools.py#LL76C23-L76C23):

```
You have specified output format (`-O/--output-fmt`) in params.extra; this is automatically inferred from output file extension.
```

The fastest and not over-engineered solution to me was to add the gap_opening variable to params, but I feel that it's also not very elegant especially when it comes to modifying the other penalty scores, but they can still be set in the extra.

### QC

* [ ] I confirm that:


dlaehnemann (changes requested):
Ah, looking at this in more detail, this makes sense. The `get_samtools_opts` is actually only used when samtools is used for sorting of the output. So this solution is probably the best and easiest.

One thing though. Could you document this further by:
* [ ] adding an entry for this new `params` field in the `meta.yaml` under `notes:`: https://github.com/snakemake/snakemake-wrappers/blob/3cfaa10fd6d36e65154cbe7d948d0915b5db83ca/bio/minimap2/aligner/meta.yaml#L13-L16
* [ ] adding this new `params: gap_opening: ""` specification to at least one of the example rules with sorted sam or bam output, so for example here: https://github.com/snakemake/snakemake-wrappers/blob/3cfaa10fd6d36e65154cbe7d948d0915b5db83ca/bio/minimap2/aligner/test/Snakefile#L60-L63

bio/minimap2/aligner/wrapper.py
```diff
@@ -45,6 +46,7 @@
     "(minimap2"
     " -t {snakemake.threads}"
     " {extra} "
+    " -O {gap_opening} "
```
dlaehnemann, line 49: Currently, the `-O ` always gets added to the command, even if no `gap_opening` is specified. So this rather needs to become:
```suggestion
    "{gap_opening} "
```

And then the `-O ` needs to be prepended to gap_opening above, if it is none-empty. This should be possible with something like:
```
if gap_opening:
    gap_opening = f"-O {gap_opening}"
```
```

**Comment by huzuner** [comment]
Thank you! How does it look now?

File: bio/minimap2/aligner/wrapper.py [hunk 2]
```diff
@@ -40,11 +41,15 @@
     else:
         raise ValueError(f"Unexpected value for params.sort: {sort}")
 
+#prepend -O to the command only if gap_opening is specified
```
dlaehnemann, line 44: Just a minor formatting suggestion:
```suggestion
# prepend -O to the command only if gap_opening is specified
```

dlaehnemann (approved):
Looks good now, just one minor formatting suggestion -- feel free to include it or not.

johanneskoester (changes requested):
I would prefer to not have a special parameter. I think it is a bug that the wrapper utils complain here about -O, they should only check sort_extra.

dlaehnemann:
But then you will have to change all the samtools wrappers, as `get_samtools_opts()` just parses `params: extra:`, no other `params: ` entry. So this would maybe have to become a `params: samtools_extra` wherever `get_samtools_opts()` is used. Or we cannot use `get_samtools_opts()` here at all. Or we have to generalize the function to have another argument which determines which `params:` field to parse. Maybe this last option (with the default set to `"extra"`) is the least breaking option.

johanneskoester:
> But then you will have to change all the samtools wrappers, as `get_samtools_opts()` just parses `params: extra:`, no other `params: ` entry. So this would maybe have to become a `params: samtools_extra` wherever `get_samtools_opts()` is used. Or we cannot use `get_samtools_opts()` here at all. Or we have to generalize the function to have another argument which determines which `params:` field to parse. Maybe this last option (with the default set to `"extra"`) is the least breaking option.

Exactly the latter was what I meant. Here is the PR: https://github.com/snakemake/snakemake-wrapper-utils/pull/30

johanneskoester:
snakemake-wrapper-utils 0.6 have been released with the fix. Hence, this PR should just add a keyword argument `param_name="sort_extra"` to the call of get_samtools_opts.

fgvieira:
This seems to be the same issue described in #624.
Do you think you could also fix the `bwa-mem`, `bwa sampe`, `bwa samse`, and `bwa sam(pe/se)` wrappers?

PS - maybe `bwa sampe` and `bwa samse` should actually be deleted, since there is already `bwa sam(pe/se)`...

huzuner:
The tests currently fail because the conda package for snakemake-wrappers-utils aren't updated yet.

fgvieira:
Some other errors, related to memory.

huzuner:
I think removing 
```
        sort_extra="--write-index",  # Extra args for samtools/picard.
```
from `Snakefile_samtools` of `bwa/mem` would fix the `You have specified writing index (`--write-index`) in `params.extra`; this is automatically infered from `idx` output file.` error but I don't know about the memory error yet.

huzuner:
We get the error:
```
Invalid maximum heap size: -Xmx0M
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```
Regardless of the changes in the current PR, `-Xmx0m` returns 0 here related to a recent improvement in gathering resources in `get_mem()` in snakemake-wrapper-utils: https://github.com/snakemake/snakemake-wrapper-utils/blob/10ecf786c1bc3bcf6197eb8c6fb7f8615f0df2ee/snakemake_wrapper_utils/snakemake.py#L1

dlaehnemann:
I just merged and released a non-zero default value setting for `get_mem` over at `snakemake-wrapper-utils`:
https://github.com/snakemake/snakemake-wrapper-utils/pull/32

Once the new version `0.6.1` of `snakemake-wrapper-utils` is up on bioconda, pointing to it should fix this issue here.

dlaehnemann:
Version `0.6.1` of `snakemake-wrapper-utils` is available via bioconda, now.

dlaehnemann:
Ah, and I merged in the latest master branch, so you should probably pull the latest changes before further code changes locally on your machine.

huzuner:
Thank you! 
I now have updated the versions of snakemake-wrapper-utils for `bwa/mem` and `bwa-mem2`.

johanneskoester (approved):

PR #1451, opened 2023-06-21
fix: inconsistent output filenames trim-galore

### LINKED ISSUE #967: Trim galore inconsitent output filenames
Trim galore internally defines output filenames from input filenames. This can lead to inconsistencies with the output filenaming. 

Wrapper:
https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/trim_galore/pe.html

Idealy one should use

--basename <PREFERRED_NAME>

Use PREFERRED_NAME as the basename for output files, instead of deriving the filenames from the input files. Single-end data would be called PREFERRED_NAME_trimmed.fq(.gz), or PREFERRED_NAME_val_1.fq(.gz) and PREFERRED_NAME_val_2.fq(.gz) for paired-end data. --basename only works when 1 file (single-end) or 2 files (paired-end) are specified, but not for longer lists.

currocam:
### Description

Now output files can be changed arbitrarily. I think it closes #967 . 

### QC

* [ ] I confirm that:


bio/trim_galore/pe/wrapper.py
```diff
@@ -30,19 +31,36 @@
...
+    # Get all files in tmpdir
+    fasta_fwd, fasta_rev, report_fwd, report_rev = (
+        snakemake.output.get(key)
+        for key in ["fasta_fwd", "fasta_rev", "report_fwd", "report_rev"]
+    )
+
+    if fasta_fwd:
+        file = [f for f in os.listdir(tmpdir) if "_val_1" in f][
+            0
+        ]  # It always have _val_1
```
fgvieira, line 52: Can you not infer the file name from `snakemake.input` like in lines 61 and 65?

bio/trim_galore/pe/wrapper.py
```diff
@@ -64,3 +58,10 @@
     if report_rev:
         file = f"{os.path.basename(snakemake.input[1])}_trimming_report.txt"
         shell("mv {tmpdir}/{file} {report_rev}")
+
+    if fasta_fwd:
+        shell("mv {tmpdir}/*_val_1* {fasta_fwd}")
```
fgvieira, line 63: I am not familiar with trimgalore, but what if there are several files?
Isn't it possible to infer the actual file name? Ex:
```
shell("mv {tmpdir}/{prefix}_val_1.fas {fasta_fwd}")
```

bio/trim_galore/pe/wrapper.py
currocam: I have taken a look at the source code, and I think it is tricky.  Different flags like "--compress" or "--basename" also affect the output name. 

https://github.com/FelixKrueger/TrimGalore/blob/4edff97d22f3837d42a29e4afbfaeb6e07ffb11b/trim_galore#LL130C1-L144C69

So, what about just use `shell("mv {tmpdir}/*_val_1* {fasta_fwd}")` ?

bio/trim_galore/pe/wrapper.py
currocam: There must be a way; here are the relevant lines (but I'm not familiar with Pearl and the source code ~3500 lines :rofl: )

https://github.com/FelixKrueger/TrimGalore/blob/4edff97d22f3837d42a29e4afbfaeb6e07ffb11b/trim_galore#LL1598C1-L1626C1

Unless the name itself of the file contains '_val_1', I believe that file should be unique.

bio/trim_galore/pe/wrapper.py
fgvieira: Can we define a `--basename` so that you at least have a fixed prefix in the `temp` folder? Maybe even set it as a random string.

What if the user sets `--compress` but defines an output ending in `.fq`?
I don't know `trimgalore` that well, but I am just trying to make sure the right file is copied. Thinking in extreme cases, if the `temp` files are not deleted (they should be cleaned by python, but...) and then the wrapper copies an old output.

bio/trim_galore/pe/wrapper.py
currocam: I think I have solved the issue (now temporary output files are inferred) @fgvieira 
:smile:

fgvieira (approved):

PR #1464, opened 2023-06-29
fix: name of bamqc

christopher-schroeder:
### Description

The subcommand name was wrong in the description. Additionally add a region param fro target analyse.

### QC

* [x] I confirm that:


fgvieira:
Shouldn't target be an input file?
And can you also add a test using a target file?

christopher-schroeder:
I thought about that. The problem is, that this parameter is optional. When developing a generic pipeline you would like to have this paramter for exomes and dont want to have it genomes. So you normally build a function. Something like

```
rule qualimap:
    input:
        bam="results/alignment/{sample}.bam",
        target=get_target
```

In the case of genomic data, get_target is None or "". But if defined as input, Snakemake will check the file for existence. But it cannot check for None or "". So I think the solution then would be

def qualimap_input(wc):
    ret = {"bam": "results/alignment/{wc.sample}.bam"}
    if target:
        ret["target"] = target

```
rule qualimap:
    input:
        unpack(qualimap_input)
```

I am not convinced about either of the 2 options and in the end I decided to go for the params method. But if you prefer the other method you can convince me easily.

fgvieira:
I think the way to go would be option 2 and have it as an input.

bio/qualimap/bamqc/wrapper.py
```diff
@@ -22,9 +22,13 @@
     del os.environ["DISPLAY"]
 
 extra = snakemake.params.get("extra", "")
+if target := snakemake.params.get("target"):
+    extra = f"{extra} --feature-file {target}"
```
johanneskoester, line 26: Can you please add this to the example Snakefile (can be commented out, but people have to see that it exists and they can use it)?

johanneskoester (changes requested):

johanneskoester:
> I thought about that. The problem is, that this parameter is optional. When developing a generic pipeline you would like to have this paramter for exomes and dont want to have it genomes. So you normally build a function. Something like
> 
> ```
> rule qualimap:
>     input:
>         bam="results/alignment/{sample}.bam",
>         target=get_target
> ```
> 
> In the case of genomic data, get_target is None or "". But if defined as input, Snakemake will check the file for existence. But it cannot check for None or "". So I think the solution then would be
> 
> ```
> def qualimap_input(wc):
>     ret = {"bam": "results/alignment/{wc.sample}.bam"}
>     if target:
>         ret["target"] = target
>     return ret
> 
> rule qualimap:
>     input:
>         unpack(qualimap_input)
> ```
> 
> I am not convinced of either of the 2 options and in the end I decided to go for the params method. But if you prefer the other method you can convince me easily.

It is also possible to return `[]` from the input function to indicate that the target is missing.

fgvieira:
yes, it can also be option 1, if you return `None` when no target.
But it should be an input file

bio/qualimap/bamqc/wrapper.py
```diff
@@ -22,9 +22,13 @@
     del os.environ["DISPLAY"]
 
 extra = snakemake.params.get("extra", "")
+if target := snakemake.params.get("target"):
```
johanneskoester, line 25: Can you please also add the new param to the example (e.g. commented out and mentioned as optional)?

johanneskoester (changes requested):

fgvieira (approved):

PR #1550, opened 2023-07-20
feat: new inphared-db wrapper

nacnoriko:
### Description

Adding a new wrapper to download the current inphared database. 

### QC

* [ ] I confirm that:

### DIFF
--- a/bio/reference/inphared-db/meta.yaml
+++ b/bio/reference/inphared-db/meta.yaml
@@ -1,0 +1,4 @@
+name: inphared-db
+description: Download sequence file from the Inphared database (https://github.com/RyanCook94/inphared/blob/main/README.md), and store them in a single .fasta file. Please check the current database available at the above link and adjust the config file. 
+authors:
+  - Noriko A. Cassman
--- a/bio/reference/inphared-db/old_wrapper.py
+++ b/bio/reference/inphared-db/old_wrapper.py
@@ -1,0 +1,80 @@
+__author__ = "Johannes Köster"
+__copyright__ = "Copyright 2019, Johannes Köster"
+__email__ = "johannes.koester@uni-due.de"
+__license__ = "MIT"
+
+import subprocess as sp
+import sys
+from itertools import product
+from snakemake.shell import shell
+
+species = snakemake.params.species.lower()
+release = int(snakemake.params.release)
+build = snakemake.params.build
+
+branch = ""
+if release >= 81 and build == "GRCh37":
+    # use the special grch37 branch for new releases
+    branch = "grch37/"
+elif snakemake.params.get("branch"):
+    branch = snakemake.params.branch + "/"
+
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+
+spec = ("{build}" if int(release) > 75 else "{build}.{release}").format(
+    build=build, release=release
+)
+
+suffixes = ""
+datatype = snakemake.params.get("datatype", "")
+chromosome = snakemake.params.get("chromosome", "")
+if datatype == "dna":
+    if chromosome:
+        suffixes = ["dna.chromosome.{}.fa.gz".format(chromosome)]
+    else:
+        suffixes = ["dna.primary_assembly.fa.gz", "dna.toplevel.fa.gz"]
+elif datatype == "cdna":
+    suffixes = ["cdna.all.fa.gz"]
+elif datatype == "cds":
+    suffixes = ["cds.all.fa.gz"]
+elif datatype == "ncrna":
+    suffixes = ["ncrna.fa.gz"]
+elif datatype == "pep":
+    suffixes = ["pep.all.fa.gz"]
+else:
+    raise ValueError("invalid datatype, must be one of dna, cdna, cds, ncrna, pep")
+
+if chromosome:
+    if not datatype == "dna":
+        raise ValueError(
+            "invalid datatype, to select a single chromosome the datatype must be dna"
+        )
+
+spec = spec.format(build=build, release=release)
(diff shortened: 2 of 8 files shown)


fgvieira:
Is the config file really necessary?
What is the `old_wrapper.py` file?

bio/reference/inphared-db/test/Snakefile
```diff
@@ -0,0 +1,12 @@
+configfile: "config.yaml"
+
+rule get_inphareddb:
+    output:
+        expand("{date}{suffix}", date=config["date"], suffix=config["suffix"])
```
dlaehnemann, line 5: One of the things we try with these wrappers, is to have them work with arbitrary file names. So the `config[]` entries should not be used in this example Snakefile, but rather in the `wrapper.py` file (via `snakemake.params.date`, for example).

```suggestion
        "resources/inphared.fasta"
```

To add the values in the config variables back into the file name, the users of the wrapper should then add python code around it. But I also like the idea of directly showcasing how to do that here. So maybe we could have two versions of calling the wrapper here, one with a fixed file name (like i suggest here), and one that contains the `config[]` entries.

bio/reference/inphared-db/meta.yaml
```diff
@@ -0,0 +1,4 @@
+name: inphared-db
+description: Download sequence file from the Inphared database (https://github.com/RyanCook94/inphared/blob/main/README.md), and store them in a single .fasta file. Please check the current database available at the above link and adjust the config file.
```
dlaehnemann, line 2:
```suggestion
description: Download sequence file from the [inphared database](https://github.com/RyanCook94/inphared/blob/main/README.md), and store them in a single .fasta file. Please check the above link for available database version and adjust the config file. 
```

bio/reference/inphared-db/test/Snakefile
```diff
@@ -0,0 +1,12 @@
+configfile: "config.yaml"
+
+rule get_inphareddb:
+    output:
+        expand("{date}{suffix}", date=config["date"], suffix=config["suffix"])    
+    params:
+        prefix = config["prefix"],
```
dlaehnemann, line 7: Maybe rename this to `url:`, as that might make its purpose a bit clearer?

```suggestion
        url = config["url"], 
```

Obviously, the same rename then applies in the `config.yaml` file.

bio/reference/inphared-db/test/old_release.smk
```diff
@@ -0,0 +1,29 @@
+rule get_genome:
```
dlaehnemann, line 1: This file can be deleted here, right? It's simply a copy-paste leftover, if I understand this correctly.

bio/reference/inphared-db/old_wrapper.py
```diff
@@ -0,0 +1,80 @@
+__author__ = "Johannes Köster"
```
dlaehnemann, line 1: This file can simply be deleted here, right?

bio/reference/inphared-db/test/old_snakefile.smk
```diff
@@ -0,0 +1,30 @@
+rule get_genome:
```
dlaehnemann, line 1: This file can simply be deleted here, right?

bio/reference/inphared-db/wrapper.py
```diff
@@ -0,0 +1,9 @@
+__author__ = "Noriko A. Cassman"
+__copyright__ = "Copyright 2023, Noriko A. Cassman"
+__email__ = "noriko.cassman@gmail.com"
+__license__ = "MIT"
+
+from snakemake.shell import shell
+
+    shell:
+        "curl {params.prefix}{params.date}{params.suffix} -o {params.date}{params.suffix}"
```
dlaehnemann, line 9: Some little things here:
1. You have to reference anything from the snakemake rule via the `snakemake` dict. So for example `snakemake.params.date`.
2. We should only use the full `output` file name here, so that users can put in there whatever they want.
3. We have to call the `shell()` function here, as this is not the body of an actual rule, but rather a plain python script.
4. We have to use the `f""` construct here to use format strings to fill in the variables in this context. Here, we are not dealing with snakemake wildcards, but rather python variables. The `{}` syntax is the same, so this is very confusing...
5. I would manually put in the separator between URL and file name here, just for a slightly clearer structure of the download link. Then, we can remove the trailing`/` in the url in `config.yaml`.

```suggestion
    shell(f"curl {snakemake.params.url}/{snakemake.params.date}{snakemake.params.suffix} -o {snakemake.output}"
```

bio/reference/inphared-db/test/config.yaml
```diff
@@ -0,0 +1,9 @@
+date:
+    "2Jul2023"
+
+suffix:
+    "_refseq_genomes.fa"
+    #"_genomes_excluding_refseq.fa"
+
+prefix:
+    "https://millardlab-inphared.s3.climb.ac.uk/"
```
dlaehnemann, line 9: To keep this in sync with the suggestions elsewhere, we would change this to:
```suggestion
url:
    "https://millardlab-inphared.s3.climb.ac.uk"
```

bio/reference/inphared-db/test/config.yaml
```diff
@@ -0,0 +1,9 @@
+date:
+    "2Jul2023"
+
+suffix:
+    "_refseq_genomes.fa"
```
dlaehnemann, line 5: Is there a smaller reference fasta file (of some small subset of viruses, e.g.), that could be used in the example? This should optimally get executed in the CI tests regularly, so shouldn't download much, if possible.

Also, we should still add an actual test run for this wrapper.

dlaehnemann (changes requested):
This looks good already. But of course, I have a bunch of suggestions. Maybe we can work on those together during the symposium or even the workshop.

Issue #32, opened 2019-11-24
number of threads in bwa mem wrapper


RvV1979:
Hi, apologies for this very basic question but I am wondering how snakemake deals with the threads parameter in the bwa mem wrapper. The wrapper uses this parameter to specify the number of threads used by bwa mem. But when also applying a sort the output is piped to either samtools or picard and this I assume also requires at least one additional thread. 

So my question is whether the this causes a mismatch between the number of threads anticipated by snakemake (i.e. the threads parameter) and the actual threads used (i.e. the thread parameter + 1) and if this could have unwanted effects? 

Thanks

johanneskoester:
Yes, this is kind of right. One should definitely invest some time to find the best combination of threads and model this properly.

christopher-schroeder:
From my experience you can have as much threads fro the sorting as you have for bwa. If you observe the mapping behaviour, you can see it performs its operation chunkwise. You have a time block where bwa uses all available threads. Then a timespan with no cpu usage follows, where data is piped out and new data is retrieved. This is exactly the moment, when you need cpu for sorting as samtools gets data.
There is another third block at the end, where bwa is finished and samtools needs to finish the sorting. But again, bwa is done and samtools may get all the threads. Its a perfect coexistance.

PR #1743, opened 2023-08-16
fix: Bwa mem threads

tdayris:
### Description

Currently, [master/bio/bwa-mem2/mem](https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/bwa-mem2/mem.html) uses more threads than the number requested by user: for `x` threads requested by user, `x` is used in bwa-mem. Then either `x` are used by samtools (view or sort), or `1` by Picard. So for `x` threads requested by user, from `x+1` to `2x` threads are used.

This leads to bwa-mem jobs using official wrappers being killed by cluster manager for not respecting fair use rules where I work. Coworkers asked for a fix.

In this PR, I introduce a function to split the total number of threads between bwa-mem in one hand, and samtools/picard in the other hand.

Ultimately, the introduced changes also allow an uncompressed SAM output.

Tests have been modified, since this wrapper now requires at least two threads when samtools/picard are used. No more when sam output is requested.

### QC

* [X] I confirm that:


fgvieira:
This issue has appeared before in #32 and the time it was concluded it was fine, since `bwa` and `samtools/picard` never run at the same time. Is this not the case for `bwa-mem2`?

tdayris:
`/usr/bin/time -v bwa-mem2 mem -t 4 ... | samtools sort -@ 3 ...` on 10 real human WES datasets.

Should return a percentage of CPU used close-to-but-below 800% if I were right, or close-to-but-below 400% if you were right. My tests shows this number being 470% ± 15%.

Bwa and Samtools do not run with all their workers in the same time. But at one moment, we go over the value of 4 threads and almost reach 5 threads.

My solution is not optimal as it would largely overestimate the number of threads used.

bio/bwa-mem2/mem/test/Snakefile
```diff
@@ -9,9 +9,9 @@ rule bwa_mem2_mem:
         "logs/bwa_mem2/{sample}.log",
     params:
         extra=r"-R '@RG\tID:{sample}\tSM:{sample}'",
-        sort="none",  # Can be 'none', 'samtools' or 'picard'.
+        sort="none",  # Can be 'none', 'samtools', 'picard' or 'sam'.
```
johanneskoester, line 12: The output format should rather be controlled by the extension, as in most of the other wrappers. Hence I would argue that the sam value here is not necessary.

johanneskoester (changes requested):

fgvieira:
> Should return a percentage of CPU used close-to-but-below 800% if I were right, or close-to-but-below 400% if you were right. My tests shows this number being 470% ± 15%.

If so, can't we just give `samtools` one thread less?
So, when running the wrapper with 4 threads, it would run `/usr/bin/time -v bwa-mem2 mem -t 4 ... | samtools sort -@ 2`.

bio/bwa-mem2/mem/test/Snakefile
tdayris: You are right, changes have been made accordingly

tdayris:
New tests on the very same 15 human WES mapping shows 380% ± 15% of CPU usage. I do not get kicked anymore. Thank you for your advises.

johanneskoester (approved):

johanneskoester (approved):

PR #1775, opened 2023-08-31
fix: for vsearch, fixed bug with bz2 output and added general log

fgvieira:
### Description

Fixed bug with `bz2` output and added the general log too keep console clean and be able to keep track of progress.

### QC

* [x] I confirm that:

### DIFF
--- a/bio/vsearch/test/Snakefile
+++ b/bio/vsearch/test/Snakefile
@@ -3,6 +3,7 @@ rule vsearch_cluster_fast:
         cluster_fast="reads/{sample}.fasta",
     output:
         profile="out/cluster_fast/{sample}.profile",
+        log="out/cluster_fast/{sample}.log",
     log:
         "logs/vsearch/cluster_fast/{sample}.log",
     params:
@@ -17,6 +18,7 @@ rule vsearch_maskfasta:
         maskfasta="reads/{sample}.fasta",
     output:
         output="out/maskfasta/{sample}.fasta",
+        log="out/maskfasta/{sample}.log",
     log:
         "logs/vsearch/maskfasta/{sample}.log",
     params:
@@ -31,6 +33,7 @@ rule vsearch_fastx_uniques:
         fastx_uniques="reads/{sample}.fastq",
     output:
         fastqout="out/fastx_uniques/{sample}.fastq",
+        log="out/fastx_uniques/{sample}.log",
     log:
         "logs/vsearch/fastx_uniques/{sample}.log",
     params:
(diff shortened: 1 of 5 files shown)


bio/vsearch/test/Snakefile
```diff
@@ -3,6 +3,7 @@ rule vsearch_cluster_fast:
         cluster_fast="reads/{sample}.fasta",
     output:
         profile="out/cluster_fast/{sample}.profile",
+        log="out/cluster_fast/{sample}.log",
```
johanneskoester, line 6: What is this additional log supposed to contain, how is it different from the other log below? And why is it not under logfiles?

johanneskoester (changes requested):

bio/vsearch/test/Snakefile
johanneskoester: I guess it is only me being uninformed about vsearch, but maybe a clarifying comment should be added.

fgvieira: `vsearch` allows to specify a log file through the command line (current `output.log`) but, to be consistent with the other wrappers and since that log is optional, I kept the `stdout` log too.

I can move `output.log` to `log.some_name`, e.g.:
```
    log:
        "logs/vsearch/cluster_fast/{sample}.log",
        log="out/cluster_fast/{sample}.log",
```
or, when not specified (since it should be an optional log):
```
    log:
        "logs/vsearch/cluster_fast/{sample}.log",
```

Do you think that would make more sense?

bio/vsearch/test/Snakefile
```diff
@@ -12,71 +12,64 @@ rule vsearch_cluster_fast:
         "master/bio/vsearch"
 
 
-rule vsearch_maskfasta:
+use rule vsearch_cluster_fast as vsearch_maskfasta with:
```
johanneskoester, line 15: I would not use `use rule` here. People might want to copy paste from this file, and for that every variant of the rule should work standalone.

johanneskoester (changes requested):

fgvieira:
I thought a bit more about the `log` file and, since it is not exactly the same as the `stdout`, it makes sense to support it. However, I think having it as `verbose` might be a bit confusing since it is not clear where that output comes from.
If we leave it an optional output, then it is clear it comes from the `vsearch` command line argument `log` (just like all the other outputs).

johanneskoester:
Mhm, but it will be lost upon error. log is really the right place for these things. If you don't like verbose, maybe call the two logs `std` and `vsearch`?

fgvieira:
That is true, the outputs are lost in error.

johanneskoester (approved):

PR #1798, opened 2023-09-12
feat: encode fastq downloader

Maarten-vd-Sande:
### Description

Download fastq files directly from the ENCODE project. 

Added python as a dependency because wrapper **needs** an environment: https://github.com/snakemake/snakemake/issues/1718

### QC

* [x] I confirm that:


Maarten-vd-Sande:
@dlaehnemann could you take a look if this is what you had in mind for the seq2science review?

bio/encode_fastq_downloader/test/Snakefile
```diff
@@ -0,0 +1,21 @@
+rule encode_fastq_download_PE:
+    output:
+        "{accession}_R1.fastq.gz",
+        "{accession}_R2.fastq.gz"
```
dlaehnemann, line 4: I think you could work with named outputs here, to avoid relying on the order of specified outputs. Here, I would suggest:

```suggestion
        r1="{accession}_R1.fastq.gz",
        r2="{accession}_R2.fastq.gz"
```

Below, for the single-end case, I would then suggest something like `se=` or `r0=`. Then, you can check for the presence of `r1` and `r2`, or `se`/`r0` in `snakemake.output`.

bio/encode_fastq_downloader/environment.yaml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - conda-forge
+  - bioconda
+  - nodefaults
+dependencies:
+  - snakemake-wrapper-utils >=0.6.2
```
dlaehnemann, line 6: You don't import (let alone use) this in the wrapper.py script, right? So you don't actually need an environment.yaml, I guess?

bio/encode_fastq_downloader/wrapper.py
```diff
@@ -0,0 +1,128 @@
...
+        exception_to_log(
+            check=inferred_layout == "single-ended",
+            msg=f"""The sample was automatically inferred to be single-ended, but it is: "{inferred_layout}".""",
+        )
+        for encff_accession in response["@graph"]:
+            encff_accession = encff_accession["accession"]
+            download_encff(encff_accession, layout, dest)
+    elif layout == "paired":
+        exception_to_log(
+            check=inferred_layout == "single-ended",
```
dlaehnemann, line 94: Copy-pasta, I guess:
```suggestion
            check=inferred_layout == "paired-ended",
```

dlaehnemann (commented):
Yes, this is exactly what I was thinking of. This is really great, many thanks!

I already have some small suggestions, but will have a more thorough tomorrow.

bio/encode_fastq_downloader/environment.yaml
Maarten-vd-Sande: I somehow got the impression the docs woudln't build without the environment.yaml, so I just put something in there for the time being... :upside_down_face: 

I removed the env file, and the docs seem to build

Maarten-vd-Sande: Nvm, `Tests / docs` passes, but `Tests / testing` doesn't :upside_down_face:

dlaehnemann: I think you have successfully identified a corner case, here... :sweat_smile: 

If I understand the `snakemake` code correctly, it will **always** look for an `environment.yaml`:

1. If the rule is a wrapper, it goes looking with `get_conda_env`:
https://github.com/snakemake/snakemake/blob/27b224ed12448df8aebc7d1ff8f25e3bf7622232/snakemake/workflow.py#L1678
2. `get_conda_env()` doesn't do any checking for existence:
https://github.com/snakemake/snakemake/blob/27b224ed12448df8aebc7d1ff8f25e3bf7622232/snakemake/wrapper.py#L62-L67

So I think you are right that the wrapper needs an `environment.yaml`. But we could create one without any packages specified; and even without any channels, I'd say. Can we try that?

Maarten-vd-Sande: Empty environment.yaml also doesn't work

Maarten-vd-Sande: I decided to specify a minimum python version as this is usually the case anyways

bio/encode_fastq_downloader/environment.yaml
```diff
@@ -0,0 +1,10 @@
+channels:
+  - conda-forge
+  - bioconda
+  - nodefaults
+# there are strictly no dependencies for this wrapper, but
+# the tests require a environment.yaml with 
+# dependencies. So this is just to please the tests :)
```
dlaehnemann, line 7: I would explicitly mention the issue over at the snakemake repo, so we can find it again in the future:
```suggestion
# There are strictly no dependencies for this wrapper, but the current handling
# of wrappers by snakemake requires an environment.yaml with dependencies.
# Once the underlying issue is fixed, this environment.yaml can be removed. See:
# https://github.com/snakemake/snakemake/issues/1718
```

We should also mention this issue either in the commit message of this change, or somewhere in the pull request. That way we get cross-references on GitHub.

bio/encode_fastq_downloader/test/Snakefile
Maarten-vd-Sande: I'm not sure what to call the single-end named outputs. Based on Illumina's explanation I'm thinking to just name them r1, and r1 + r2 for paired-end: https://knowledge.illumina.com/software/general/software-general-reference_material-list/000002211

> For a single-read run, one Read 1 (R1) FASTQ file is created for each sample per flow cell lane. For a paired-end run, one R1 and one Read 2 (R2) FASTQ file is created for each sample for each lane. FASTQ files are compressed and created with the extension *.fastq.gz.

Not entirely convinced of this, but perhaps better than thinking of our own format?

dlaehnemann: You are right, from the general logic of Illumina, single-end reads simply sequence `read1`. `se` and `r0` are just things I have seen in the wild and never questioned. But it sounds better and more consistent to go for your suggestion:
* `r1=` only for single-end
* `r1=` and `r2=` for paired-end

And you can still clearly introduce logic for handling both cases (and throwing an error if neither is the case). Thanks for being so thorough!

johanneskoester (approved):

PR #1822, opened 2023-09-26
feat: Gseapy

tdayris:
### Description

This PR adds `gseapy` to the list of available wrappers. I plan to add replot, dotplots, heatmaps, and barplots functions in the future.

### QC

* [X] I confirm that:


bio/gseapy/gsea/test/Snakefile
```diff
@@ -0,0 +1,69 @@
+rule test_gseapy_enrichr:
+    input:
+        gene_list="genes_list.txt",
+        # gmt="", Optional path to a GMT file
+    output:
+        directory("results_enrichr"),
```
johanneskoester, line 6: I think it is preferable to have the individual output files here. Or are they unpredictable/of unknown multiplicity?

johanneskoester (changes requested):

bio/gseapy/gsea/test/Snakefile
tdayris: In this case, we have both scenarios.

There a three known and predictable files which can be tracked. Depending on user parameters and GSEApy results, there is an unknown number of files named after the original prefix defines by user (parameters), the database defined by user (parameters) and the pathways significantly enriched (tool results).

Downstream analysis is based on that directory architecture. We can reproduce it later.

Let next commit be my attempt at naming as much results as possible.

johanneskoester (approved):

PR #1873, opened 2023-10-13
fix: Update environment.yaml of MultiQC

KuechlerO:
### Description

Python version is fixed to versions<3.12.0 to handle dependency on imp-module, which is not available in newer python versions. 
The imp module is [deprecated in favor of importlib, and removed from the standard library since python version 3.12.0](https://docs.python.org/3.11/library/imp.html). Thus, depending on the python version your program-call/conda-environment is using, you will encounter a `ModuleNotFoundError` Error.

For further information, see discussion on StackOverflow:  https://stackoverflow.com/questions/77273344/multiqc-snakemake-wrapper-modulenotfounderror-no-module-named-imp

### DIFF
--- a/bio/multiqc/environment.yaml
+++ b/bio/multiqc/environment.yaml
@@ -3,5 +3,4 @@ channels:
   - bioconda
   - nodefaults
 dependencies:
-  - python >=3.7.0,<3.12.0
-  - multiqc =1.16
+  - multiqc =1.17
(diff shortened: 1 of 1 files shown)


fgvieira:
This issue is currently being fixed by MutiQC ewels/multiqc#2112 and is expected to be fixed Monday, so it might be better to wait a bit.

bio/multiqc/environment.yaml
```diff
@@ -3,4 +3,5 @@ channels:
   - bioconda
   - nodefaults
 dependencies:
+  - python >=3.7.0,<3.12.0
   - multiqc =1.16
```
fgvieira, line 7:
```suggestion
  - multiqc =1.17
```

fgvieira (approved):

PR #1880, opened 2023-10-19
feat: cnvkit export wrapper

Smeds:
### Description

wrapper for cnvkit export

### QC

* [x] I confirm that:


bio/cnvkit/export/meta.yaml
```diff
@@ -0,0 +1,12 @@
+name: "cnvkit export"
+description: >
+  Convert copy number ratio tables (.cnr files) or segments (.cns) to another format..
+url: https://cnvkit.readthedocs.io/en/stable/pipeline.html
```
Smeds, line 4:
```suggestion
url: https://cnvkit.readthedocs.io/en/stable/importexport.html#export
```

bio/cnvkit/export/meta.yaml
```diff
@@ -0,0 +1,12 @@
+name: "cnvkit export"
+description: >
+  Convert copy number ratio tables (.cnr files) or segments (.cns) to another format..
```
fgvieira, line 3:
```suggestion
  Convert copy number ratio tables (.cnr files) or segments (.cns) to another format.
```

bio/cnvkit/export/wrapper.py
```diff
@@ -0,0 +1,21 @@
...
+__license__ = "MIT"
+
+from snakemake.shell import shell
+from os.path import splitext
+
+log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+
+extra = snakemake.params.get("extra", "")
+
+file_extension = splitext(snakemake.output[0])[1][1:]
```
fgvieira, line 13: Would this also work with a `vcf.gz`?

test.py
```diff
@@ -1180,6 +1180,49 @@ def test_bwa_mapping_meta():
         ],
     )
 
+@skip_if_not_modified
+def test_cnvkit_target():
```
fgvieira, line 1184:
```suggestion
def test_cnvkit_export():
```

bio/cnvkit/export/wrapper.py
Smeds: I will do a small test to see if the software can output a compressed vcf

Smeds: The tool didn't support vcf.gz as output! But I have update the wrapper to support ".gz" as output.

bio/cnvkit/export/meta.yaml
```diff
@@ -0,0 +1,12 @@
+name: "cnvkit export"
+description: >
+  Convert copy number ratio tables (.cnr files) or segments (.cns) to another format.
+url: https://cnvkit.readthedocs.io/en/stable/importexport.html#export
+input:
+  - cns: cns or cnr file(s)
+output:
+  - bed/vcf/cdt/seq file
```
fgvieira, line 8:
```suggestion
  - bed/vcf/vcf.gz/cdt/seq file
```

fgvieira (approved):

PR #2382, opened 2023-11-28
fix: explicitly add python2.7 requirement for manta wrapper to avoid python 1.6 being pulled

ipstone:
### Description

The current manta wrapper will result in pulling python 1.6 into the conda environment when using --use-conda for snakemake runs.

This update just added `python=2.7` requirement to the manta wrapper requirement, which fixed the issue on my system.

Adding python2.7 requirement for manta wrapper fixed the problem of solver would pull python 1.6 when using manta wrapper.


ipstone:
@fgvieira based on these source codes: https://github.com/Illumina/manta/blob/master/src/python/bin/configManta.py

it seems that manta only supports python2 for the scripts setup/calling the binaries. so I guess python=2.7 is safe, though I didn't specifically test if python3 works.

bio/manta/environment.yaml
```diff
@@ -3,5 +3,6 @@ channels:
   - bioconda
   - nodefaults
 dependencies:
+  - python =2.7
```
fgvieira, line 6:
```suggestion
  - python =2.7.15
```

bio/manta/environment.yaml
```diff
@@ -3,5 +3,6 @@ channels:
...
   - manta =1.6.0
   - bcftools =1.17
```
fgvieira, line 8:
```suggestion
  - bcftools =1.18
```

fgvieira:
Maybe the best way would be to create a new `environment.linux-64.pin.txt` file pinning all packages in the environment.

PR #2640, opened 2024-02-04
Add Orthanq wrapper

huzuner:
### QC

* [ ] I confirm that:


bio/orthanq/hla/candidates/environment.yaml
```diff
@@ -0,0 +1,2 @@
+dependencies:
```
fgvieira, line 1:
```suggestion
channels:
  - conda-forge
  - bioconda
  - nodefaults
dependencies:
```

bio/orthanq/virus/candidates/environment.yaml
```diff
@@ -0,0 +1,2 @@
+dependencies:
```
fgvieira, line 1:
```suggestion
channels:
  - conda-forge
  - bioconda
  - nodefaults
dependencies:
```

bio/orthanq/hla/candidates/meta.yaml
```diff
@@ -0,0 +1,7 @@
+name: orthanq candidates hla
+description: |
+  Orthanq candidates hla subcommand generates candidate variants required for HLA quantification.
+  All required input is downloaded by following the instructions here: https://orthanq.github.io/docs/usage.html#preparation
+url: https://github.com/orthanq/orthanq
+authors:
+  - Hamdiye Uzuner
```
fgvieira, line 7: It would be nice to have some info on the `input` and `output` files.

bio/orthanq/virus/candidates/meta.yaml
```diff
@@ -0,0 +1,6 @@
+name: orthanq candidates virus
+description: |
+  Orthanq candidates virus subcommand generates candidate variants required for viral lineage quantification.
+url: https://github.com/orthanq/orthanq
+authors:
+  - Hamdiye Uzuner
```
fgvieira, line 6: It would be nice to have some info on the input and output files.

bio/orthanq/virus/candidates/test/Snakefile
```diff
@@ -0,0 +1,7 @@
+rule orthanq_candidates_virus:
+    output:
+        directory("viral_candidates"),
```
fgvieira, line 3: What files are created in this directory? Would it be possible to name them?

fgvieira:
Don't have any experience with `orthanq` but, from the command line, it seems that it could make more sense to have `candidates\hla` and `cadidates\virus`, no?

It seems `candidates\hla` uses a temp folder; can you use the `tmpdir` resource?

bio/orthanq/virus/candidates/meta.yaml
huzuner: The virus part is currently in WIP.

bio/orthanq/virus/candidates/test/Snakefile
huzuner: Many files are created in this folder and it's not possible to name them all here. That was one reason I implemented the output of the tool as a folder. If there is any improvement regarding the tool output, I'll update it here aswell :)

bio/orthanq/virus/candidates/environment.yaml
huzuner: thanks for the input :)

bio/orthanq/hla/candidates/environment.yaml
huzuner: thanks for the input :)

fgvieira:
Hi @huzuner, did you have the time to look into the comments above?

huzuner:
> From what I can see in the docs, it seems that the tool has three subcommands and `hla`/`virus` is an option. If so, I'd organize the wrapper the same way as (e.g.) `samtools` (`orthanq/candidates`, `orthanq/preprocess`, and `orthanq/call`) with `hla`/`virus` as a parameter. However, since `orthanq` is developed by @johanneskoester lab, I think it would be nice to hear from him.
> 
> If the program supports a `temp` folder, then it would be nice if the wrappers would take care of it. For example, as in:
> 
> https://github.com/snakemake/snakemake-wrappers/blob/f0b638a55280c0727ccef5aab24e03b3d565f290/bio/samtools/sort/wrapper.py#L25-L30

I thought about your suggestion but I couldn't be sure whether that would be a better way of organizing it or not. My idea initially was that because the tool has two applications, hla and virus, I wanted to organize the folders in that manner in order to make this more prominent, but your idea also does not seem wrong. In the end I guess I'd rather have it as it is now if that's not really a major issue. We can of course always reorganize it later.

In addition, the tool does not have a temp option at the moment.

bio/orthanq/hla/call/test/Snakefile
```diff
@@ -0,0 +1,12 @@
+rule orthanq_call_hla:
+    input:
+        haplotype_calls="haplotype_calls.bcf",
+        haplotype_variants="haplotype_variants.vcf",
+        xml="hla.xml"
+    output:
+        "calls.csv",
+    params: prior="diploid"
```
fgvieira: Maybe use the extra for both `prior` and `sample-name`:
```suggestion
    params:
        extra="--prior diploid",
```

bio/orthanq/hla/candidates/wrapper.py
```diff
@@ -0,0 +1,16 @@
+__author__ = "Hamdiye Uzuner"
+__copyright__ = "Copyright 2024, Hamdiye Uzuner"
+__license__ = "MIT"
+
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+# Optional flag for outputting bcf instead of vcf.
+output_bcf = snakemake.params.get("output_bcf", False)
```
fgvieira: Maybe check the extension from the output file, instead of using a `param`. You can use functions from `snakemake-wrapper-utils`.

bio/orthanq/virus/candidates/wrapper.py
```diff
@@ -0,0 +1,12 @@
...
+__copyright__ = "Copyright 2024, Hamdiye Uzuner"
+__license__ = "MIT"
+
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+shell(
+        "orthanq candidates virus --genome {snakemake.input.genome} --lineages "
+        " {snakemake.input.lineages} --output {snakemake.output[0]} 2> {log}"
```
fgvieira:
```suggestion
        " {snakemake.input.lineages} --output {snakemake.output[0]} {log}"
```

fgvieira:
> I thought about your suggestion but I couldn't be sure whether that would be a better way of organizing it or not. My idea initially was that because the tool has two applications, hla and virus, I wanted to organize the folders in that manner in order to make this more prominent, but your idea also does not seem wrong. In the end I guess I'd rather have it as it is now if that's not really a major issue. We can of course always reorganize it later.

I think the main advantage is that we'd only need half the wrappers, since they are quite similar.

### Call
```
orthanq call hla --haplotype-calls {snakemake.input.haplotype_calls} --haplotype-variants {snakemake.input.haplotype_variants} --prior {prior_opt} --xml {snakemake.input.xml} --output {snakemake.output[0]} {sample_name_opt} {log}

orthanq call virus --haplotype-calls {snakemake.input.haplotype_calls} --haplotype_variants {snakemake.input.candidates_vcf} --prior {prior_opt} --output {snakemake.output[0]} {log}
```

### Candidates
```
orthanq candidates hla --genome {snakemake.input.genome} --alleles {snakemake.input.alleles} --xml {snakemake.input.xml} {output_bcf_flag} --output {snakemake.output[0]} {log}

orthanq candidates virus --genome {snakemake.input.genome} --lineages {snakemake.input.lineages} --output {snakemake.output[0]} {log}
```

### Preprocess
```
orthanq preprocess virus --genome {snakemake.input.genome} --candidates {snakemake.input.candidates_vcf} --reads {snakemake.input.reads} --output {snakemake.output[0]} {log}

orthanq preprocess hla --genome {snakemake.input.genome} --haplotype-variants {snakemake.input.haplotype_variants} --reads {snakemake.input.reads} --vg-index {snakemake.input.vg_index} --output {snakemake.output[0]} {log}
```

bio/orthanq/hla/call/test/Snakefile
huzuner: I'd rather not have this in extra, since it's a required parameter.

bio/orthanq/hla/candidates/wrapper.py
huzuner: Output format cannot be determined from the output because the output is provided as a folder instead of a file for hla application
.

huzuner:
Can I get another review?

fgvieira:
Looks quite good! Almost there:
- add tests to [test_wrappers.py](https://github.com/snakemake/snakemake-wrappers/blob/master/test_wrappers.py)
- format wrapper with `black`

bio/orthanq/test/Snakefile
```diff
@@ -0,0 +1,98 @@
+# Candidates HLA
+rule orthanq_candidates_hla:
+    input:
+        genome="genome_hla.fasta",
+        xml="hla.xml",
+        alleles="hla_alleles.fasta",
+    output:
+        directory("hla_candidates")
```
fgvieira, line 8:
```suggestion

```

fgvieira:
you can organize by type (e.g.):
- `reads/`
- `refs/`
- `variants/`

bio/orthanq/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+outputs:
+  - Quantified haplotypes in CSV (from `call`)
+  - Preprocessed BCF file (from `preprocess`)
+  - Candidate variants in VCF or BCF format (from `candidates`)
+
+params:
+  - command: One of `candidates`, `preprocess`, or `call`
+  - subcommand: One of `hla` or `virus`
+  - prior: Prior model used in Bayesian quantification (`uniform` or `diploid`) - required for `call`
+  - sample_name: Sample name (optional; only used in `call hla` for multi-sample VCFs)
```
fgvieira: Can't these also be specified in `params.extra`?

huzuner: prior is necessary but sample_name can go to extra

huzuner:
The tests should now work fine now but I now also need to configure the candidates_hla further so that we make sure it generates the vcfs in the following way:
rule orthanq_candidate_variants:
```
    input:
        hla_genes="results/preparation/hla_gen.fasta",
        xml="results/preparation/hla.xml",
        genome=genome,
    output:
        bcfs=expand(
            "results/orthanq-candidate-calls/{caller}.hla-variants.bcf",
            caller=[f"orthanq-{locus}" for locus in config["hla_typing"].get("loci")],
        ),
    log:
        "logs/orthanq-candidates.log",
    conda:
        "../envs/orthanq.yaml"
    params:
        output_folder=lambda wc, output: os.path.dirname(output.bcfs[0]),
    threads: 8
    shell:
        "orthanq candidates hla --alleles {input.hla_genes} --genome {input.genome} --xml {input.xml} "
        "--threads {threads} --output-bcf --output {params.output_folder} 2> {log}"
 ```
Even though the tool supports only directory input (because we don't know what and how many of the HLA loci are requested) e.g. A.vcf, B.vcf ..etc.

File: test_wrappers.py [hunk 11]
```diff
@@ -7090,3 +7092,80 @@ def test_rasterio_clip_geotiff(run):
...
+    run(
+        "bio/orthanq",
+        [
+            "snakemake",
+            "--cores",
+            "1",
+            "out/calls_virus.csv",
+            "--use-conda",
+        ],
+    )
```
fgvieira, line 7112:
```suggestion


def test_orthanq(run):
    run(
        "bio/orthanq",
        [
            "snakemake",
            "--cores",
            "1",
            "--use-conda",
            "out/candidates",
            "out/candidates.vcf",
#             "out/preprocess_hla.bcf",
            "out/preprocess_virus.bcf",
            "out/calls_hla.csv",
            "out/calls_virus.csv",
        ],
    )
```

bio/orthanq/test/Snakefile
```diff
@@ -0,0 +1,98 @@
+# Candidates HLA
+rule orthanq_candidates_hla:
+    input:
+        genome="refs/genome_hla.fasta",
+        xml="refs/alleles.xml",
+        alleles="refs/alleles.fasta",
+    output:
+        candidates_dir=directory("out/candidates")
```
fgvieira: Unnecessary naming.
```suggestion
        directory("out/candidates")
```

bio/orthanq/wrapper.py
```diff
@@ -0,0 +1,77 @@
...
+genome = snakemake.input.get("genome", "")
+extra = snakemake.params.get("extra", "")
+
+cmd = f"orthanq {command} {subcommand}"
+
+# Determine output depending on command/subcommand
+if command == "candidates" and subcommand == "hla":
+    output = snakemake.output.candidates_dir
+else:
+    output = snakemake.output[0]
```
fgvieira:
```suggestion
```

bio/orthanq/wrapper.py
```diff
@@ -0,0 +1,77 @@
...
+    elif subcommand == "virus":
+        pass  # already complete
+    else:
+        raise ValueError(f"Unsupported subcommand for 'call': {subcommand}")
+
+else:
+    raise ValueError(f"Unsupported command: {command}")
+
+# Finalize with output and log
+cmd += f" --output {output} {log}"
```
fgvieira, line 75:
```suggestion
cmd += f" --output {snakemake.output[0]} {log}"
```

bio/orthanq/test/Snakefile
huzuner: This is indeed necessary because this subcommand requires a directory to be provided but it must also ensure the VCF files are created in another output. I tried to explain this above.

huzuner: *not in another output but inside the given directory.

fgvieira: OK, then try to implement the BCF logic and request another review.

fgvieira:
And why is test `preprocess_hla` commented?

bio/orthanq/test/Snakefile
huzuner: The candidates hla command currently generates VCF/BCF files for four HLA alleles (A.vcf, B.vcf, C.vcf, and DQB1.vcf). However, this set of loci may expand in the near future. Additionally, the pipeline also produces an intermediate BAM file, which is sometimes necessary for debugging.

To accommodate these growing and flexible output requirements, I've renamed the output (candidates_dir) rather than listing individual files. This ensures we can handle a variable number of loci and any supplementary files without constantly updating the rule outputs.

Here’s an example of how the rule is now structured:

```
rule orthanq_candidate_variants:
    input:
        ...
    output:
        candidates_dir="results/orthanq-candidate-calls",
        bcfs=expand(
            "results/orthanq-candidate-calls/{locus}.vcf",
            locus=["A", "B", "C", "DQB1"],
        )
    log:
        ...
    conda:
        "../envs/orthanq.yaml"
    shell:
        "orthanq candidates hla --alleles {input.hla_genes} --genome {input.genome} "
        "--xml {input.xml} ---output {output.candidates_dir} 2> {log}"
```
This setup provides more flexibility and robustness for future changes, especially when integrating the rule into larger workflows.

fgvieira: If I understand correctly, you say that all output goes in `results/orthanq-candidate-calls`, but that you also want to add output files that are inside that folder (BCF and BAM files).

#### Solution 1
If so, then the easy solution (but not very flexible) is to specify all output files and infer the output directory:
```
rule orthanq_candidate_variants:
    input:
        ...
    output:
        expand(
            "results/orthanq-candidate-calls/{locus}.vcf",
            locus=["A", "B", "C", "DQB1"],
        ),
        "results/orthanq-candidate-calls/output.bam"
```

and in the `wrapper.py`:
```
import os
output = os.path.commonprefix(snakemake.output)
```

this requires that all outputs have the same prefix and the user needs to know the name of the files.

#### Solution 2
Use a temporary directory as `output`, and copy the files that the rule requires to their final destination.
```
rule orthanq_candidate_variants:
    input:
        ...
    output:
        vcf=expand(
            "results/orthanq-candidate-calls/{locus}.vcf",
            locus=["A", "B", "C", "DQB1"],
        ),
        bam="results/orthanq-candidate-calls/output.bam"
```

You can look at wrapper `bio/tmb/pyeffgenomesize/wrapper.py` for inspiration.

fgvieira: A couple of ideas, since you are the main developer of orthanq. :smile: 
Instead of having a vcf/bcf per locus, could you output a single vcf/bcf where, either:
- add an INFO tag with the locus name. you can filter for each locus with `bcftools filter -i INFO/locus = "A"`
- or, also output a BED file with the coordinates of each locus. you can filter for each locus with `bcfools filter -R locus.bed`

huzuner: I gave this some more thought and decided to stick with using the unnamed output directory. Debugging really shouldn’t be happening at the wrapper level anyway.

I also considered adding parameters to select specific HLA genes for output, but that didn’t seem to make much sense either, since we typically need all of them.

Sorry for the back-and-forth on this. Let’s just move forward with the unnamed output as before.

huzuner: > A couple of ideas, since you are the main developer of orthanq. 😄 Instead of having a vcf/bcf per locus, could you output a single vcf/bcf where, either:
> 
>     * add an INFO tag with the locus name. you can filter for each locus with `bcftools filter -i INFO/locus = "A"`
> 
>     * or, also output a BED file with the coordinates of each locus. you can filter for each locus with `bcfools filter -R locus.bed`

Having separate VCFs per locus isn’t just an output formatting choice—it’s actually tied to how Orthanq processes the data, as it considers all variations within a given locus together. Changing this would also impact how the model itself functions. That said, it's a valid point and definitely worth thinking about. Thanks for bringing it up.

bio/orthanq/wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+
+shell(cmd)
```
fgvieira: Just a suggestion (not tested) to simplify the wrapper, mostly by:
- removing checks, since some are expected to be done by the program (e.g. if command or subcommand are valid), and some are already performed by `snakemake` (e.g. mandatory params).
- simplify the command line, leaving it up to the user to decide which input files are valid for each command/subcommand
- 

```suggestion
extra = snakemake.params.get("extra", "")


# Parse inputs
for key, value in snakemake.input.items():
    key = key.replace("_", "-")
    if value:
        extra += f" --{key} {value}"


# Parse params
if command == "call":
    extra += f" --prior {snakemake.params.prior}"


# Finalize with output and log
shell("orthanq {snakemake.params.command} {snakemake.params.subcommand} {extra} --output {snakemake.output[0]} {log}")
```

bio/orthanq/wrapper.py
```diff
@@ -0,0 +1,56 @@
...
+__license__ = "MIT"
+
+from snakemake.shell import shell
+
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+extra = snakemake.params.get("extra", "")
+
+# Required parameters
+command = snakemake.params.command  # candidates, preprocess, call
+subcommand = snakemake.params.subcommand  # hla, virus
```
fgvieira:
```suggestion

```

bio/orthanq/wrapper.py
```diff
@@ -0,0 +1,56 @@
...
+if haplotype_calls:
+    extra += f" --haplotype-calls {haplotype_calls}"
+
+vg_index = snakemake.input.get("vg_index")
+if vg_index:
+    extra += f" --vg-index {vg_index}"
+
+
+# Parse params
+if command == "call":
```
fgvieira:
```suggestion
if snakemake.params.command == "call":
```

huzuner: oh, right.

PR #2649, opened 2024-02-15
feat: Add configurable protocol support to ensembl reference download

pettyalex:
This PR adds support for configurable protocol, ftp, http, or https to ensembl reference data download and vep cache download.

I was evaluating snakemake-wrapper and widely used workflows as a tool for my group to potentially use, and immediately ran into our firewall rules. By default we're able to make outgoing HTTP and HTTPS requests, but not FTP. I can request a change to our firewall, but it also would be helpful if these rules that download reference files were able to download them over HTTP, as ftp.ensembl.org has has been available over HTTP for a very long time.

If you'd prefer, I can create an issue to track this. I also intend to update test cases on these wrappers, and if you are happy with this pattern I could apply it across all wrappers in this repository that are currently hard-coded to make ftp requests.

### QC

* [ ] I confirm that:

### DIFF
--- a/bio/reference/ensembl-annotation/wrapper.py
+++ b/bio/reference/ensembl-annotation/wrapper.py
@@ -19,6 +19,7 @@ gtf_release = release
 out_fmt = Path(snakemake.output[0]).suffixes
 out_gz = (out_fmt.pop() and True) if out_fmt[-1] == ".gz" else False
 out_fmt = out_fmt.pop().lstrip(".")
+protocol = snakemake.params.get("protocol", "ftp")
 
 
 branch = ""
@@ -48,7 +49,7 @@ else:
     )
 
 
-url = "ftp://ftp.ensembl.org/pub/{branch}release-{release}/{out_fmt}/{species}/{species_cap}.{build}.{gtf_release}.{flavor}{suffix}".format(
+url = "{protocol}://ftp.ensembl.org/pub/{branch}release-{release}/{out_fmt}/{species}/{species_cap}.{build}.{gtf_release}.{flavor}{suffix}".format(
     release=release,
     gtf_release=gtf_release,
     build=build,
@@ -58,6 +59,7 @@ url = "ftp://ftp.ensembl.org/pub/{branch}release-{release}/{out_fmt}/{species}/{
     suffix=suffix,
     flavor=flavor,
     branch=branch,
+    protocol=protocol,
 )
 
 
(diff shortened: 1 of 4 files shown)


fgvieira:
Thanks for your contribution.
Can you add a test case for (e.g.) HTTP and HTTPS?

pettyalex:
I've run snakefmt, black and the linter over these changes. I believe this to be ready to merge.

pettyalex:
@fgvieira I fixed all the tests, and now see that what I had previously done introduced multiple ways to generate the desired output files.

I also unintentionally ran black over `test.py`, which it seems like is not being done? Should I revert running black over test.py, or should we have it running through black as well?

test.py
```diff
@@ -5801,11 +5843,28 @@ def test_vep_cache():
     )
 
 
+@skip_if_not_modified
+def test_vep_cache_https_protocol():
+    run(
+        "bio/vep/cache",
+        ["snakemake", "--cores", "1", "resources/vep/cache", "--use-conda", "-F"],
```
fgvieira, line 5850:
```suggestion
        ["snakemake", "--cores", "1", "resources/vep/cache", "--use-conda", "-F", "-s", "vep_cache_https.smk"],
```

test.py
```diff
@@ -5801,11 +5843,28 @@ def test_vep_cache():
...
-        ["snakemake", "--cores", "1", "resources/vep/plugins", "--use-conda", "-F"],
+        [
+            "snakemake",
+            "--cores",
+            "1",
+            "resources/vep/plugins",
+            "--use-conda",
+            "-F",
+            "-s",
+            "vep_cache_https.smk",
```
fgvieira, line 5866:
```suggestion
```

test.py
```diff
@@ -5428,17 +5461,26 @@ def test_ensembl_variation_with_contig_lengths():
 
 
 @skip_if_not_modified
-def test_ega_fetch():
+def test_ensembl_variation_old_release():
```
fgvieira, line 5464:
```suggestion
def test_ensembl_variation_old_release_https_protocol():
```

johanneskoester (changes requested):
Fantastic, thanks! Can you please also add the respective protocol lines to each Snakefile in the test folder? Those are used to generate the wrapper docs, from which people can copy paste into their workflows.

fgvieira:
PR #2928 allowed for custom URLs when downloading from VEP cache. Maybe something similar can be done here for `ensembl` `annotation`/`variation`/`sequence`?

fgvieira:
Can you fix the conflicts?

PR #2719, opened 2024-03-02
fix: deeptools bamcoverage fix

jeffhsu3:
Fixes a small issue when including the genome and read_length pulling from the pre-computed effective genome sizes.  The returned genome size is an int rather than a string.  

### QC

* [x] I confirm that:


fgvieira:
Can you add the new test to `test.py`?

bio/deeptools/bamcoverage/wrapper.py
```diff
@@ -74,7 +74,9 @@
     read_length = snakemake.params.get("read_length")
     if genome and read_length:
         effective_genome_size = "--effectiveGenomeSize "
-        effective_genome_size += default_effective_genome_size[genome][str(read_length)]
+        effective_genome_size += str(
+            default_effective_genome_size[genome][str(read_length)]
+        )
```
fgvieira, line 79:
```
effective_genome_size = "--effectiveGenomeSize {}".format(default_effective_genome_size[genome][str(read_length)])
```

jeffhsu3: To handle cases where `genome` or `read_length` are not found in default_effective_genome_size, I've wrapped the line in a try-except block with a message on the available `genome` and `read_length` table from deeptools.  Also updated the values from the most recent deeptools and made the suggested changes to have it be more readable.

bio/deeptools/bamcoverage/wrapper.py
```diff
@@ -73,10 +105,17 @@
...
+        except KeyError:
+            raise KeyError(
+                (
+                    "A valid genome and read_length need to be provided. See "
+                    "https://deeptools.readthedocs.io/en/latest/content/feature/effectiveGenomeSize.html"
+                )
+            )
 else:
-    effective_genome_size = "--effectiveGenomeSize " + str(effective_genome_size)
+    effective_genome_size = f"--effectiveGenomeSize {str(effective_genome_size)}"
```
fgvieira, line 118:
```suggestion
    effective_genome_size = f"--effectiveGenomeSize {effective_genome_size}"
```

fgvieira (approved):

PR #2722, opened 2024-03-06
feat: MultiQC automatically detects configuration from input file(s)

tdayris:
This PR lets MultiQC automatically detect configuration file, when user provides them in input, as described in [official documentation](https://multiqc.info/docs/getting_started/config/).

### QC

* [X] I confirm that:


bio/multiqc/wrapper.py
```diff
@@ -26,6 +26,15 @@
     input_data = set(snakemake.input)
 
 
+# Automatically detect configuration files when provided
+# in input. For other ways to provide configuration to
+# multiqc, see: https://multiqc.info/docs/getting_started/config/
+mqc_config = ""
+for fp in set(snakemake.input):
+    if str(fp).lower().endswith("multiqc_config.yaml"):
```
fgvieira, line 34: Does the file name need to be hard-coded? Could we use a named input?
And maybe also accepting a list, in case the user has several config files.

tdayris:
> Is the new test being triggered in `test.py`?

It was unstashed, my fault. It's in now. Thanks !

bio/multiqc/wrapper.py
tdayris: Well, in that case, they have to be named in the snakefile. Would a `config`-key in `input` suits the job?

fgvieira: I was thinking something along those lines.

bio/multiqc/test/Snakefile
```diff
@@ -25,3 +25,18 @@ rule multiqc_file:
         "logs/multiqc.log",
     wrapper:
         "master/bio/multiqc"
+
+
+rule test_multiqc_config:
+    input:
+        expand("samtools_stats/{sample}.txt", sample=["a", "b"]),
+        "config/multiqc_config.yaml",
```
fgvieira, line 33:
```suggestion
        config="config/multiqc_config.yaml",
```

bio/multiqc/wrapper.py
```diff
@@ -26,6 +26,15 @@
     input_data = set(snakemake.input)
 
 
+# Automatically detect configuration files when provided
+# in input. For other ways to provide configuration to
+# multiqc, see: https://multiqc.info/docs/getting_started/config/
+mqc_config = snakemake.input.get("config", "")
```
fgvieira, line 32: I guess we need to account for the case when it is not a list, no?

tdayris: I thought Snakemake was already performing these checks, but you're right, it should be explicitly handled.

bio/multiqc/wrapper.py
```diff
@@ -26,6 +26,17 @@
...
+if isinstance(mqc_config, list):
+    for fp in mqc_config:
+        extra += f" --config {fp} "
+elif mqc_config:
+    extra += f" --config {mqc_config} "
```
fgvieira, line 37:
```suggestion
        extra += f" --config {fp}"
elif mqc_config:
    extra += f" --config {mqc_config}"
```

fgvieira:
Can it be a problem that the config will also be included as input?
https://github.com/snakemake/snakemake-wrappers/blob/7611d30950b85360cb91ff2b58895ae788e05c68/bio/multiqc/wrapper.py#L22-L26

tdayris:
> Can it be a problem that the config will also be included as input?
> 
> https://github.com/snakemake/snakemake-wrappers/blob/7611d30950b85360cb91ff2b58895ae788e05c68/bio/multiqc/wrapper.py#L22-L26

No, it does not. However I used `pop` instead of `get` method to remove the key `config` if it exists, and moved the block above the `input_data` bloc.

tdayris:
Having config in input does not raise any error. However not having them will speed-up the process.

PR #2725, opened 2024-03-07
feat: add aria2c wrapper

fgvieira:
Add wrapper for aria2c, since it allows (among others):
- download of several protocols (e.g.  HTTP/HTTPS, FTP, SFTP, BitTorrent and Metalink)
- parallel downloads
- automated checksum check
- pre-allocate disk space 

### QC

* [x] I confirm that:

## Summary by CodeRabbit

- **New Features**
  - Introduced a new wrapper for the aria2c download utility, supporting multiple checksum types (MD5, SHA-1, SHA-224, SHA-256, SHA-384, SHA-512, Adler32) for file integrity verification.
  - Added a comprehensive test suite for the aria2c wrapper, including sample checksum files and Snakemake rules to validate downloads with various hash algorithms.
  - Provided environment and metadata files to ensure reproducible setups and clear tool documentation.

- **Tests**
  - Implemented automated tests to verify the aria2c wrapper's functionality and checksum verification.

- **Chores**
  - Updated GitHub Actions workflow to include an additional storage plugin for Snakemake.


utils/aria2c/meta.yaml
```diff
@@ -0,0 +1,17 @@
+name: aria2
+url: https://github.com/aria2/aria2/
+description: >
+  aria2 is a lightweight multi-protocol & multi-source, cross platform download utility operated in command-line. It supports HTTP/HTTPS, FTP, SFTP, BitTorrent and Metalink.
+authors:
+  - Filipe G. Vieira
+input:
+  - file with checksum (optional)
```
johanneskoester, line 8: why is this a file, and not just a param?

fgvieira: To use a param, we need to know the checksum beforehand, but with an input file it can be dynamic (you can have another rule downloading it). Checksums are usually stored in files in the same folder of the file to download. For example, for the [RefSeq donkey genome](https://ftp.ncbi.nlm.nih.gov/genomes/refseq/vertebrate_mammalian/Equus_asinus/representative/GCF_016077325.2_ASM1607732v2/), the genome is `GCF_016077325.2_ASM1607732v2_genomic.fna.gz` but the checksum is inside the file `md5checksums.txt`.

The issue is similar to snakemake/snakemake#1972.

johanneskoester: Ok, makes sense in this case. But if the checksum is not present somewhere, things become more cumbersome as one first has to write a file. Suggestion, make it a param, and use the fact that params can read input files:

```
rule gtdb_taxonomy:
    input:
        storage.http("https://data.gtdb.ecogenomic.org/releases/release220/220.0/MD5SUM.txt"),
    output:
        tsv=temp("temp/taxonomy/gtdb.tsv.gz"),
    log:
        "logs/taxonomy/gtdb.log",
    params:
        url="https://data.gtdb.ecogenomic.org/releases/release220/220.0/ar53_taxonomy_r220.tsv.gz",
        hash_function="md5",
        extra="--file-allocation none --retry-wait 5 --console-log-level warn --log-level notice",
        checksum=lambda w, input: open(input[0]).read().strip()
    threads: 2
    wrapper:
        wrapper_ver + "/utils/aria2c"
```

Snakemake will evaluate the param when the input is already present.

johanneskoester: One could also simplify the pattern by adding another semantic helper function to the Snakemake DSL, e.g. 

```python
def parse_input(input_item, parser=None):
    def inner(wildcards, input):
        with open(input.get(input_item), "r") as infile:
            if parser is None:
                return infile.read().strip()
            else:
                return parser(infile)
    return inner
```

fgvieira: That could be away, but only if the checksum file only contains one hash (the one for the file being downloaded). However, most times they contain checksums for more than one file (see example). So, the wrapper needs to match the name of the file being downloaded with those present in the checksums file and retrieve the appropriate hash.

johanneskoester: Then, above rule would become

```
rule gtdb_taxonomy:
    input:
        checksum=storage.http("https://data.gtdb.ecogenomic.org/releases/release220/220.0/MD5SUM.txt"),
    output:
        tsv=temp("temp/taxonomy/gtdb.tsv.gz"),
    log:
        "logs/taxonomy/gtdb.log",
    params:
        url="https://data.gtdb.ecogenomic.org/releases/release220/220.0/ar53_taxonomy_r220.tsv.gz",
        hash_function="md5",
        extra="--file-allocation none --retry-wait 5 --console-log-level warn --log-level notice",
        checksum=parse_input("checksum")
    threads: 2
    wrapper:
        f"{wrapper_ver}/utils/aria2c"
```

johanneskoester: If special parsing is needed one would pass this as the parser argument, like this (excuse the switch to polars, I am trying to migrate my thinking to it currently):

```python
import polars as pl
import polars.selectors as cs

def extract_checksum(infile):
    return pl.read_csv(
        infile,
        separator=" ",
        has_header=False
    ).filter(
        cs.by_index(1).str.ends_with("ar53_taxonomy_r220.tsv.gz"
    ).select(cs.by_index(0)).item()

rule gtdb_taxonomy:
    input:
        checksum=storage.http("https://data.gtdb.ecogenomic.org/releases/release220/220.0/MD5SUM.txt"),
    output:
        tsv=temp("temp/taxonomy/gtdb.tsv.gz"),
    log:
        "logs/taxonomy/gtdb.log",
    params:
        url="https://data.gtdb.ecogenomic.org/releases/release220/220.0/ar53_taxonomy_r220.tsv.gz",
        hash_function="md5",
        extra="--file-allocation none --retry-wait 5 --console-log-level warn --log-level notice",
        checksum=parse_input("checksum", parser=extract_checksum)
    threads: 2
    wrapper:
        f"{wrapper_ver}/utils/aria2c"
```

johanneskoester: In addition, we could offer a helper function that parses such checksum files, like

```python
def parse_checksum_input(input_item):
    def extract_checksum(infile):
        return dict(pl.read_csv(
            infile,
            separator=" ",
            has_header=False
        ).iter_rows())

    return parse_input(input_item, parser=extract_checksum)
```

Then, the wrapper could implement logic to access the returned dict and get the required checksum by comparing with the url.

fgvieira: I like the idea of having a helper function as it makes the code much more readable. Not sure which one might be better, but the second one would be hard to use if not using wrappers, no?
If so, maybe go for the first one; we'd just need to also pass the name of the file to search for (e.g. `ar53_taxonomy_r220.tsv.gz`).

johanneskoester: yes, and one could simply allow that all additional kwargs passed to parse_input are passed to the given parser function.

fgvieira:
Done! 
I've included the function`parse_input` for testing, but the idea is to move it to `snakemake` as a helper function, no? If so, will you do it or should I open a PR?

On another issue, one test is currently disabled since `snakemake-storage-plugin-http` is not available in the test environment. Should we leave it as-is (to keep the two repos tests separate), or will you enable it?

johanneskoester:
> I've included the function`parse_input` for testing (used `panda` since `polars` creates a conflict with `aria2c` dependencies), but the idea is to move it to `snakemake` as a helper function, no? If so, will you do it or should I open a PR?
> 
> On another issue, one test is currently disabled since `snakemake-storage-plugin-http` is not available in the test environment. Should we leave it as-is (to keep the two repos tests separate), or will you enable it?

Feel free to add snakemake-storage-plugin-http to the test env.
Would be great if you could make a PR that adds both helper functions as a PR to ioutils.py in Snakemake.

johanneskoester (approved):

PR #2729, opened 2024-03-08
feat: Rseqc read gc

tdayris:
This PR adds [RSeQC read_gc](https://rseqc.sourceforge.net/#read-gc-py) to the list of available wrappers.

I've been trying to make a single big multi-command wrapper for all RSeQC sub-programs, but 21 of them produces unique output files with fixed extension name, come as a single file, some as pairs, trio, or more. (_e.g._ bam2fq producing one or two fastq, bam2wig producing pairs of wiggles and requires an additional dependency, read_gc producing up to 4 files ending with `*.GC.*`)

It seems to me that multiple wrappers will be easier to develop, read and maintain.

But in the end, wrappers for sub-programs like `bam_stat.py` and `sc_bam_stat.py` or `infer_experiment.py` will be very similar and very simple.

### QC

* [X] I confirm that:


bio/rseqc/read_gc/wrapper.py
```diff
@@ -0,0 +1,44 @@
...
+extra = snakemake.params.get("extra", "")
+log = snakemake.log_fmt_shell(stdout=True, stderr=True, append=True)
+
+
+with TemporaryDirectory() as tempdir:
+    shell(
+        "read_GC.py {extra} "
+        "--input-file {snakemake.input} "
+        "--out-prefix {tempdir}/out "
+        "{log} "
```
fgvieira, line 22: Not quite sure but, what would happen if this rule is re-run? Would log be appended to the previous run log?

tdayris: I just did the test:
* changed the wrapper prefix from `master` to `file://path/to/wrapper`
* ran the test file 3 times, using snakemake v8.5.3

There is no repetition of the log content:

```$ cat rseqc.log
 Read BAM file ...  Done
writing GC content ...
writing R script ...
null device 
          1 
renamed '/tmp/tmpupksz9fd/out.GC.xls' -> 'A.gc_percent.xls'
renamed '/tmp/tmpupksz9fd/out.GC_plot.r' -> 'A.script.r'
renamed '/tmp/tmpupksz9fd/out.GC_plot.pdf' -> 'A.plot.pdf'
```

fgvieira (approved):

PR #2732, opened 2024-03-08
feat: All ngsderive subcommands

tdayris:
This PR adds all [ngsderive](https://github.com/stjudecloud/ngsderive) subcommands in a single wrapper, just like XSV and SeqKit.

### QC

* [X] I confirm that:


bio/ngsderive/meta.yaml
```diff
@@ -0,0 +1,16 @@
+name: ngsderive
+description: Backwards computing information from next-generation sequencing data and annotating splice junctions
+url: "https://github.com/stjudecloud/ngsderive"
+authors:
+  - "Thibault Dayris"
+input:
+  - ngs: "Path to BAM/SAM/Fastq file. SAM/BAM files should be indexed."
+  - gene_model: "Path to sorted GTF/GFF file. Should be tabix indexed."
+output:
+  - out: Path to output file
```
fgvieira, line 10:
```suggestion
  - tsv: Path to output file
```

bio/ngsderive/test/Snakefile
```diff
@@ -0,0 +1,85 @@
...
+        "master/bio/ngsderive"
+
+
+rule test_ngsderive_junction_annotation:
+    input:
+        ngs="A.rg.bam",
+        gene_model="annotation.sorted.gtf.gz",
+    output:
+        tsv="A.junctions.tsv",
+        junction_dir=directory("junctions"),
```
fgvieira, line 21: What are the output files generated into this folder? Would it be possible to name them explicitly as output files?

tdayris: It depends on the content of the annotation and the number of input files. Here, in the tests, there was only one sample and one annotation, so the `--junction-files-dir` contains only one file.

We could also expect a list and/or deduce a common prefix. The user would have to guess the number of output files produced.

Let me make the changes to accept both `directory` or a list of output files.

bio/ngsderive/meta.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - "Thibault Dayris"
+input:
+  - ngs: "Path to BAM/SAM/Fastq file. SAM/BAM files should be indexed."
+  - gene_model: "Path to sorted GTF/GFF file. Should be tabix indexed."
+output:
+  - tsv: Path to output file
+  - junctions: Optional path to junction directory, or list of paths to junction files with a common prefix
+params:
+  - subcommand: Name of the `ngsderive` subcommand
+  - extra: Optional paramters provided to the subcommand, besides `-o`, `-g`
```
fgvieira, line 14:
```suggestion
  - extra: Optional parameters, besides `-o`, `-g`
```

bio/ngsderive/meta.yaml
```diff
@@ -0,0 +1,16 @@
...
+authors:
+  - "Thibault Dayris"
+input:
+  - ngs: "Path to BAM/SAM/Fastq file. SAM/BAM files should be indexed."
+  - gene_model: "Path to sorted GTF/GFF file. Should be tabix indexed."
+output:
+  - tsv: Path to output file
+  - junctions: Optional path to junction directory, or list of paths to junction files with a common prefix
+params:
+  - subcommand: Name of the `ngsderive` subcommand
```
fgvieira, line 13:
```suggestion
  - command: Name of the `ngsderive` subcommand
```

bio/ngsderive/test/Snakefile
```diff
@@ -0,0 +1,101 @@
...
+        "master/bio/ngsderive"
+
+
+rule test_ngsderive_junction_annotation_list:
+    input:
+        ngs="A.rg.bam",
+        gene_model="annotation.sorted.gtf.gz",
+    output:
+        tsv="A.junctions_list.tsv",
+        junction_files_dir=["junctions/A.rg.bam.junctions.tsv"],
```
fgvieira, line 37: Maybe just another name, so it is also easier to parse in the wrapper:
```suggestion
        junction_files=["junctions/A.rg.bam.junctions.tsv"],
```

bio/ngsderive/test/Snakefile
fgvieira: Cool, just one last thing, do you think it would make sense to rename `junction_dir`? Since now it can accept both, it might be misleading... maybe just `junction`.

johanneskoester (approved):

PR #2734, opened 2024-03-11
feat: Goleft indexcov

tdayris:
THis PR add [goleft indexcov](https://github.com/brentp/goleft/tree/master/indexcov) to the list of available wrappers.

### QC

* [X] I confirm that:


bio/goleft/indexcov/wrapper.py
```diff
@@ -0,0 +1,55 @@
...
+            "{tempdir}/out/out-indexcov.bed.gz "
+            "{snakemake.output.bed} "
+            "{log} "
+        )
+
+    # Number of files withing tempdir output
+    # is defined by user command line parameters
+    # and/or chromosomes present in the fasta index
+    if "html" in snakemake.output.keys():
+        # Not create autimatically by Snakemake
```
fgvieira, line 52:
```suggestion
        # Not create automatically by Snakemake
```

bio/goleft/indexcov/wrapper.py
```diff
@@ -0,0 +1,55 @@
...
+    shell(
+        "goleft indexcov "
+        "{extra} "
+        "--directory {tempdir}/out "
+        "--fai {snakemake.input.fai} "
+        "{snakemake.input.aln} "
+        "{log} "
+    )
+
+    if "ped" in snakemake.output.keys():
```
fgvieira, line 24: Some wrappers use the form:
```suggestion
    if snakemake.output.get("ped"):
```

not sure if it makes a difference.

tdayris: I test existence of the key in an object. The suggestion tries to get a key from an object and, it it fails, returns `None` object, then evaluates `None` as `False` or not-`None` as `True`.

In terms of time it's almost alike:

``` python
In [1]: def foo(d):
   ...:     if d.get("key"):
   ...:         return "Case1"
   ...:     return "Case2"
   ...: 

In [2]: def bar(d):
   ...:     if "key" in d:
   ...:         return "Case1"
   ...:     return "Case2"
   ...: 

In [3]: d = {"key": "exists"}

In [4]: %timeit foo(d)
92.6 ns ± 1.4 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit bar(d)
76.2 ns ± 0.12 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: e = {"not_key": "does not exists"}

In [7]: %timeit foo(e)
85.7 ns ± 0.128 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [8]: %timeit bar(e)
73.1 ns ± 0.148 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

Time difference won't make any difference even if my solution is faster. If you see a difference I did not mention, please tell me.

Do you want me to make the changes ?

bio/goleft/indexcov/wrapper.py
```diff
@@ -0,0 +1,55 @@
...
+            "{tempdir}/out/out-indexcov.bed.gz "
+            "{snakemake.output.bed} "
+            "{log} "
+        )
+
+    # Number of files withing tempdir output
+    # is defined by user command line parameters
+    # and/or chromosomes present in the fasta index
+    if snakemake.output.get("html"):
+        # Not create automatically by Snakemake
```
fgvieira, line 52: I was under the impression `snakemake` always created the output folders.

tdayris: I thought the same, but if I do not run `mkdir`, then I get a shell error with `html` not being a directory... I can update the branch to raise that error if you'd like.

tdayris: See error on [a2d704e](https://github.com/snakemake/snakemake-wrappers/pull/2734/commits/a2d704e70783dbee0ff956af1d3149f0ef9e472c)

``` bash
2024-03-12T14:20:59.5983080Z indexcov finished: see /tmp/tmpvwnjokdn/out/index.html for overview of output
2024-03-12T14:20:59.5983737Z renamed '/tmp/tmpvwnjokdn/out/out-indexcov.ped' -> 'indexcov.ped'
2024-03-12T14:20:59.5984331Z renamed '/tmp/tmpvwnjokdn/out/out-indexcov.roc' -> 'indexcov.roc'
2024-03-12T14:20:59.5984957Z renamed '/tmp/tmpvwnjokdn/out/out-indexcov.bed.gz' -> 'regions.bed.gz'
2024-03-12T14:20:59.5985469Z mv: target 'report' is not a directory
```

So I'll include again the `mkdir` in the wrapper.

fgvieira: @johanneskoester can you confirm this behavior? Is this intended?

johanneskoester: Yes. Snakemake creates all parent dirs up to the path given. But the path itself, i.e. the leaf of the path, the last item, is not created because Snakemake cannot know whether the process wants to do it by itself.

johanneskoester:
@fgvieira feel free to merge if you are happy with the PR.

bio/goleft/indexcov/wrapper.py
```diff
@@ -0,0 +1,55 @@
...
+            "{log} "
+        )
+
+    # Number of files withing tempdir output
+    # is defined by user command line parameters
+    # and/or chromosomes present in the fasta index
+    if snakemake.output.get("html"):
+        # Not created automatically by Snakemake
+        shell("mkdir --parents --verbose {snakemake.output.html} {log}")
+
```
fgvieira, line 54:
```suggestion
        from pathlib import Path
        Path(snakemake.output.html).mkdir(parents=True, exist_ok=True)
```

fgvieira (approved):

PR #2739, opened 2024-03-12
feat: pyTMB

tdayris:
This PR adds [`pyTMB`](https://github.com/bioinfo-pf-curie/TMB) to the list of available wrappers.

### QC

* [X] I confirm that:


bio/tmb/pytmb/meta.yaml
```diff
@@ -0,0 +1,17 @@
+name: "pyTMB.py"
+description: Calculate a Tumor Mutational Burden (TMB) score from a VCF file
+url: "https://github.com/bioinfo-pf-curie/TMB?tab=readme-ov-file#tumor-mutational-burden"
+authors:
+  - "Thibault Dayris"
+input:
```
fgvieira, line 6: Isn't the `bed` file missing?

bio/tmb/pytmb/meta.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - vcf: Path to input variants (`vcf`, `vcf.gz`, or `bcf` formatted)
+  - db_config: Path to database config file (`yaml` formatted)
+  - var_config: Path to variant config file (`yaml` formatted)
+output:
+  - res: Path to TMB results
+  - vcf: Optional path to variants considered for TMB calculation
+params:
+  - extra: Optional parameters provided to `pyTMB.py`, besides `-i`, `--dbConfig`, `--varConfig`, `--bed`, or `--export`
+notes: |
+  `output.vcf` creates a temporary file, which prefix is equal to the prefix of the input vcf file. Use `shadow` option
```
fgvieira, line 16: would it be possible to deal with this?

bio/tmb/pytmb/wrapper.py
```diff
@@ -0,0 +1,47 @@
...
+shell(
+    "pyTMB.py {extra} "
+    "{db_config} {var_config} {bed} "
+    "--vcf {snakemake.input.vcf} "
+    "> {snakemake.output.res} "
+    "{log} "
+)
+
+# The export file is created in the execution directory
+# with no option to control its prefix (other than the
```
fgvieira, line 42: maybe run it in a temp dir?

tdayris: I thought that was the point of `shadow: 'shallow'` directive in Snakefile.

Anyway, I made the changes.

johanneskoester: The point of wrappers is to hide such details. shadow is more to avoid such issues in an ad-hoc setting.

bio/tmb/pytmb/meta.yaml
tdayris: It is done through a temporary directory as you suggested

bio/tmb/pytmb/wrapper.py
tdayris: Changes have been made accordingly.

johanneskoester (approved):

Issue #2756, opened 2024-03-19
the output files generated by using STAR in snakemake-wrappers should be more comprehensive.


zhongzheng1999:
I'm currently working on setting up a simple RNA-seq analysis pipeline using snakemake, and I've integrated the STAR wrapper from snakemake-wrappers. However, I've run into an issue where specifying `--quantMode TranscriptomeSAM `prevents me from extracting the `Aligned.toTranscriptome.out.bam` file as an output. I believe this file is essential for subsequent transcript quantification.

After examining the corresponding `wrapper.py` file, I noticed that within the `with tempfile.TemporaryDirectory() as tmpdir: `block, you use the `cat` command to output specific temporary files as result files. However, I observed that you only provided options such as `ReadsPerGene.out.tab`, `Chimeric.out.junction`, `SJ.out.tab`, `etc.`, but there is no code for outputting `Aligned.toTranscriptome.out.bam.`

zhongzheng1999:
When I use the --outStd BAM_SortedByCoordinate parameter, the output for {snakemake.output.aln} is the Aligned.out.bam file. However, I'm wondering if using --quantMode TranscriptomeSAM would mean that Aligned.toTranscriptome.out.bam is output instead of Aligned.out.bam. I've checked the STAR documentation but couldn't find an explanation for this issue.

zhongzheng1999:
@kashyapchhatbar @fgvieira

fgvieira:
From STAR's manual:

> With `--quantMode TranscriptomeSAM` option STAR will output alignments translated into transcript coordinates in the `Aligned.toTranscriptome.out.bam` file (in addition to alignments in genomic coordinates in `Aligned.*.sam/bam` files).

So, I'd say that a new file is created. Can you double-check it?

fgvieira:
If you modified `wrapper.py`, it would be great if you could submit a PR.

PR #2772, opened 2024-03-27
perf: Added 'TranscriptomeSAM' parameter

zhongzheng1999:
#2756 Added 'TranscriptomeSAM' parameter to point to transcriptome-based alignment results!


zhongzheng1999:
This may require further confirmation as to whether it makes sense to specify the aligned file as `Aligned.toTranscriptome.out.bam` directly, which might also be affected by the --outStd parameter, and thus output the `SAM` file?

PR #2929, opened 2024-05-07
feat: add wrapper to samtools collate

fgvieira:
### QC

* [x] I confirm that:


cmeesters:
@fgvieira happy to review, and I already triggered a merge with main - yet I have a paper to submit today and will find time tomorrow

bio/samtools/collate/environment.yaml
```diff
@@ -0,0 +1,7 @@
+channels:
+  - conda-forge
+  - bioconda
+  - nodefaults
+dependencies:
+  - samtools =1.20
+  - snakemake-wrapper-utils =0.6.2
```
cmeesters, line 7: should we not read >= for such cases?

cmeesters (approved):
for the test 2 threads and otherwise 8 seems fine - it is my experience, too, that samtools scales to about 8 threads max. regardless of the machine (sometimes only up to 2-4)

bio/samtools/collate/environment.yaml
fgvieira: It was my impression that versions are "fixed" and are updated through a GitHub action (so every version bump is tested).

cmeesters: ok, cool - did not know that.

Please excuse me, I need to take a week off, definitively going offline, now. So, most reviews which are pending are at @dlaehnemann, right?

fgvieira: No problem and thanks for the reviews.

PR #2940, opened 2024-05-21
fix: remove NA string replacement, fixed upstream

lparsons:
Wrapper previously failed if the optional output `qc_metrics` was not supplied due to NA replacement code that was not in a conditional. The code is no longer required since NA strings are now output correctly in https://github.com/deeptools/deepTools/pull/1002

### QC

* [x] I confirm that:

### DIFF
--- a/bio/deeptools/plotfingerprint/test/Snakefile
+++ b/bio/deeptools/plotfingerprint/test/Snakefile
@@ -15,7 +15,7 @@ rule plot_fingerprint:
         "logs/deeptools/plot_fingerprint.log",
     params:
         # optional parameters
-        "--numberOfSamples 200 ",
+        extra="--numberOfSamples 200",
     threads: 8
     wrapper:
         "master/bio/deeptools/plotfingerprint"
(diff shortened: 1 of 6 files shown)


bio/deeptools/plotfingerprint/test/Snakefile
```diff
@@ -2,21 +2,20 @@ rule plot_fingerprint:
...
+        qc_metrics="plot_fingerprint/qc_metrics.txt",
     log:
-        "logs/deeptools/plot_fingerprint.log"
+        "logs/deeptools/plot_fingerprint.log",
     params:
         # optional parameters
-        "--numberOfSamples 200 "
-    threads:
-        8
+        "--numberOfSamples 200 ",
```
fgvieira, line 18: For the sake of consistency, can you use `params.extra` instead?

```suggestion
        extra="--numberOfSamples 200",
```

fgvieira, line 18:
```suggestion
        extra="--numberOfSamples 200",
```

fgvieira:
Looks good!
Wanted to make a suggestion, and ended up committing. Can you use `params.extra` instead?

lparsons:
> Looks good! Sorry, but wanted to make a suggestion and ended up committing. Can you use `params.extra` instead?

Sure, I can do that. I took a look at the other deeptools wrappers and tried to make them more consistent, so I updated plotheatmap and plotprofile as well.

fgvieira (approved):

PR #3072, opened 2024-07-17
feat: add new wrapper to create annotation tables via Ensembl biomart

dlaehnemann:
### QC

* [x] I confirm that:


bio/reference/ensembl-transcripts-to-genes-mapping/test/Snakefile
```diff
@@ -0,0 +1,14 @@
+rule create_transcripts_to_genes_mapping:
+    output:
+        mapping="resources/ensembl_transcripts_to_genes_mapping.tsv.gz",  # .gz extension is optional, but recommended
```
johanneskoester, line 3: Maybe offer writing to parquet as the default? We should in general try to move away from text formats as much as possible since the dust around arrow has settled now.

bio/reference/ensembl-transcripts-to-genes-mapping/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+    bioconductor-biomart, optionally adding in additional columns (for example
+    HGNC gene symbols and genecards IDs for the human reference). For
+    reference, have a look at the Ensembl biomart online
+    (``https://www.ensembl.org/biomart/martview``) or at the documentation for
+    bioconductor-biomart
+    (``https://bioconductor.org/packages/devel/bioc/vignettes/biomaRt/inst/doc/accessing_ensembl.html``).
+authors:
+  - David Lähnemann
+output:
+  - tab-separated values (TSV) file, with columns ensembl_transcript_id and ensembl_gene_id, plus optional columns (can be gzipped)
```
johanneskoester, line 13: Is it necessary to constrain this to these defaults? Why not offering a wrapper to obtain an arbitrary combination of attributes from biomart?

dlaehnemann: It's a specific use-case that comes up repeatedly. And I think having some minimal defaults makes sense, although these could also simply be the default list specified in the example rule.

So sure, we could generalize this further. One further question then is, where to stop generalizing: Should we also allow for other `biomart`s than the `genes` one?

bio/reference/ensembl-transcripts-to-genes-mapping/test/Snakefile
dlaehnemann: OK, the best option (because most lightweight dependency in R) for this should be nanoparquet:
https://r-lib.github.io/nanoparquet/

However, this is not yet packaged on conda-forge. Could you head over and approve the `conda-forge/r` team being listed in the new package that I am creating for this? The linter kindly asked for this. I think you simply need to comment here:
https://github.com/conda-forge/staged-recipes/pull/27021

johanneskoester: Done.

bio/reference/ensembl-transcripts-to-genes-mapping/meta.yaml
dlaehnemann: OK, the generalization is in this commit:
https://github.com/snakemake/snakemake-wrappers/pull/3072/commits/1dd6145c85159e7bcf3a890903659fc9f446baa9

I'll debug as necessary. And I'll add in more examples/tests once `r-nanoparquet` is usable via `conda-forge` and we can also write to parquet.

bio/reference/ensembl-transcripts-to-genes-mapping/test/Snakefile
dlaehnemann: With [`r-nanoparquet` now in `conda-forge`](https://github.com/conda-forge/staged-recipes/pull/27021), I have implemented parquet support and added a respective example and test case.

bio/reference/ensembl-biomart-table/wrapper.R
```diff
@@ -0,0 +1,127 @@
...
+library("biomaRt")
+
+wanted_biomart <- snakemake@params[["biomart"]]
+# bioconductor-biomart needs the species as something like `hsapiens` instead
+# of `homo_sapiens`
+wanted_species <- str_replace(
+  snakemake@params[["species"]],
+  "^(\\w)\\w+_(\\w+)$",
+  "\\1\\2"
+)
```
tedil, line 25: is this valid across all species?

bio/reference/ensembl-biomart-table/wrapper.R
```diff
@@ -0,0 +1,127 @@
...
+  grch <- "37"
+  version <- NULL
+  cli_warn(c(
+    "As you specified build 'GRCH37' in your configuration yaml, biomart forces",
+    "us to ignore the release you specified ('{release}')."
+  ))
+} else {
+  grch <- NULL
+  version <- wanted_release
+}
```
tedil, line 45: I believe it may be possible to get a versioned grch37 via the branch option, cf [wrappers/ensembl-sequence](https://github.com/snakemake/snakemake-wrappers/blob/9d47dd76995e4baeb2b9ea7b3278cc713e9e4f17/bio/reference/ensembl-sequence/wrapper.py#L16)

bio/reference/ensembl-biomart-table/wrapper.R
```diff
@@ -0,0 +1,127 @@
...
+  }
+  mart
+}
+
+gene_ensembl <- get_mart(wanted_biomart, wanted_species, wanted_build, version, grch, "gene_ensembl")
+
+if ( !is.null(wanted_chromosome) ) {
+  table <- getBM(
+    attributes = wanted_columns,
+    filters = 'chromosome_name',
```
tedil, line 79: Since this wrapper has the generic name "ensembl-biomart-table", perhaps it should be possible to be able to specify filters in the config?

tedil (commented):
Just some general remarks for now ;)

bio/reference/ensembl-biomart-table/wrapper.R
dlaehnemann: Of course it is not. :disappointed:

There seem to be 2 component and 3 component species names. Only the last component of the species name is fully kept in the biomart dataset names, the other components are abbreviated to their first letter.

So there's a more robust parsing now, that throws an error if it encounters anything other than 2 or 3 components. It's in:
https://github.com/snakemake/snakemake-wrappers/pull/3072/commits/d90c492dc01753e30dbaec26282d00b3fb85e159

dlaehnemann: Ah, and for possible species names, check:
https://www.ensembl.org/info/about/species.html

But this seems to contain more species (324) than biomart has available for the `genes` biomart (214). Nothing ever sums up properly... :shrug:

And for documentation on getting the species in the `genes` biomart, here's a snippet:
```
library("tidyverse")
library("biomaRt")
genes <- useEnsembl(biomart = "genes")
genes_datasets <- listDatasets(genes) |> as_tibble()
species <- genes_datasets |>
  select(dataset) |>
  separate_wider_delim(dataset, delim = "_", names = c("species", "dataset", "ensembl")) |>
  select(species) |>
  distinct()
```

bio/reference/ensembl-biomart-table/wrapper.R
dlaehnemann: That's for FTP download links. But in the function `useEnsembl()` used here, you get an error if you specify both `grch` and `version` at the same time:
https://github.com/grimbough/biomaRt/blob/ddf78e12b8614205251e37cb2c7b24d9f3908b85/R/ensembl.R#L206-L207

:shrug:

bio/reference/ensembl-biomart-table/wrapper.R
dlaehnemann: Interestingly, non-existing filter values (as opposed to filter names) are simply ignored. This is now noted in the docs, as well. And I also added an example of the syntax for multiple filter values in one filter:
https://github.com/snakemake/snakemake-wrappers/pull/3072/commits/7175f6dca659e1dd2c975ee0d99b69eb6e219981

dlaehnemann:
Thanks. Those comments improved the wrapper quite a bit further. Care for a "final review"?

tedil (approved):
Very nice and well documented! LGTM

PR #3165, opened 2024-09-03
feat: create nanosim wrapper for all three modes of simulator.py

dlaehnemann:
respective tests already passed locally

### QC

## Summary by CodeRabbit

- **New Features**
	- Introduced new configuration files for setting up a reproducible environment for the NanoSim simulator on Linux 64-bit systems.
	- Added a `Snakefile` for structured rules to simulate genomic, transcriptomic, and metagenomic data.
	- Created a `wrapper.py` to enhance usability within a Snakemake pipeline.
	- Added a `meta.yaml` file providing essential metadata and configuration parameters for the NanoSim simulator.

- **Bug Fixes**
	- Addressed compatibility issues with pretrained models by pinning specific package versions.

- **Tests**
	- Expanded the test suite with new functions for testing NanoSim across different biological contexts.


bio/nanosim/simulator/environment.yaml
```diff
@@ -0,0 +1,12 @@
+channels:
+  - conda-forge
+  - bioconda
+  - nodefaults
+dependencies:
+  # bioconda recipe needs patching to work with pretrained models, but the
+  # patch only applies to versions <= 3.1:
+  # https://github.com/bioconda/bioconda-recipes/pull/40449/files
+  # TODO: update to latest version, if this gets resolved
+  - nanosim =3.1
```
fgvieira, line 10: There is version `3.2.0` and that bioconda PR has been merged.
Shouldn't we bump the version?

fgvieira:
Instead of the extra snakefile to download the envs for testing, would it make sense for the wrapper to also allow a gzipped model?
This could be internally decompressed to a temp folder and directly specified as input (even if remote with `model=storage.http(https://github.com/bcgsc/NanoSim/raw/master/pre-trained_models/human_NA12878_DNA_FAB49712_guppy.tar.gz)`).

bio/nanosim/simulator/environment.yaml
dlaehnemann: Hmm, maybe my explanatory comment needs tweaking. The patch was merged, when version 3.1 was the current version, and it only applies to versions `<= 3.1`, as the linked code shows. But version 3.2 would also need the patch, and it simply isn't applied for that version. So for the purposes of this wrapper, this resulted in a broken package environment for me. Do you have a suggestion for a wording that would make this clearer for you?

dlaehnemann:
I think I would prefer actually listing the required files with `multiext()` like this, to explicitly specify which files are needed for which type of simulation (they slightly differ...). In addition, one might want to add a wrapper for training the models at some point, and that should also track the individual files it produces, and one should make sure that one maps to the other.

And yes, I agree that the download via the extra snakefile is a bit of a hacky workaround, but I think for testing purposes this is good enough / OK.

bio/nanosim/simulator/environment.yaml
```diff
@@ -0,0 +1,12 @@
+channels:
+  - conda-forge
+  - bioconda
+  - nodefaults
+dependencies:
+  # bioconda recipe needs patching to work with pretrained models, but the
+  # patch only applies to versions <= 3.1:
```
fgvieira, line 7:
```suggestion
  # patch has only been applied to versions <= 3.1:
```

bio/nanosim/simulator/environment.yaml
fgvieira: Do you know when it will be patched? Just because we will always get an autobump PR every week.

bio/nanosim/simulator/environment.yaml
dlaehnemann: Much better, thanks!

bio/nanosim/simulator/environment.yaml
dlaehnemann: Ah, yes, didn't think about the autobump PRs.

One quick upstream fix could be to extend the patch to `<=3.2`.

But we also don't know when this fix will become obsolete. The real fix would be for `nanosim` to have clear and somewhat robust dependencies, but they seem to be all over the place, with even trained models depending on particular versions of `scikit-learn` and `numpy`. So I was hoping I don't have to wade into fixing this "more properly"...

But we could merge this wrapper as is, because this does work and passes the tests. And we could then try the quick upstream fix, and see whether this then works with the first autobump that comes up...

bio/nanosim/simulator/wrapper.py
```diff
@@ -0,0 +1,163 @@
...
+        " {extra} "
+        " {fq} "
+        " {perfect} "
+        " --num_threads {snakemake.threads} "
+        " --output {prefix}; "
+        " mv {prefix}_{sample_infix}aligned_reads{extension} {snakemake.output.reads}; "
+        " mv {prefix}_{sample_infix}aligned_error_profile {snakemake.output.errors}; "
+        " {mv_unaligned_reads}"
+        ") {log}; "
+    )
```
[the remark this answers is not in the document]
dlaehnemann: What unit tests would you suggest?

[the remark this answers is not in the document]
dlaehnemann: Almost all of these cases or suggestions are covered by the tests provided via the `test/Snakefile`. And unit testing basic Python functionality like the proper creation and cleanup of a temporary directory would be re-inventing the wheel. This functionality should be unit tested in Python itself.

Also, there are some inaccuracies in the previous summary. The script or the example and test invocations in `test/Snakefile` do not specify a subcommand. The subcommand is inferred from the pattern of named input files.

dlaehnemann:
@fgvieira any more time for a review, or should I ask around?

fgvieira:
@dlaehnemann sorry for the delay, but I was looking a bit into it and, from bcgsc/NanoSim#162, it seems that the `scikit-learn <=0.22.1` dependency is not needed to run NanoSim, only to use the available (and outdated) pre-trained models. But what if users want to create their own models? It would be a bit strange to constraint on using an old version of `scilit-learn`, no?

dlaehnemann:
Thanks for looking into this further. A thorough review is always appreciated. And any ideas for making this more useful are welcome.

We do need those pre-trained models, if we want to do the CI tests here. And I have only seen the dorado model for the `genome` mode as a more recently trained model. I have also dug into the manuscripts and their data availability indications, and didn't find anything more recent. They say they will provide some newer models in the future, but that note is from November last year. Currently, all the existing models for metagenome and transcriptome mode are older and will require the version constraint.
One alternative would be to include a full trained model from a more recent version in our own testing data, but I think it would be too big (60-80 MB) to drag along in the repo.
The other alternative would be to implement a nanosim model training step wrapper, and try to test the two wrappers in tandem. But this is definitely another couple of days' work, and shifts the problem to then having to download a full-blown Nanopore dataset and do the training on it, all in the GitHub Actions CI. This might work, but experience shows it's never easy... :sweat_smile:

Or do you have any good ideas how to got about this?

fgvieira:
Well, I guess no perfect solution here. :smile: 
I agree that including a full trained model is a bit too much to have in the repo.

I can see three options:

1. the ideal scenario would be for the authors to update the pretrained models, but we cannot wait for it.
2. add a wrapper to train a model on a mock/small dataset (and test both wrappers at once)
3. leave it as is, and update things when 1. is done.

You know the program better, so I'll leave it up to you.

fgvieira (approved):

dlaehnemann:
I cross-referenced this once more over in the NanoSim repo, and will merge it as is, for now. They have recently provided a newly trained model for genomic DNA, but not yet for transcriptomic or metagenomic data. So I would merge this as is for now, and we can then hopefully remove the TODO in the `environment.yaml` soon (so your point `3.`).

PR #3250, opened 2024-09-17
feat: Enhance phys/root/filter functionality

GoodingJamie:
This PR expands on the functionality of the wrapper written by @laf070810 for `ROOT.RDataFrame.Filter()`. The core of the wrapper is still the same; however, this PR adds the option to provide the selection criteria as a list of criterion strings or a dictionary wherein the keys define names for each criterion. A `verbose` mode is added for users to receive a report of their selections, as per [`ROOT.RDataFrame.Report()`](https://root.cern/doc/master/classROOT_1_1RDF_1_1RInterface.html#acac57f46fa563b5033d69142d6874424).

### QC

## Summary by CodeRabbit

- **New Features**
	- Introduced three new filtering rules: `filter_str`, `filter_list`, and `filter_dict`, enhancing data processing flexibility.
	- Added support for multi-threading in data filtering based on specified thread count.
	- Included a verbose reporting feature for better insight into data processing operations.

- **Bug Fixes**
	- Improved criteria parsing to accept string, list, or dictionary formats for filtering.

- **Tests**
	- Expanded the testing suite with new test functions for each filtering method, ensuring robust functionality.

### DIFF
--- a/phys/root/filter/wrapper.py
+++ b/phys/root/filter/wrapper.py
@@ -17,7 +17,7 @@ elif isinstance(_smk_criteria, List):
     criteria = _smk_criteria
     labels = _smk_criteria
 elif isinstance(_smk_criteria, Dict):
-    criteria, labels = _smk_criteria.items()
+    labels, criteria = _smk_criteria.items()
 else:
     raise TypeError("Parameter 'criteria' should be type of 'str', 'list' or 'dict'")
 
@@ -26,6 +26,8 @@ branches_to_save = snakemake.params.get("branches_to_save", None)
 df = ROOT.RDataFrame(snakemake.params.input_tree_name, snakemake.input[0])
 for criterion, label in zip(criteria, labels):
     df = df.Filter(criterion, label)
+
+
 if branches_to_save is not None:
     df.Snapshot(
         snakemake.params.output_tree_name,
@@ -34,6 +36,7 @@ if branches_to_save is not None:
     )
 else:
     df.Snapshot(snakemake.params.output_tree_name, snakemake.output[0])
+
 
 if snakemake.params.get("verbose", False):
     report = df.Report()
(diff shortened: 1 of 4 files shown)


phys/root/filter/wrapper.py
```diff
@@ -1,18 +1,39 @@
...
 import ROOT
 
-ROOT.EnableImplicitMT(snakemake.threads)
+if snakemake.threads > 1:
+    ROOT.EnableImplicitMT(snakemake.threads)
+else:
+    ROOT.DisableImplicitMT()
+
+# Parse criteria
+_smk_criteria = snakemake.params.get("criteria", "true")
```
fgvieira, line 15: Won't `"true"` be interpreted as a string filtering criteria with `'criterion = "true"; label = "true"`? Is this intended?

fgvieira:
@coderabbitai review

phys/root/filter/wrapper.py
GoodingJamie: Ah, yes it will be! This is the behaviour in master currently—the only use case I can think of would be to copy a tree in one file to a differently named tree in another file, but there's `rootcp` for this (may be the subject of a PR in the near future!). So I think here it would be safe to change this and require that `criteria` is passed:

```suggestion
_smk_criteria = snakemake.params.get("criteria", "true")
```
```suggestion
_smk_criteria = snakemake.params.criteria
```

where a missing `params.criteria` should raise a `TypeError` when being converted to `criteria` and `labels`.

fgvieira: But isn't `criteria` optional? At least that is what it says on the `meta.yaml` file.

GoodingJamie: `criteria` is optional currently, but there isn't a use case I can think of for not passing `criteria` which isn't already covered by `rootcp` (for which I've opened #3251). I'll update this in `meta.yaml` to reflect this

fgvieira (approved):

GoodingJamie:
Thanks a lot @fgvieira! 😄

PR #3305, opened 2024-10-22
fix: Hisat2 index improvements

tavareshugo:
The previous version of this script created an output directory based on the file name prefix of the index file, which resulted in an empty directory being created. 
The current changes make sure that: 

* The output directory is instead created from the directory specified in `snakemake.output`.
* The `params.prefix` is used as a file prefix for the output files.
* The documentation clarifies its use.
* Ran `pytest test.py -v -k test_hisat2_index` which passed

Tagging @simojoe as the original author of the wrapper. 

### QC

## Summary by CodeRabbit

- **New Features**
	- Added a new author, Hugo Tavares, to the HISAT2 index documentation.
	- Enhanced output specifications for HISAT2 index files, allowing for multiple specific file extensions.
	- Improved the indexing process by enabling the handling of multiple FASTA files in a single command.

- **Bug Fixes**
	- Corrected output directory handling to ensure proper file extension generation based on genome size.

- **Documentation**
	- Updated descriptions for input and output sections in the HISAT2 index configuration for better clarity.
	- Added a reference URL for further information on HISAT2 usage.

### DIFF
--- a/bio/hisat2/index/wrapper.py
+++ b/bio/hisat2/index/wrapper.py
@@ -22,14 +22,17 @@ if not "." in fasta:
     input_seq += "-c "
 input_seq += ",".join(fasta) if isinstance(fasta, list) else fasta
 
-hisat_dir = snakemake.params.get("prefix", "")
-if hisat_dir:
-    os.makedirs(hisat_dir)
+# check that file prefix is not a path
+prefix = snakemake.params.get("prefix", "")
+assert "/" not in prefix, "params 'prefix' should be a filename prefix, not a path (do not include '/')."
+
+# make output directory
+os.makedirs(snakemake.output, exist_ok=True)
 
 shell(
     "hisat2-build {extra} "
     "-p {snakemake.threads} "
     "{input_seq} "
-    "{snakemake.params.prefix} "
+    "{snakemake.output}/{prefix} "
     "{log}"
 )
--- a/bio/hisat2/index/wrapper.py
+++ b/bio/hisat2/index/wrapper.py
@@ -27,12 +27,13 @@ prefix = snakemake.params.get("prefix", "")
 assert "/" not in prefix, "params 'prefix' should be a filename prefix, not a path (do not include '/')."
 
 # make output directory
-os.makedirs(snakemake.output, exist_ok=True)
+outdir = snakemake.output[0]
+os.makedirs(outdir, exist_ok=True)
 
 shell(
     "hisat2-build {extra} "
     "-p {snakemake.threads} "
     "{input_seq} "
-    "{snakemake.output}/{prefix} "
+    "{outdir}/{prefix} "
     "{log}"
 )
--- a/bio/hisat2/index/wrapper.py
+++ b/bio/hisat2/index/wrapper.py
@@ -22,49 +22,13 @@ if not "." in fasta:
     input_seq += "-c "
 input_seq += ",".join(fasta) if isinstance(fasta, list) else fasta
 
-
-# Function to validate output file extensions and get common prefix
-def get_prefix(files):
-    # Check all files have the same dirname
-    dirname = os.path.dirname(files[0])
-    if not all(os.path.dirname(f) == dirname for f in files):
-        raise ValueError("Output files must all share the same output directory path.")
-
-    # Valid file extensions from hisat2-build docs
-    ht2_extensions = {f".{i}.ht2" for i in range(1, 9)}
-    ht2l_extensions = {f".{i}.ht2l" for i in range(1, 9)}
-    all_extensions = ht2_extensions | ht2l_extensions
-
-    # Get user file extensions
-    extensions = []
-    for file in files:
-        for ext in all_extensions:
-            if file.endswith(ext):
-                extensions.append(ext)
-
-    # Check all 8 files are specified in the output with the correct extensions
-    if all(ht2 in extensions for ht2 in ht2_extensions) or all(
-        ht2l in extensions for ht2l in ht2l_extensions
-    ):
-        prefix = os.path.commonprefix(list(files)).rstrip(".")
-    else:
-        raise ValueError(
-            "Output files must have either '.ht2' or '.ht2l' extensions for all 8 files."
-        )
-
-    if not prefix:
-        raise ValueError("All output files must share the same prefix.")
-
-    return prefix
-
-
 # get common prefix
-prefix = get_prefix(snakemake.output)
-
-# make output directory
-if os.path.dirname(prefix):
-    os.makedirs(os.path.dirname(prefix), exist_ok=True)
+prefix = os.path.commonprefix(snakemake.output).rstrip(".")
 
 shell(
-    "hisat2-build {extra} " "-p {snakemake.threads} " "{input_seq} " "{prefix} " "{log}"
+    "hisat2-build {extra}"
+    " -p {snakemake.threads}"
+    " {input_seq}"
+    " {prefix}"
+    " {log}"
 )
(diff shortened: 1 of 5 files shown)


fgvieira:
@tavareshugo thanks for fixing these wrappers!

In this case, would it make more sense to drop `params` and infer the prefix from the output files? This would also require for the output files to be ll specified, which would also make the wrapper a bit more explicit.

Taking `bwa-mem` as an example, the index output files could be specified as:
https://github.com/snakemake/snakemake-wrappers/blob/2256246fc6fab81444253119a06c273bd86ee325/bio/bwa/index/test/Snakefile#L4-L5

and the prefix could be inferred as:
https://github.com/snakemake/snakemake-wrappers/blob/2256246fc6fab81444253119a06c273bd86ee325/bio/hisat2/align/wrapper.py#L30

tavareshugo:
Yes, I thought the same using `multiext()`, and it would be great if we could make that work. The problem is that HISAT does the following (from [docs](http://daehwankimlab.github.io/hisat2/manual/), my emphasis):

> hisat2-build builds a HISAT2 index from a set of DNA sequences. hisat2-build outputs a set of 6 files with suffixes .1.ht2, .2.ht2, .3.ht2, .4.ht2, .5.ht2, .6.ht2, .7.ht2, and .8.ht2. **In the case of a large index these suffixes will have a ht2l termination.** 

And: 

>  The user need not worry about whether a particular index is small or large; the wrapper scripts will automatically build and use the appropriate index.

So, the issue is that it's not easy to know whether the output files will have `.ht2` or `.ht2l` extension, making it hard to take the approach used with other tools such as `bwa`. Unless there is some snakemake syntax to specify two alternative output file extensions?

fgvieira:
The extension would not affect the prefix inference part. As for the test case, we know it will be a small index, so it would not be a problem either.

Generally, I'd say that the user needs to account for that and check if it will be a small or large index. Alternatively, the user can specify `--large-index` in params to force a large index.

What do you think?

tavareshugo:
I suppose we could leave it to the user, although that removes some of the automation and flexibility. 
For example, my use case for this wrapper is to pre-index genomes for users in a large server. These genomes can range from small to large and I wanted to have a single HISAT2 rule that works for all genomes. 

But I guess we could do what you say and force `--large-index` for all our genomes...

OK, I'll make the change and update the pull request.

tavareshugo:
@fgvieira I've got something working locally, and the automated test (`pytest`) is passing, but I'd like to test locally and I'm trying: 

```bash
snakemake --cores 1 --use-conda -F --wrapper-prefix /home/hugot/temp/snakemake-wrappers/@hisat2_index_improvements --printshellcmds --show-failed-logs hisat2_index/genome.1.ht2
```

But I get error: `Failed to open source file /home/hugot/temp/snakemake-wrappers/@hisat2_index_improvementsmaster/bio/hisat2/index/environment.yaml`

So, clearly I'm specifying the `--wrapper-prefix` wrong. Would you be able to help me with the syntax here? My local repo is in `/home/hugot/temp/snakemake-wrappers/` and my branch is called `hisat2_index_improvements`.

fgvieira:
Can you try following these instructions:
https://snakemake-wrappers.readthedocs.io/en/stable/contributing.html#testing-locally

tavareshugo:
Hi @fgvieira that works fine, using `pytest test.py -v -k test_hisat2_index` the automated test passes. 
But I wanted to test the local version of the workflow on some toy dataset, because it's easier to examine the output files and really make sure it's doing what I think it is. With `pytest` I can't quite see how to access the output files, I think they go into a temp directory and get removed after.

tavareshugo:
I'm not the one adding `master` to it, it seems like `snakemake` is doing this. If I run what you say, it tries: `Failed to open source file /home/hugot/temp/snakemake-wrappers/master/bio/hisat2/index/environment.yaml`

So, it's adding `master/bio/hisat2/index/environment.yaml`. From the docs: 

> --wrapper-prefix
> Prefix for URL created from wrapper directive (default: https://github.com/snakemake/snakemake-wrappers/raw/). Set this to a different URL to use your fork or a local clone of the repository, e.g., use a git URL like ‘git+file://path/to/your/local/clone@’.
> Default: 'https://github.com/snakemake/snakemake-wrappers/raw/'

I've also tried `--wrapper-prefix 'git+file://home/hugot/temp/snakemake-wrappers/@hisat2_index_improve
ments'` but it errors: 

```
Failed to read source git+file://home/hugot/temp/snakemake-wrappers/@hisat2_index_improvements/bio/hisat2/index@master from git repo.
```

It's concatenating `/bio/hisat2/index@master` to my input. 

(sorry this is now getting off topic, I can open an issue on the snakemake repo if that's better)

fgvieira:
Snakemake concatenates your wrapper prefix to the wrapper path, so just change in the test Snakefile:
```
wrapper:
    "master/bio/hisat2/index"
```
to:
```
wrapper:
    "/bio/hisat2/index"
```

Then run with `--wrapper-prefix file://home/hugot/temp/snakemake-wrappers`.
If all works fine, remember to revert the changes before committing. 😄

tavareshugo:
Oh, I knew it had to be a silly thing like that! 🤦 
OK, it's working. I'll update the pull request in a bit, thanks for your help!

tavareshugo:
I've done a few changes from our discussion:

- wrapper has a function to check for common prefix (ensuring valid extensions were specified)
- output directory is created from the dirname of this prefix
- updated the docs (added myself as additional author)

The automatic `pytest` passes and I've also done some adhoc testing locally trying out different output file specifications to ensure the parser does the job correctly. It all looks like it's working on my side, so hopefully good to merge 🤞

bio/hisat2/index/wrapper.py
```diff
@@ -22,14 +22,49 @@
...
+
+    return prefix
+
+
+# get common prefix
+prefix = get_prefix(snakemake.output)
+
+# make output directory
+if os.path.dirname(prefix):
+    os.makedirs(os.path.dirname(prefix), exist_ok=True)
```
fgvieira, line 66: This should not be necessary, as snakemake should create all needed output directories.

bio/hisat2/index/wrapper.py
```diff
@@ -22,14 +22,49 @@
...
 
 shell(
-    "hisat2-build {extra} "
-    "-p {snakemake.threads} "
-    "{input_seq} "
-    "{snakemake.params.prefix} "
-    "{log}"
+    "hisat2-build {extra} " "-p {snakemake.threads} " "{input_seq} " "{prefix} " "{log}"
```
fgvieira, line 69:
```suggestion
    "hisat2-build {extra}"
    " -p {snakemake.threads}"
    " {input_seq}"
    " {prefix}"
    " {log}"
```

bio/hisat2/index/wrapper.py
```diff
@@ -22,14 +22,49 @@
...
+        )
+
+    if not prefix:
+        raise ValueError("All output files must share the same prefix.")
+
+    return prefix
+
+
+# get common prefix
+prefix = get_prefix(snakemake.output)
```
fgvieira, line 62: Why not just?
```suggestion
prefix = os.common_prefix(snakemake.output).rstrip(.)
```

If there are any inconsistency on dirnames/prefix/extensions, then the rule will fail since some (or all) output files will be missing, no?

bio/hisat2/index/wrapper.py
tavareshugo: yes of course, I was doing it based on the previous version of the code. I'll remove these

bio/hisat2/index/wrapper.py
tavareshugo: I think this is what's triggering the warning from `black` now. But I agree it's a bit more readable and fits the other wrappers.

bio/hisat2/index/wrapper.py
tavareshugo: made the simplification - does look much nicer to look at :)

fgvieira (approved):

PR #3432, opened 2024-11-06
feat: add expected output checking to `test_wrapper.py` `run()` function (fix: ensembl-reference wrapper download >1 chromosome)

dlaehnemann:
This PR evolved. I also added expected output checking for wrapper tests, as this was the reason we previously didn't catch the other fixed functionality not working.

Previously, the `ensembl-sequence` wrapper would only download the first chromosome, if a list of chromosomes was specified.

### QC

## Summary by CodeRabbit

- **New Features**
	- Enhanced error messages for clarity when selecting individual chromosomes.
	- Improved control flow for downloading sequence data, allowing for a more efficient exit from the loop.
	- New instructions for testing contributions, including expected output file comparisons.

- **Bug Fixes**
	- Updated error messages for unsuccessful downloads to provide more specific feedback.
	- Maintained existing error handling for invalid datatype values to ensure robust performance.

- **Tests**
	- Updated test case to reflect changes in input and output file names, ensuring accurate validation of results.


bio/reference/ensembl-sequence/wrapper.py
```diff
@@ -47,15 +47,15 @@
...
+            "Invalid datatype. To select individual chromosomes, the datatype must be dna."
         )
 
 url = snakemake.params.get("url", "ftp://ftp.ensembl.org/pub")
 spec = spec.format(build=build, release=release)
 url_prefix = f"{url}/{branch}release-{release}/fasta/{species}/{datatype}/{species.capitalize()}.{spec}"
 
-success = False
 for suffix in suffixes:
+    success = False
```
fgvieira, line 58: Shouldn't this be outside the loop to check if at least one suffix was successful?

dlaehnemann: If it is outside of the loop, and we are requesting multiple chromosomes, this will turn true on any working chromosome, and then stay that way. So we will not get any debugging output and error thrown, in case any of the chromosomes is not available. So for the chromosomes case, we should reset this for every `suffix in suffixes`. For the other case, checking whether `"dna.primary_assembly.fa.gz"` or `"dna.toplevel.fa.gz"` is available, it will `break` out of the `suffix in suffixes` loop right after setting `success = True` and will otherwise be left with `success = False` after the last suffix that runs into the `except:`.

fgvieira: Ok, I thought the idea of `success` was to check that at least one succeeded.

fgvieira: But this way it only checks if the last chromosome was available, no?

dlaehnemann: Ah yes, you are right. Very good catch. Let me think about what the best solution is...

fgvieira: Maybe moving the error checking directly to the try/except?

bio/reference/ensembl-sequence/wrapper.py
```diff
@@ -47,33 +47,45 @@
...
         shell("curl -sSf {url} > /dev/null 2> /dev/null")
     except sp.CalledProcessError:
-        continue
+        if chromosome:
+            print(
+                f"Unable to download the requested chromosome sequence from Ensembl at: {url_prefix}.{suffix}.",
+                file=sys.stderr,
+            )
+            break
+        else:
```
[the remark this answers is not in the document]
dlaehnemann: The idea of the suggestion is good, but the suggested syntax does not exist. The shell module has no method logger. This lives in `snakemake.logging.logger`. Also, the `continue` instead of the `break` would actually change code behaviour.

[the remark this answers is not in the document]
dlaehnemann: The `break` here is valid, as we don't need to test all remaining chromosomes if one fails. There is no chance of generating valid output, then.

bio/reference/ensembl-sequence/wrapper.py
```diff
@@ -47,33 +47,45 @@
...
+            "Invalid datatype. To select individual chromosomes, the datatype must be dna."
         )
 
 url = snakemake.params.get("url", "ftp://ftp.ensembl.org/pub")
 spec = spec.format(build=build, release=release)
 url_prefix = f"{url}/{branch}release-{release}/fasta/{species}/{datatype}/{species.capitalize()}.{spec}"
 
-success = False
 for suffix in suffixes:
+    success = False
```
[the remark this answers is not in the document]
dlaehnemann: If one chromosome fails, there is no chance of getting valid output and the user will have to fix. So no need to track more than one failure.

bio/reference/ensembl-sequence/wrapper.py
dlaehnemann: This should be resolved.

dlaehnemann:
@fgvieira This is now ready for another round of review. I also changed the title, because I used the concrete example here to implement expected output file testing in `test_wrappers.py`.

test_wrappers.py
```diff
@@ -5381,10 +5385,13 @@ def test_ensembl_sequence_chromosome(run):
...
 
-def test_ensembl_sequence_chromosomes(run):
+def test_ensembl_sequence_multiple_chromosomes(run):
     run(
         "bio/reference/ensembl-sequence",
-        ["snakemake", "--cores", "1", "refs/chr1_and_chr2.fasta", "--use-conda", "-F"],
+        ["snakemake", "--cores", "1", "refs/chr6_and_chr1.fasta", "--use-conda", "-F"],
+        expected_comparisons={
+            "refs/chr6_and_chr1.fasta": "expected/chr6_and_chr1.fasta"
+        }
```
johanneskoester, line 5394:
```suggestion
        compare_results={
            "refs/chr6_and_chr1.fasta": "expected/chr6_and_chr1.fasta"
        }
```

johanneskoester (changes requested):
Nice idea! I think the argument name was not so intuitive. I am not sure whether it is better with my suggestion. What do you think?

test_wrappers.py
dlaehnemann: I'm opting for the longer, but more descriptive `compare_results_with_exppected`.

johanneskoester (approved):

PR #3518, opened 2024-12-03
feat: bowtie2 can now sort with samtools and picard

jlanga:
This PR shamelessly copies functionalities from bwa-mem2 to sort or not, do it by coordinate or queryname, and choose between samtools and picard to do so

### QC


bio/bowtie2/align/environment.yaml
```diff
@@ -3,7 +3,7 @@ channels:
   - bioconda
   - nodefaults
 dependencies:
-  - bowtie2 =2.5.4
-  - picard-slim =3.3.0
+  - bowtie2 =2.5
```
fgvieira, line 6: For versions control
```suggestion
  - bowtie2 =2.5.4
```

bio/bowtie2/align/environment.yaml
```diff
@@ -3,7 +3,7 @@ channels:
...
+  - picard-slim =3.3
   - samtools =1.21
-  - snakemake-wrapper-utils =0.6.2
+  - snakemake-wrapper-utils =0.6
```
fgvieira, line 9:
```suggestion
  - snakemake-wrapper-utils =0.6.2
```

bio/bowtie2/align/environment.yaml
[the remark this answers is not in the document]
jlanga: Forgot to push. It was already solved. Sorry.

bio/bowtie2/align/environment.yaml
jlanga: Forgot to push. It was already solved. Sorry.

jlanga:
I'll continue next week

jlanga:
@fgvieira I think I am done with this wrapper.

I've left commented the part about outputting extra files from bowtie2 (metrics, unaligned, unpaired, concordant and unconcordant). Some of them introduce a lot of complexity in the code because bowtie can output 1 or 2 fastq files depending on if the input is SE or PE. These files can be easily obtained with `samtools view` and `samtools fastq`. Also if you want gzipped files you have to use different flags (`--un-conc` becomes `--un-conc-gz` or `--un-conc-bz2`). It is a headache.

Let me know if you want me to keep that functionality on to uncomment it.

fgvieira:
It seems you still have un-pushed commits. Can you push them?

jlanga:
I don't understand. All my commits are on the `bowtie2_sort` branch. I have nothing else pending to push in my computer.

jlanga:
I clicked the Update Branch button. I hope it was that.

jlanga:
I see that we both removed the patch number, but in a comment later you appended the one for `picard-slim`:

```
- picard-slim =3.3.0
```

Is that correct? That one has to be fixed to 3.3.0?

jlanga:
Ok. I see other wrappers with picard-slim =3.3.0. Fixing that and repinning everything.

fgvieira:
but also bowtoe2 and wrapper utils

jlanga:
Ok. Pinned everything with the exception of samtools. The pin file stays the same after that.

fgvieira:
Why the total re-write of the wrapper? As I see it, you only needed to add the logic for `picard`, like (for example) in the `bwa` wrapper:
https://github.com/snakemake/snakemake-wrappers/blob/7b6aa5ce164a4f1c45716ae9c56dbe6d330c08f3/bio/bwa/mem/wrapper.py#L43-L66

and change the line:
https://github.com/snakemake/snakemake-wrappers/blob/7b6aa5ce164a4f1c45716ae9c56dbe6d330c08f3/bio/bowtie2/align/wrapper.py#L97

jlanga:
My initial plan was to just do either samtools view or sort, but I saw the wrapper from [bwa-mem2](https://github.com/snakemake/snakemake-wrappers/blob/7b6aa5ce164a4f1c45716ae9c56dbe6d330c08f3/bio/bwa-mem2/mem/wrapper.py) and I saw that it could be done for sam, bam and cram, sorted and unsorted, and with samtools and picard.

But bwa-mem2 only aligns and generates the sam file. Bowtie2 can produce extra outputs.

And now you brought to my attention `fgbio SortSam`.

And there are wrappers specific to a single sort program like [bwa-mem2 and samblaster](https://github.com/snakemake/snakemake-wrappers/blob/master/bio/bwa-mem2/mem-samblaster/wrapper.py)

What is the official view from snakemake wrappers? A monolithic wrapper htat handles multiple tool combinations, or multiple silly wrappers?

If we go through the silly wrappers, we can delete 100+ lines of code just for validation.

fgvieira:
If it is just different sorting programs, I'd say a single wrapper.

johanneskoester:
I would view this from the user perspective. Since it is so common to sort after alignment, it is nice to have a reasonable default for that direcly built into a wrapper. But we don't need to support all combinations. Just the current state of the art (i.e. the fastest sorter) that supports all relevant output files (bam and cram).

PR #3648, opened 2025-02-11
fix: remove nanosim TODO by using new pre-trained models and latest version of nanosim in the wrapper

dlaehnemann:
This pull request leverages two things:

1. There are recent pre-trained models, generated with `nanosim` `3.2`, which should be compatible with the latest versions of `nanosim` itself. These come from these two pull requests:
  * https://github.com/bcgsc/NanoSim/pull/224
  * https://github.com/bcgsc/NanoSim/pull/238
2. In addition, we could patch the bioconda recipe for `nanosim` a bit and update it to the latest version:
  * https://github.com/bioconda/bioconda-recipes/pull/53528

With `scikit-learn` fixed to the version used for training those recent models, we can now update this wrapper to use these latest models for testing and the wrapper should in general work with those recent models.
 

### QC

## Summary by CodeRabbit

- **New Features**
  - Expanded simulation workflows with additional model inputs and updated training datasets across genome, transcriptome, and metagenome processes.

- **Chores**
  - Upgraded `nanosim` dependency to version 3.2.3 and removed outdated compatibility constraints for `numpy`.

- **Tests**
  - Revised simulation output references in tests to align with the new dataset paths and updated directory structures.


dlaehnemann:
@fgvieira This is a follow-up to your [comments in a previous nanosim wrapper PR](https://github.com/snakemake/snakemake-wrappers/pull/3165#issuecomment-2348361174), so maybe you're interested in reviewing this.

fgvieira:
It looks much better now! 😄 
Just wondering if we should place the test [brca2.expression_abundance.tsv](https://github.com/snakemake/snakemake-wrappers/blob/fix/update-nanosim-recent-models/bio/nanosim/simulator/test/brca2.expression_abundance.tsv), `config` and `resources` together, so things are a bit cleaner.

dlaehnemann:
This was actually intentional, as it follows our current best practices structure for workflows (and in this case the test is a mini-workflow, with the downloading step):
https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html

But if you feel that this is too confusing, we can also flatten this out.

fgvieira:
I don't have a good overview of `nanosim`, but from the test examples it seems that `brca2.expression_abundance.tsv` and `config/sample_x.abundances.tsv` are just input files, no? Like e.g. `resources/brca2.genome.fa`.

dlaehnemann:
Ah yeah, at least `brca2.expression_abundance.tsv` and `config/sample_x.abundances.tsv` should be consistently placed somewhere. Whether they are `config/` is probably a matter of opinion.
In the simulation, this is is something the user needs to provide, so a workflow would optimally determine this from some configuration -- either by curating these files manually, which would then make them eligible for placement in the `config/` folder, I'd say; or by creating those files in a rule (for example from real data), which would put them under `results/`.
So my suggestion would be to also put the `brca2.expression_abundance.tsv` file into the `config/` folder. What do you think?

Otherwise, I would keep the references in the `resources/` folder, as this is something that is usually downloaded by an encompassing workflow, and this way we reflect those best practices in our example rules. Same with the model downloads.
And the model downloads might actually warrant a separate wrapper in itself, but for now these wrappers do what I need. So I'd keep this for another time, if anybody wants to reuse them...

fgvieira (approved):

dlaehnemann:
As always, many thanks for the review!

PR #3981, opened 2025-04-04
feat: Add mmseqs2 main workflows

fgvieira:
### QC


bio/mmseqs2/db/environment.yaml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - conda-forge
+  - bioconda
+  - nodefaults
+dependencies:
+  - mmseqs2 =17.b804f
```
dlaehnemann: Should we switch to the latest version?
```suggestion
  - mmseqs2 =18.8cc5c
```

bio/mmseqs2/workflows/environment.yaml
```diff
@@ -0,0 +1,7 @@
+channels:
+  - conda-forge
+  - bioconda
+  - nodefaults
+dependencies:
+  - mmseqs2 =17.b804f
```
dlaehnemann: Should we switch to the latest version?
```suggestion
  - mmseqs2 =18.8cc5c
```

bio/mmseqs2/db/environment.yaml
dlaehnemann: If you do, don't forget updating the pinning file(s)...

bio/mmseqs2/db/wrapper.py
```diff
@@ -0,0 +1,31 @@
...
+
+with tempfile.TemporaryDirectory() as tmpdir:
+    # Modules with threads
+    if snakemake.params.module in ["databases"]:
+        extra = f"--threads {snakemake.threads} {extra}"
+    # Modules with no temp folder
+    if snakemake.params.module in ["createdb"]:
+        tmpdir = ""
+
+    shell("mmseqs {snakemake.params.module} {input} {out} {tmpdir} {extra} {log}")
```
dlaehnemann: I think you want to directly get the `snakemake.output.db` and then use this for the output folder and basename parsing. In doing so, I think we can always assume that this will be a list for this database creation, right?

Also, I would exactly match the module and raise an error, if it doesn't match one of the two known modules. Although I am not sure if it might implicitly handle more than the two database modules mentioned here?

And I would directly implement the arbitrary output files.

Here's a suggestion, but I haven't tested if this works:
```suggestion
[out_folder, out_basename] = os.path.split(os.path.commonprefix(snakemake.output.db))

with tempfile.TemporaryDirectory() as tmp_dir and tempfile.TemporaryDirectory() as tmp_out:
    # Modules with threads
    if "databases" in snakemake.params.module:
        extra = f"--threads {snakemake.threads} {extra}"
    # Modules with no temp folder
    elif snakemake.params.module == "createdb":
        tmpdir = ""
    else:
        raise ValueError("The module specified under params is not known for the db wrapper: '{snakemake.params.module}'.")

    shell(
        f"( mmseqs {snakemake.params.module} {input} {tmp_out}/{out_basename} {tmp_dir} {extra};"
        f"mv {tmp_out) {out_folder} {log}"
    )
```

Also, at least the arbitrary output files mechanism also applies for the `workflows` wrapper.

bio/mmseqs2/db/meta.yaml
```diff
@@ -0,0 +1,13 @@
...
+description: |
+  ultra fast and sensitive sequence search and clustering suite
+authors:
+  - Filipe G. Vieira
+input:
+  - input FAS file
+output:
+  - output: DB files
+params:
+  - module: workflow to use
```
dlaehnemann: I would explicitly list the available modules here:

```suggestion
  - module: With this wrapper, you can either `download existing databases <https://github.com/soedinglab/mmseqs2/wiki#downloading-databases>`_ with the ``database`` module (see the linked documentation for what is available) or `create a new database <https://github.com/soedinglab/mmseqs2/wiki#creating-a-seqtaxdb>`_ with the ``createdb`` and ``createtaxdb`  modules. 
```

bio/mmseqs2/db/test/Snakefile
```diff
@@ -0,0 +1,49 @@
...
+            "_h.index",
+        ),
+    log:
+        "logs/createdb/{sample}.log",
+    params:
+        module="createdb",
+        extra="-v 3",
+    threads: 1
+    wrapper:
+        "master/bio/mmseqs2/db"
```
dlaehnemann, line 68: From what I understand, you also need the `createtaxdb` module to create a full database:
https://github.com/soedinglab/mmseqs2/wiki#creating-a-seqtaxdb

Or was it intentional to not include this module here?

fgvieira: As far as I remember, `createtaxdb` is not very `snakemake` friendly, since it changes the DB in place. Meaning that the input and output are the same. However, I've added an example with a `touch()` file as flag.

bio/mmseqs2/db/wrapper.py
fgvieira: Not sure why I added that note, but I think it should actually work to specify an arbitrary output DB name, no?
Maybe I forgot to clean it up... can you double check?

dlaehnemann: Ah, sure. If we just do the `os.path.commonprefix(snakeamke.otuput.db)` and pass it to msseqs as `{out}`, that should also be fine.

bio/mmseqs2/db/test/Snakefile
dlaehnemann: Ah yes, annoying pattern. But sounds like a reasonable solution.

bio/mmseqs2/workflows/meta.yaml
```diff
@@ -0,0 +1,14 @@
...
+  ultra fast and sensitive sequence search and clustering suite
+authors:
+  - Filipe G. Vieira
+input:
+  - query: input query FAS file(s)
+  - target: input target FAS file(s) or DB
+output:
+  - output: FAS, cluster or DB file(s)
+params:
+  - module: workflow to use
```
dlaehnemann: I would also add the supported modules for this wrapper, with linkouts to the respective docs. This makes it a lot easier to know what the wrapper supports and how to tune it with `params: extra=`, if necessary.

bio/mmseqs2/workflows/wrapper.py
```diff
@@ -0,0 +1,37 @@
...
+
+extra = snakemake.params.get("extra", "")
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+
+target = snakemake.input.get("target", "")
+if isinstance(target, list):
+    target = os.path.commonprefix(snakemake.input.target)
+
+# TODO: arbitrary output file names
```
dlaehnemann: I think this wrapper also allows arbitrary output file names already, right?

bio/mmseqs2/workflows/wrapper.py
```diff
@@ -0,0 +1,37 @@
...
+out = snakemake.output
+if isinstance(out, list):
+    out = os.path.commonprefix(out).rstrip("_")
```
dlaehnemann: I think, the `rstrip("_")` is dangerous, here, as the common prefix might contain this symbol and it should be kept. But just removing the `rstrip()` should be fine:

```suggestion
    out = os.path.commonprefix(out)
```

bio/mmseqs2/db/test/Snakefile
```diff
@@ -0,0 +1,67 @@
...
+
+
+rule mmseqs2_createtaxdb:
+    input:
+        db="out/createdb/{sample}/",
+        tax_dump=storage.http("http://ftp.ncbi.nlm.nih.gov/pub/taxonomy/taxdump.tar.gz"),
+        tax_map="seqs/{sample}.map",
+    output:
+        # Touch empty file since createtaxdb edits DB in place
```
dlaehnemann:
```suggestion
        # Touch empty file since createtaxdb edits DB in place.
        # Make sure to include this file as input in rules that need the annotated database.
```

PR #4121, opened 2025-05-16
feat: add wrapper for MEGAHIT

alienzj:
add wrapper for MEGAHIT, which is an ultra-fast and memory-efficient NGS assembler. It is optimized for metagenomes, but also works well on generic single genome assembly (small or mammalian size) and single-cell assembly.

### QC

## Summary by CodeRabbit

- **New Features**
  - Added support for the MEGAHIT assembler, including environment setup and metadata.
  - Introduced a Snakemake workflow to automate downloading test data and running MEGAHIT assembly.
  - Provided a wrapper script for seamless integration of MEGAHIT within Snakemake workflows.

- **Tests**
  - Added automated tests to verify MEGAHIT wrapper functionality.

### DIFF
--- a/bio/megahit/meta.yaml
+++ b/bio/megahit/meta.yaml
@@ -1,0 +1,23 @@
+name: "megahit"
+
+url: https://github.com/voutcn/megahit
+
+description: |
+  MEGAHIT is an ultra-fast and memory-efficient NGS assembler. It is optimized for metagenomes, but also works well on generic single genome assembly (small or mammalian size) and single-cell assembly.
+  Input options that can be specified for multiple times (supporting plain text and gz/bz2 extensions).
+
+input:
+  - reads: list of reads in FASTQ format
+  - r1: forward reads
+  - r2: reverse reads
+  - interleaved: interleaved reads
+  - unpaired: unpaired reads
+
+output:
+  - contigs: output file with contigs
+  - log: log file
+  - json: options json file
+
+authors:
+  - Jie Zhu
+  - Filipe G. Vieira
--- a/bio/megahit/test/Snakefile
+++ b/bio/megahit/test/Snakefile
@@ -5,7 +5,7 @@ rule run_megahit:
     input:
         reads=["test_reads/sample1_R1.fastq.gz", "test_reads/sample1_R2.fastq.gz"],
     output:
-        contigs="assembly/contigs.fasta",
+        contigs="assembly/final.contigs.fasta",
     benchmark:
         "logs/benchmarks/assembly/megahit.txt"
     params:
--- a/bio/megahit/test/Snakefile
+++ b/bio/megahit/test/Snakefile
@@ -1,0 +1,29 @@
+container: "docker://continuumio/miniconda3:4.4.10"
+
+
+rule run_megahit:
+    input:
+        reads=["test_reads/sample1_R1.fastq.gz", "test_reads/sample1_R2.fastq.gz"],
+    output:
+        contigs="assembly/contigs.fasta",
+    benchmark:
+        "logs/benchmarks/assembly/megahit.txt"
+    params:
+        # all parameters are optional
+        extra="--min-count 10 --k-list 21,29,39,59,79,99,119,141",
+    log:
+        "logs/megahit.log",
+    threads: 8
+    resources:
+        mem_mb=250000,
+    wrapper:
+        "master/bio/megahit"
+
+
+rule download_test_reads:
+    output:
+        ["test_reads/sample1_R1.fastq.gz", "test_reads/sample1_R2.fastq.gz"],
+    log:
+        "logs/download.log",
+    shell:
+        "(wget -O - https://zenodo.org/record/3992790/files/test_reads.tar.gz | tar -xzf -) > {log} 2>&1"
--- a/bio/megahit/wrapper.py
+++ b/bio/megahit/wrapper.py
@@ -1,0 +1,73 @@
+"""Snakemake wrapper for megahit."""
+
+__author__ = "Jie Zhu @alienzj"
+__copyright__ = "Copyright 2025, Jie Zhu"
+__email__ = "alienchuj@gmail.com"
+__license__ = "MIT"
+
+import os, tempfile, shutil
+from snakemake.shell import shell
+from snakemake_wrapper_utils.snakemake import get_mem
+
+# get output_dir and files from output
+output_dir = os.path.split(snakemake.output[0])[0]
+contigs_file = snakemake.output.get("contigs", os.path.join(output_dir, "contigs.fa"))
+contigs_file_original = os.path.join(output_dir, "final.contigs.fa")
+options_file = snakemake.output.get("options", os.path.join(output_dir, "options.json"))
+log_file = snakemake.output.get("log", os.path.join(output_dir, "log"))
+
+# parse params
+extra = snakemake.params.get("extra", "")
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+memory_requirements = get_mem(snakemake, out_unit="KiB") * 1024
+
+# parse short reads
+if hasattr(snakemake.input, "reads"):
+    reads = snakemake.input.reads
+else:
+    reads = snakemake.input
+
+input_arg = ""
+
+# handle named inputs if available
+if hasattr(snakemake.input, "r1") and hasattr(snakemake.input, "r2"):
+    input_arg += " -1 {} -2 {} ".format(snakemake.input.r1, snakemake.input.r2)
+elif len(reads) >= 2:
+    input_arg += " -1 {} -2 {} ".format(reads[0], reads[1])
+
+# handle interleaved reads if specified
+if hasattr(snakemake.input, "interleaved"):
+    input_arg += " --12 {} ".format(snakemake.input.interleaved)
+elif len(reads) >= 3 and not hasattr(snakemake.input, "r1"):
+    input_arg += " --12 {} ".format(reads[2])
+
+# handle additional reads if specified
+if hasattr(snakemake.input, "unpaired"):
+    input_arg += " --read {} ".format(snakemake.input.unpaired)
+elif len(reads) >= 4 and not hasattr(snakemake.input, "r1"):
+    input_arg += " --read {} ".format(reads[3])
+
+
+with tempfile.TemporaryDirectory(dir=os.path.dirname(output_dir)) as temp_dir:
+    output_temp_dir = os.path.join(temp_dir, "temp")
+
+    shell(
+        "megahit "
+        " -t {snakemake.threads} "
+        " -m {memory_requirements} "
+        " -o {output_temp_dir} "
+        " {input_arg} "
+        " {extra} "
+        " > {snakemake.log[0]} 2>&1 "
+    )
+
+    if os.path.exists(os.path.join(output_temp_dir, "done")):
+        shell("rm -rf {output_dir}")
+        shutil.move(output_temp_dir, output_dir)
+
+        if (
+            os.path.exists(contigs_file_original)
+            and os.path.exists(options_file)
+            and os.path.exists(log_file)
+        ):
+            shutil.move(contigs_file_original, contigs_file)
--- a/bio/megahit/wrapper.py
+++ b/bio/megahit/wrapper.py
@@ -58,7 +58,7 @@ with tempfile.TemporaryDirectory(dir=os.path.dirname(output_dir)) as temp_dir:
         " -o {output_temp_dir} "
         " {input_arg} "
         " {extra} "
-        " > {snakemake.log[0]} 2>&1 "
+        " {log}"
     )
 
     if os.path.exists(os.path.join(output_temp_dir, "done")):
--- a/bio/megahit/wrapper.py
+++ b/bio/megahit/wrapper.py
@@ -41,8 +41,7 @@ elif len(reads) >= 4 and not hasattr(snakemake.input, "r1"):
     input_arg += " --read {} ".format(reads[3])
 
 # run megahit
-output_dir = os.path.dirname(snakemake.output.get("contigs"))
-with tempfile.TemporaryDirectory(dir=os.path.dirname(output_dir)) as temp_dir:
+with tempfile.TemporaryDirectory() as temp_dir:
     output_temp_dir = os.path.join(temp_dir, "temp")
 
     shell(
--- a/bio/megahit/wrapper.py
+++ b/bio/megahit/wrapper.py
@@ -1,24 +1,17 @@
 """Snakemake wrapper for megahit."""
 
-__author__ = "Jie Zhu @alienzj"
-__copyright__ = "Copyright 2025, Jie Zhu"
+__author__ = "Jie Zhu @alienzj, Filipe G. Vieira @fgvieira"
+__copyright__ = "Copyright 2025, Jie Zhu, Filipe G. Vieira"
 __email__ = "alienchuj@gmail.com"
 __license__ = "MIT"
 
-import os, tempfile, shutil
+import os, tempfile
 from snakemake.shell import shell
 from snakemake_wrapper_utils.snakemake import get_mem
 
-# get output_dir and files from output
-output_dir = os.path.split(snakemake.output[0])[0]
-contigs_file = snakemake.output.get("contigs", os.path.join(output_dir, "contigs.fa"))
-contigs_file_original = os.path.join(output_dir, "final.contigs.fa")
-options_file = snakemake.output.get("options", os.path.join(output_dir, "options.json"))
-log_file = snakemake.output.get("log", os.path.join(output_dir, "log"))
-
 # parse params
 extra = snakemake.params.get("extra", "")
-log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+log = snakemake.log_fmt_shell(stdout=True, stderr=True, append=True)
 memory_requirements = get_mem(snakemake, out_unit="KiB") * 1024
 
 # parse short reads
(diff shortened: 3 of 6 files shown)


test_wrappers.py
```diff
@@ -6398,6 +6398,24 @@ def test_metaspades(run):
...
+        [
+            "snakemake",
+            "run_megahit",
+            "--cores",
+            "2",
+            "--use-conda",
+            "--resources",
+            "mem_mem=1000",
+            "time=15",
+            "--show-failed-logs",
```
fgvieira, line 6413:
```suggestion
```

bio/megahit/test/Snakefile
```diff
@@ -0,0 +1,31 @@
+container: "docker://continuumio/miniconda3:4.4.10"
+
```
fgvieira, line 2:
```suggestion
```

bio/megahit/test/Snakefile
```diff
@@ -0,0 +1,31 @@
...
+    input:
+        reads=["test_reads/sample1_R1.fastq.gz", "test_reads/sample1_R2.fastq.gz"],
+    output:
+        contigs="assembly/contigs.fasta"
+    benchmark:
+        "logs/benchmarks/assembly/megahit.txt"
+    params:
+        # all parameters are optional
+        klist="21,29,39,59,79,99,119,141",
+        extra="--min-count 10"
```
fgvieira, line 14: Why not"
```suggestion
        extra="--min-count 10 --klist 21,29,39,59,79,99,119,141"
```

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,97 @@
...
+            input_arg += " --read {3}".format(*reads)
+
+    shell("rm -rf {output_dir}")
+
+    shell(
+        "megahit "
+        " -t {snakemake.threads} "
+        " -m {memory_requirements} "
+        " -o {output_dir} "
+        " --k-list {klist} "
```
fgvieira, line 60:
```suggestion
```

fgvieira (changes requested):

bio/megahit/test/Snakefile
```diff
@@ -0,0 +1,29 @@
...
+
+
+rule download_test_reads:
+    output:
+        ["test_reads/sample1_R1.fastq.gz", "test_reads/sample1_R2.fastq.gz"],
+    log:
+        "logs/download.log",
+    shell:
+        " wget https://zenodo.org/record/3992790/files/test_reads.tar.gz >> {log} 2>&1 ; "
+        " tar -xzf test_reads.tar.gz >> {log} 2>&1"
```
fgvieira, line 29:
```suggestion
        " (wget -O - https://zenodo.org/record/3992790/files/test_reads.tar.gz | tar -xzf -) > {log} 2>&1"
```

bio/megahit/test/Snakefile
fgvieira: Is this container needed?

bio/megahit/meta.yaml
```diff
@@ -0,0 +1,7 @@
+name: "megahit"
+url: https://github.com/voutcn/megahit
+description: |
+  MEGAHIT is an ultra-fast and memory-efficient NGS assembler. It is optimized for metagenomes, but also works well on generic single genome assembly (small or mammalian size) and single-cell assembly.
+  Input options that can be specified for multiple times (supporting plain text and gz/bz2 extensions).
+authors:
+  - Jie Zhu
```
fgvieira, line 7:
```suggestion
  - Jie Zhu
input:
  - reads: list of reads in FASTQ format
output:
  - contigs: output file with contigs
```

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,107 @@
...
+    output_dir = snakemake.output.dir
+
+else:
+    # get output_dir file from output
+    if hasattr(snakemake.output, "contigs"):
+        output_file = snakemake.output.contigs
+    else:
+        output_file = snakemake.output[0]
+
+    output_dir = os.path.split(output_file)[0]
```
fgvieira, line 24: I think it is best that all output files have to be specified, so that `snakemake` can check their existence.

```suggestion
output_dir = os.path.split(snakemake.output[0])[0]
```

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+else:
+    # options.json file exitst already I restart from previous run
+
+    shell(
+        "echo '\n\nRestart MEGAHIT \n Remove pipline_state file copy files to force copy files if necessary.' >> {log[0]}"
+    )
+
+    shell("megahit " " --continue " " -o {output_dir} " " >> {snakemake.log[0]} 2>&1 ")
```
fgvieira, line 84: Not sure it makes sense to support re-running, since `snakemake` deletes all output files if it fails.
```suggestion
```

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,107 @@
...
+    elif len(reads) >= 3 and not hasattr(snakemake.input, "r1"):
+        input_arg += " --12 {} ".format(reads[2])
+
+    # handle additional reads if specified
+    if hasattr(snakemake.input, "unpaired"):
+        input_arg += " --read {} ".format(snakemake.input.unpaired)
+    elif len(reads) >= 4 and not hasattr(snakemake.input, "r1"):
+        input_arg += " --read {} ".format(reads[3])
+
+    shell("rm -rf {output_dir}")
```
fgvieira, line 64: `Snakemake` would clean-up the folder.
```suggestion
```

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,107 @@
...
+
+# parse params
+extra = snakemake.params.get("extra", "")
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+if hasattr(snakemake.resources, "mem_mb"):
+    mem_bp = snakemake.resources.mem_mb * 1000000
+    memory_requirements = f"{mem_bp}"
+else:
+    memory_requirements = "0.9"
```
fgvieira, line 35:
```suggestion
memory_requirements = get_mem(snakemake, out_unit="KiB") / 1024
```

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,107 @@
+"""Snakemake wrapper for megahit."""
+
+__author__ = "Jie Zhu @alienzj"
+__copyright__ = "Copyright 2025, Jie Zhu"
+__email__ = "alienchuj@gmail.com"
+__license__ = "MIT"
+
+import os, shutil
+from snakemake.shell import shell
```
fgvieira, line 9:
```suggestion
from snakemake.shell import shell
from snakemake_wrapper_utils.snakemake import get_mem
```

fgvieira (changes requested):

bio/megahit/test/Snakefile
alienzj: Great!

bio/megahit/wrapper.py
alienzj: `Snakemake` only removes files specified in the `output` section if they were generated before the rule failed.
Adding support for `MEGAHIT`'s continuation mode would improve workflow robustness.

bio/megahit/wrapper.py
alienzj: This shell command is necessary; otherwise, `MEGAHIT` will fail.

bio/megahit/test/Snakefile
alienzj: This was required to avoid a `snakemake` lint error.

bio/megahit/wrapper.py
alienzj: The `--memory` option in `MEGAHIT` accepts values in `base pairs (bp)` as memory units.

alienzj: From `MEGAHIT`  documentation:
```
Hardware options:
    -m/--memory   <float> max memory in byte to be used in SdBG construction
                  (if set between 0-1, fraction of the machine's total memory) [0.9]
```

fgvieira: According to the documentation, the memory is in `bytes`.

alienzj: The `--memory` parameter expects values in base pairs (bp). As shown in https://github.com/voutcn/megahit/blob/master/src/megahit#L596 and https://github.com/voutcn/megahit/blob/master/src/megahit#L234, memory should be calculated as:
`<GB_value> × 1024³` (i.e., GB to bytes conversion).

Example:
```
For 10GB allocation:
10 * 1024 * 1024 * 1024 = 10737418240 bp
```

fgvieira: The value returned by `return int(float(os.popen("free").readlines()[1].split()[1]) * 1024)` is in bytes (not base pairs).

bio/megahit/wrapper.py
fgvieira: If `megahit` does not allow for the output directory to exist, use `tempfile` and move files to their final destination when `megahit` finishes. You can see an example in:
https://github.com/snakemake/snakemake-wrappers/blob/63f5e877e2191e6f17871e97d4cb6a353f85ef30/bio/biobambam2/bamsormadup/wrapper.py#L34-L38

bio/megahit/wrapper.py
fgvieira: I don't think support for resume runs is advisable as it might lead to corrupted runs, if a re-run uses files that are the result of a failed run. On top of that, benchmark stats are no longer valid.

bio/megahit/wrapper.py
fgvieira: Actually, if you use `tempfile`, there is no need to define `output_dir`.

bio/megahit/wrapper.py
alienzj: Yes. It's bytes.

bio/megahit/wrapper.py
alienzj: I don’t think relying on `/tmp` is practical for large metagenomic assemblies because temporary data can grow very large, and we can’t guarantee that `/tmp` has sufficient storage space when using `tmepfile` method.

bio/megahit/wrapper.py
alienzj: Yes, when rerun, benchmark stats is not true.
OK, I will remove this part code.

bio/megahit/wrapper.py
alienzj: So the code should be `memory_requirements = get_mem(snakemake, out_unit="KiB") * 1024` ?

bio/megahit/wrapper.py
alienzj: Based on the `megahit` output in the assembly directory, I believe the `contigs.fasta` file alone would be sufficient (or: specifying just the contigs output directory would be enough too).

```
λ ll assembly
  inode Permissions  Size User    Group Date Modified Date Accessed Name
3666787 .rw-r-----    262 alienzj users 20 May 11:09  20 May 11:08  checkpoints.txt
3666976 .rw-r-----  902Ki alienzj users 20 May 11:09  20 May 11:09  contigs.fasta
3666977 .rw-r-----      0 alienzj users 20 May 11:09  20 May 11:09  done
3666783 drwxr-x---      - alienzj users 20 May 11:09  20 May 11:08  intermediate_contigs
3666784 .rw-r-----  138Ki alienzj users 20 May 11:09  20 May 11:08  log
3666785 .rw-r-----  1.1Ki alienzj users 20 May 11:08  20 May 11:08  options.json
```

bio/megahit/wrapper.py
fgvieira: yes

bio/megahit/wrapper.py
fgvieira: temp folder can be specified at the rule level by setting `params.tmpdir`.

bio/megahit/wrapper.py
fgvieira: But maybe in the future other users also need other files, like `options.json` or `log`.

bio/megahit/wrapper.py
alienzj: Thanks.

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,73 @@
...
+    output_temp_dir = os.path.join(temp_dir, "temp")
+
+    shell(
+        "megahit "
+        " -t {snakemake.threads} "
+        " -m {memory_requirements} "
+        " -o {output_temp_dir} "
+        " {input_arg} "
+        " {extra} "
+        " > {snakemake.log[0]} 2>&1 "
```
fgvieira, line 61:
```suggestion
        " {log}"
```

test_wrappers.py
```diff
@@ -6398,6 +6398,20 @@ def test_metaspades(run):
     )
 
 
+def test_megahit(run):
+    run(
+        "bio/megahit",
+        [
+            "snakemake",
+            "run_megahit",
```
fgvieira, line 6406:
```suggestion
```

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,73 @@
...
+    if os.path.exists(os.path.join(output_temp_dir, "done")):
+        shell("rm -rf {output_dir}")
+        shutil.move(output_temp_dir, output_dir)
+
+        if (
+            os.path.exists(contigs_file_original)
+            and os.path.exists(options_file)
+            and os.path.exists(log_file)
+        ):
+            shutil.move(contigs_file_original, contigs_file)
```
fgvieira, line 73:
```suggestion
    # Ensure user can name each file according to their need
    output_mapping = {
        "contigs": f"{output_temp_dir}/contigs.fa",
        "json": f"{output_temp_dir}/options.json",
        "log": f"{output_temp_dir}/log",
    }

    for output_key, temp_file in output_mapping.items():
        output_path = snakemake.output.get(output_key)
        if output_path:
            shell("mv --verbose {temp_file:q} {output_path:q} {log}")
```

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,73 @@
...
+# get output_dir and files from output
+output_dir = os.path.split(snakemake.output[0])[0]
+contigs_file = snakemake.output.get("contigs", os.path.join(output_dir, "contigs.fa"))
+contigs_file_original = os.path.join(output_dir, "final.contigs.fa")
+options_file = snakemake.output.get("options", os.path.join(output_dir, "options.json"))
+log_file = snakemake.output.get("log", os.path.join(output_dir, "log"))
+
+# parse params
+extra = snakemake.params.get("extra", "")
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
```
fgvieira, line 21:
```suggestion
log = snakemake.log_fmt_shell(stdout=True, stderr=True, append=True)
```

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,73 @@
...
+import os, tempfile, shutil
+from snakemake.shell import shell
+from snakemake_wrapper_utils.snakemake import get_mem
+
+# get output_dir and files from output
+output_dir = os.path.split(snakemake.output[0])[0]
+contigs_file = snakemake.output.get("contigs", os.path.join(output_dir, "contigs.fa"))
+contigs_file_original = os.path.join(output_dir, "final.contigs.fa")
+options_file = snakemake.output.get("options", os.path.join(output_dir, "options.json"))
+log_file = snakemake.output.get("log", os.path.join(output_dir, "log"))
```
fgvieira, line 17:
```suggestion
```

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,73 @@
...
+
+    shell(
+        "megahit "
+        " -t {snakemake.threads} "
+        " -m {memory_requirements} "
+        " -o {output_temp_dir} "
+        " {input_arg} "
+        " {extra} "
+        " {log}"
+    )
```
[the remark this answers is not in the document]
fgvieira: That parameters is already passed in `extra`

bio/megahit/wrapper.py
```diff
@@ -0,0 +1,73 @@
...
+    input_arg += " --12 {} ".format(reads[2])
+
+# handle additional reads if specified
+if hasattr(snakemake.input, "unpaired"):
+    input_arg += " --read {} ".format(snakemake.input.unpaired)
+elif len(reads) >= 4 and not hasattr(snakemake.input, "r1"):
+    input_arg += " --read {} ".format(reads[3])
+
+
+with tempfile.TemporaryDirectory(dir=os.path.dirname(output_dir)) as temp_dir:
```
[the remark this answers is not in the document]
alienzj: I use with `tempfile.TemporaryDirectory(dir=os.path.dirname(output_dir)) as temp_dir`: because metagenomics workflows often require substantial storage space, and the default `/tmp/{temp_dir}` location may have insufficient capacity.

test_wrappers.py
alienzj: Why remove this line?

bio/megahit/wrapper.py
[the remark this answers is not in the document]
alienzj: If a user specifies `resources.tmpdir`, will `tempfile.TemporaryDirectory()` use it as the system temp directory (like `/tmp`)?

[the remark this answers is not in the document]
alienzj: I see. Thanks.

fgvieira:
Great work @alienzj !

Can you also add the `environment.linux-64.pin.txt` file? You can generate it by running `snakedeploy pin-conda-envs environment.yaml`.

alienzj:
Thanks. Added.

fgvieira (approved):

fgvieira:
Thanks again for contributing to the wrapper! :+1:

PR #4286, opened 2025-06-25
Rasterio clipping of Cloud Optimized GeoTIFF

irm-codebase:
This is a small wrapper to 'clip' GeoTIFF files around the web. Useful for geographic analyses where you are only interested in a small area of the world.

TIFF files can be huge (several GB), so 'clipping' them ensures your workflows are lean.

The wrapper comes with two 'flavours':
- Providing a vector file (e.g., geoparquet, geojson).
- Providing a 'bounding box' (for cases where you have no shapefile).

You can optionally 'extend' the bounding box by a few degrees to ensure all relevant pixels are within it.

### QC


test_wrappers.py
```diff
@@ -6771,4 +6771,18 @@ def test_pygadm_item(run):
...
+            "geo/rasterio/clip-cog",
+            [
+                "snakemake",
+                "--cores",
+                "2",
+                "--use-conda",
+                "-F",
+                f"results/{country}.tiff"
+            ]
+        )
```
fgvieira:
```suggestion
    run(
        "geo/rasterio/clip-cog",
        [
            "snakemake",
            "--cores",
            "2",
            "--use-conda",
            "-F",
            "results/montenegro.tiff",
            "results/switzerland.tiff",
        ]
    )
```

geo/rasterio/clip-cog/test/Snakefile
```diff
@@ -0,0 +1,28 @@
+rule from_file:
+    input:
+        # Geometry file
+        vector="montenegro.parquet",
+    output:
+        # Clipped region
+        path="results/montenegro.tiff",
+    params:
+        buffer_degrees=1,
+        tiff_url="https://zenodo.org/records/14920365/files/minic_edtm_m_960m_s_20000101_20221231_go_epsg.4326_v20241230.tif"
```
fgvieira: Shouldn't this be an input file?
This way the user could also provide a local file, no?

irm-codebase: Can snakemake detect `input` files that are not local?
The thing here is that these files are very large (+1 TB or more for highly resolved datasets).

This wrapper is meant to 'slice' a section of these files and download only that tiny bit... which is why I put URL as a parameter.

irm-codebase: Specifying it as `input` would mean that the file has already been downloaded... which defeats the purpose of the wrapper. COG files must be served via HTTP protocols (https://cogeo.org/). Honestly I am not sure if the protocol would work in the 'local' case.

fgvieira: I see... still, do you think it would be possible to allow both ways? I mean, that the wrapper supports slicing directly from an URL, but also from a local/input file? I think it might be nice to make it as general as possible.

fgvieira:
I am not sure about those suggested changes either, so I'll leave it up to you since you know the program best. 😄

geo/rasterio/clip-cog/test/Snakefile
irm-codebase: Makes sense. I updated the script to take either a local raster (input) or a remote COG (parameter).
I renamed it to clip-geotiff to be more explicit.

I've chosen to not accept other types of raster to keep the complexity low. This will only work with TIFF/TIF files.

fgvieira:
And can you also fix the conflicts?

irm-codebase:
@fgvieira done!
I also removed a couple of checks (it can now accept any URL, not just those ending in tif/tiff) to support different file hosting websites, since some remove file extensions.

geo/rasterio/clip-geotiff/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - vector: Optional. Vector file to use as bbox.
+output:
+  - path: Location of the output .TIFF file.
+params:
+  - bbox: Optional. A bounding box in the shape of [minx, miny, maxx, maxy].
+  - buffer_degrees: Optional. Degrees to extend the bounding box in all directions.
+  - cog_url: URL of the raster file.
+notes: |
+  Important: either the bbox or the vector file should be provided.
+  Giving both will raise an error.
```
fgvieira:
```suggestion
  - raster: Raster GeoTiff (optional).
  - vector: Vector file to use as bbox (optional).
output:
  - path: Output .TIFF file.
params:
  - bbox: Bounding box in the shape of [minx, miny, maxx, maxy]  (optional).
  - buffer_degrees: Degrees to extend the bounding box in all directions  (optional).
  - cog_url: URL of the raster file (optional).
notes: |
  - Either the bbox or the vector file should be provided, but not both.
```

fgvieira (approved):

fgvieira:
Thanks for your contribution! :+1:

irm-codebase:
No problem! Thanks for the super fast review!

PR #4320, opened 2025-07-13
re-make rasterio clip wrapper

irm-codebase:
This is an improvement / pseudo-replacement of #4286 (the introduction of the rasterio/clip-geotiff wrapper).

The aim is to fix 3 issues:
- It opened the full GeoTIFF. These files can be quite large (50 TB+), and in these cases it lead to wrapper crashes.
- It did not save the log!
- It did use `rasterio`'s native `clip` command in the CLI.

This new version uses the CLI to enable GDAL / rasterio to take control for the processing and connection, removing the need of fancy python stuff. I've renamed the wrapper to `rasterio/clip` to match the naming in the CLI and updated the names of the inputs / outputs to fit the name of those people would expect to see in the CLI too.

Other niceties:
- Added defaults that should work in most machines.
- Added logic to ensure the bounding box of the 'cut' always fits the target raster.
- It's possible to clip using local raster, not just vector files.
- new parameters that allow superusers to tweak the GDAL environment variables and the GeoTIFF driver settings (`--co CONFIG=VALUE`)

**Limitations**

Unfortunately there is an issue with GDAL 3.10 that leads to multi-threaded downloads failing. It has been fixed in 3.11, but rasterio currently does not support it. I've protected users against this case and added a few notes for future fixes. I hope that is ok.

### QC


geo/rasterio/clip/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  Three different types of clipping references are possible:
+  * a local vector file
+  * a local raster file
+  * a bounding box (in the CRS specified)
+authors:
+  - Ivan Ruiz Manuel
+input:
+  - raster: target raster file to clip (optional).
+  - like_raster: raster file to use as clipping reference (optional).
+  - like_vector: vector file to use as clipping reference (optional).
```
fgvieira, line 27: What do the `like_` prefix means? Isn't it to specify what part of the TIF should be clipped? If so, why `like`? I find it a bit confusing. :smile:

irm-codebase: I am (unfortunately...) following the naming convention of `rio clip` here, and I must admit I also find it confusing...
Still, it's closer to what a user of the CLI would see/expect, so I kept the odd naming.

In the CLI it works like this:
-  `rio clip input.tiff output.tiff --like other.tiff`
- `rio clip BIGFILE RESULT --like SMALLFILE`

fgvieira: Is there any way to tell form the files whether it is `like_raster` or `like_vector`? Since they are mutually exclusive, I was wondering if it would be possible to have just a single `input.like` and the wrapper would figure out which one it is.

irm-codebase: Not without risking breakages, think... There is a broad range of file extensions that could be used here, and raster files and vectors can both be quite big, meaning incorrect access might brick the process.

Separating them seemed like the safest approach.
The alternative would be having a single 'like' input and a parameter that tells the wrapper what said input is... which adds extra steps on the user side.

fgvieira: No, then might be better as is. Thanks!

fgvieira (approved):

irm-codebase:
Thank you!

PR #4426, opened 2025-08-19
fix: Support both reference.dict and reference.fasta.dict in GRIDSS wrappers

RitzelC:
Reference setup has changed to produce reference.fasta.dict instead of reference.dict.

Here I have changed it to just check the existance of the given dictionary without asuming a fixed ending like .dict.

## Summary by CodeRabbit

* **Refactor**
  * Standardized input naming in assembly workflows for consistency.

* **Bug Fixes**
  * Prevents incorrect derivation of reference dictionary paths by using the provided dictionary input, avoiding mismatches in processing.
  * Updated command invocation to match renamed assembly inputs.

* **Documentation**
  * Clarified missing/dictionary error messages to aid troubleshooting.


bio/gridss/assemble/wrapper.py
```diff
@@ -31,7 +31,7 @@
             )
         )
 
-dictionary = path.splitext(reference)[0] + ".dict"
+dictionary = reference + ".dict"
 if not path.exists(dictionary):
     raise ValueError(
         "{dictionary}.dict missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
```
fgvieira:
```suggestion
        "{dictionary} missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
```

bio/gridss/assemble/wrapper.py
```diff
@@ -45,6 +45,6 @@
     "--threads {snakemake.threads} "  # Threads
     "--workingdir {snakemake.params.workingdir} "  # Working directory
     "--assembly {snakemake.output.assembly} "  # Assembly output
-    "{snakemake.input.bams} "
+    "{snakemake.input.bam} "  # changed to bam for consistancy
```
fgvieira:
```suggestion
    "{snakemake.input.bam} "
```

bio/gridss/preprocess/wrapper.py
```diff
@@ -31,10 +31,10 @@
             )
         )
 
-dictionary = path.splitext(reference)[0] + ".dict"
+# dictionary = path.splitext(reference)[0] + ".dict" --> leads to incorrect paths (reference.dict), which is incompatible with setup reference (reference.fa.dict)
 if not path.exists(dictionary):
     raise ValueError(
-        "{dictionary}.dict missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
+        "Dictionary {dictionary} missing or wrong naming (reference.fa.dict). Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
```
fgvieira:
```suggestion
        "{dictionary} missing or wrong naming (reference.fa.dict). Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
```

bio/gridss/assemble/wrapper.py
```diff
@@ -31,10 +31,10 @@
...
         )
 
-dictionary = path.splitext(reference)[0] + ".dict"
+dictionary = reference + ".dict"
 if not path.exists(dictionary):
     raise ValueError(
-        "{dictionary}.dict missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
+        "{dictionary} missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
             dictionary=dictionary
         )
```
fgvieira:
```suggestion
        f"{dictionary} missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard."
```

bio/gridss/call/wrapper.py
```diff
@@ -31,10 +31,10 @@
...
         )
 
-dictionary = path.splitext(reference)[0] + ".dict"
+#dictionary = path.splitext(reference)[0] + ".dict"
 if not path.exists(dictionary):
     raise ValueError(
-        "{dictionary}.dict missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
+        "{dictionary} missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
             dictionary=dictionary
         )
```
fgvieira:
```suggestion
        f"{dictionary} missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard."
```

bio/gridss/preprocess/wrapper.py
```diff
@@ -31,10 +31,10 @@
...
         )
 
-dictionary = path.splitext(reference)[0] + ".dict"
+# dictionary = path.splitext(reference)[0] + ".dict" --> leads to incorrect paths (reference.dict), which is incompatible with setup reference (reference.fa.dict)
 if not path.exists(dictionary):
     raise ValueError(
-        "{dictionary}.dict missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
+        "{dictionary} missing or wrong naming (reference.fa.dict). Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
             dictionary=dictionary
         )
```
fgvieira:
```suggestion
        f"{dictionary} missing or wrong naming (reference.fa.dict). Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard."
```

bio/gridss/assemble/test/Snakefile
```diff
@@ -33,8 +33,8 @@ reference_index_endings = (".amb",".ann", ".bwt", ".pac", ".sa", ".gridsscache",
 
 rule gridss_assemble:
     input:
-        bams=expand("mapped/{sample}.bam", sample=samples),
-        bais=expand("mapped/{sample}.bam.bai", sample=samples),
+        bam=expand("mapped/{sample}.bam", sample=samples),
+        bai=expand("mapped/{sample}.bam.bai", sample=samples),
         reference="reference/genome.fasta",
         dictionary="reference/genome.dict",
```
fgvieira, line 39: Maybe also rename to `ref` and `dict`, to shorten the names.

bio/gridss/call/wrapper.py
```diff
@@ -31,12 +31,10 @@
             )
         )
 
-dictionary = path.splitext(reference)[0] + ".dict"
+#dictionary = path.splitext(reference)[0] + ".dict"
```
fgvieira, line 34:
```suggestion
```

bio/gridss/assemble/wrapper.py
```diff
@@ -31,12 +31,10 @@
             )
         )
 
-dictionary = path.splitext(reference)[0] + ".dict"
+dictionary = reference + ".dict"
```
fgvieira, line 34: Why not use `snakemake.input.dictionary` like in the other wrappers?

fgvieira, line 34:
```suggestion
```

bio/gridss/preprocess/wrapper.py
```diff
@@ -31,12 +31,10 @@
...
 
-dictionary = path.splitext(reference)[0] + ".dict"
+# dictionary = path.splitext(reference)[0] + ".dict" --> leads to incorrect paths (reference.dict), which is incompatible with setup reference (reference.fa.dict)
 if not path.exists(dictionary):
     raise ValueError(
-        "{dictionary}.dict missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
-            dictionary=dictionary
-        )
+        f"{dictionary} missing or wrong naming (reference.fa.dict). Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard."
```
fgvieira:
```suggestion
```

bio/gridss/assemble/wrapper.py
```diff
@@ -31,12 +31,10 @@
...
 if not path.exists(dictionary):
     raise ValueError(
-        "{dictionary}.dict missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
-            dictionary=dictionary
-        )
+        f"{dictionary} missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard."
```
fgvieira:
```suggestion
```

bio/gridss/call/wrapper.py
```diff
@@ -31,12 +31,10 @@
...
 if not path.exists(dictionary):
     raise ValueError(
-        "{dictionary}.dict missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
-            dictionary=dictionary
-        )
+        f"{dictionary} missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard."
```
fgvieira:
```suggestion
```

bio/gridss/assemble/test/Snakefile
```diff
@@ -33,8 +33,8 @@ reference_index_endings = (".amb",".ann", ".bwt", ".pac", ".sa", ".gridsscache",
...
         indices=multiext("reference/genome.fasta", *reference_index_endings),
```
fgvieira:
```suggestion
        ref="reference/genome.fasta",
        dict="reference/genome.dict",
        idx=multiext("reference/genome.fasta", *reference_index_endings),
```

bio/gridss/assemble/test/Snakefile
fgvieira: Shouldn't this be a temporary directory?

RitzelC: Dear @fgvieira,
sorry for the very late reply and thanks for taking care of formatting etc. 
Yes, this is a temporary directory and it should be cleaned up after an successful run.
I am currently reviewing my code and will have a look at this fix again.
Would you need more input from my side to merge the pull request?

Best regards

bio/gridss/call/wrapper.py
fgvieira:
```suggestion
    "--reference {snakemake.input.reference} "  # Reference
```

bio/gridss/preprocess/wrapper.py
fgvieira:
```suggestion
    "--reference {snakemake.input.reference} "  # Reference
```

bio/gridss/preprocess/wrapper.py
```diff
@@ -31,12 +31,10 @@
...
     )
 
 shell(
```
fgvieira:
```suggestion
extra = snakemake.params.get("extra", "")

shell(
```

bio/gridss/call/wrapper.py
fgvieira:
```suggestion
extra = snakemake.params.get("extra", "")

shell(
```

bio/gridss/assemble/wrapper.py
```diff
@@ -31,12 +31,10 @@
...
     )
```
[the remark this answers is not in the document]
RitzelC: Good suggestion.

bio/gridss/call/wrapper.py
RitzelC: I removed the java option, which can be applied as an "extra" option.

bio/gridss/assemble/wrapper.py
RitzelC: I removed the dictionary check competely and added the extension to the check of file endings above.
In theory, this check only has to be applied in the preprocess step, because downstream steps use the output of the preprocessing and the same references.

bio/gridss/call/wrapper.py
RitzelC: I removed the dictionary check competely and added the extension to the check of file endings above.

bio/gridss/call/wrapper.py
RitzelC: Applied the change

bio/gridss/preprocess/wrapper.py
RitzelC: I removed the dictionary check competely and added the extension to the check of file endings above.

RitzelC (commented):
Dear @fgvieira,
I applied your suggested changes and tried to simplify the dictionary checking.
For my local test, it was working fine.

I realized that I removed some the library parameter from the Snakefile. This could be a breaking change.
How would you proceed? Keep the unused parameter anyways?

bio/gridss/assemble/test/Snakefile
RitzelC: The problem is that the working directory is shared between all rules and cannot be marked as temporary directory. I guess I would have to rewrite the structure of the wrapper to combine all rules into one pipeline.

RitzelC (commented):
The problem with the dictionary should be fixed with this issue, but I did not yet the working directory as temporary directory.

bio/gridss/assemble/test/Snakefile
fgvieira: What do you mean by "shared between all rules"? In the wrapper test there is only one rule.

And according to the [docs](https://github.com/PapenfussLab/GRIDSS#working_dir), `working_dir` is used to store intermediate results, so it should not be needed after the command finished successfully, no? Also, it sounds like this option is not mandatory; would the wrapper work without it?

fgvieira:
It seems there is a [tmp_dir](https://github.com/PapenfussLab/GRIDSS#tmp_dir) option. can you also add it and set it to `resources.tmpdir`?

bio/gridss/preprocess/wrapper.py
```diff
@@ -7,43 +7,38 @@
 
 from snakemake.shell import shell
 from os import path
+from shlex import quote
```
fgvieira:
```suggestion
import tempfile
from snakemake.shell import shell
from os import path
from shlex import quote
```

bio/gridss/preprocess/wrapper.py
```diff
@@ -7,43 +7,38 @@
...
 
 # Creating log
 log = snakemake.log_fmt_shell(stdout=True, stderr=True)
 
 # Placeholder for optional parameters
 extra = snakemake.params.get("extra", "")
+tmpdir = snakemake.resources.get("tmpdir")
+tmpdir_arg = f" TMP_DIR={quote(str(tmpdir))}" if tmpdir else ""
```
fgvieira:
```suggestion
```

bio/gridss/preprocess/wrapper.py
```diff
@@ -7,43 +7,38 @@
...
 shell(
     "(gridss -s preprocess "  # Tool
-    "--reference {reference} "  # Reference
+    "--reference {snakemake.input.reference} "  # Reference
     "--threads {snakemake.threads} "
     "--workingdir {snakemake.params.workingdir} "
     "{snakemake.input.bam} "
-    "{extra}) {log}"
+    "{extra}"
+    "{tmpdir_arg}) {log}"
```
fgvieira:
```suggestion
    "TMPDIR={tmpdir}
) {log}"
```

bio/gridss/preprocess/wrapper.py
```diff
@@ -7,43 +7,38 @@
...
 
-dictionary = path.splitext(reference)[0] + ".dict"
-if not path.exists(dictionary):
-    raise ValueError(
-        "{dictionary}.dict missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
-            dictionary=dictionary
-        )
-    )
-
 shell(
```
fgvieira:
```suggestion
with tempfile.TemporaryDirectory() as tmpdir:
    shell(
```

bio/gridss/call/meta.yaml
```diff
@@ -3,6 +3,7 @@ description: >
     GRIDSS is a module software suite containing tools useful for the detection of genomic rearrangements.
     It includes a genome-wide break-end assembler, as well as a structural variation caller for Illumina sequencing data.
     ``call`` performs variant calling.
+    ``TMP_DIR`` is automatically set by ``resources.tmpdir`` when provided.
     Documentation at: https://github.com/PapenfussLab/gridss
```
fgvieira:
```suggestion
url: https://github.com/PapenfussLab/GRIDSS
description: >
   GRIDSS is a module software suite containing tools useful for the detection of genomic rearrangements.
   It includes a genome-wide break-end assembler, as well as a structural variation caller for Illumina sequencing data.
   ``call`` performs variant calling.
```

bio/gridss/assemble/meta.yaml
```diff
@@ -3,6 +3,7 @@ description: >
     GRIDSS is a module software suite containing tools useful for the detection of genomic rearrangements.
     It includes a genome-wide break-end assembler, as well as a structural variation caller for Illumina sequencing data.
     ``assemble`` performs GRIDSS breakend assembly.
+    ``TMP_DIR`` is automatically set by ``resources.tmpdir`` when provided.
     Documentation at: https://github.com/PapenfussLab/gridss
```
fgvieira:
```suggestion
url: https://github.com/PapenfussLab/GRIDSS
description: >
   GRIDSS is a module software suite containing tools useful for the detection of genomic rearrangements.
   It includes a genome-wide break-end assembler, as well as a structural variation caller for Illumina sequencing data.
   ``assemble`` performs GRIDSS breakend assembly.
```

bio/gridss/preprocess/meta.yaml
```diff
@@ -3,6 +3,7 @@ description: >
     GRIDSS is a module software suite containing tools useful for the detection of genomic rearrangements.
     It includes a genome-wide break-end assembler, as well as a structural variation caller for Illumina sequencing data.
     ``preprocess`` pre-processes input BAM files (can be run per file).
+    ``TMP_DIR`` is automatically set by ``resources.tmpdir`` when provided.
 url: https://github.com/PapenfussLab/gridss
```
fgvieira:
```suggestion
url: https://github.com/PapenfussLab/GRIDSS
description: >
   GRIDSS is a module software suite containing tools useful for the detection of genomic rearrangements.
   It includes a genome-wide break-end assembler, as well as a structural variation caller for Illumina sequencing data.
   ``preprocess`` pre-processes input BAM files (can be run per file).
```

fgvieira:
Can you change the wrappers to use the same TMP_DIR logic as in preprocess?

RitzelC:
I applied the same TMP_DIR logic, but unfortunately it was not that easy to pass the parameter to gridss. So I had to pass it via the --picardoptions parameter.
For my local environment it seems to work fine :)
Thanks for your help. Do I have to trigger the rerun of the tests somehow?

RitzelC:
But the way the temporary directory is created, the directory cannot be set through a snakemake parameter.

fgvieira:
> But the way the temporary directory is created, the directory cannot be set through a snakemake parameter.

`resources.tmpdir` sets `TMPDIR`, and that is the default folder for `tempfile`.

bio/gridss/assemble/wrapper.py
```diff
@@ -23,28 +24,23 @@
...
     raise ValueError("Please set input.reference to provide reference genome.")
 
-for ending in (".amb", ".ann", ".bwt", ".pac", ".sa"):
+for ending in (".amb", ".ann", ".bwt", ".pac", ".sa", ".dict"):
     if not path.exists("{}{}".format(reference, ending)):
         raise ValueError(
             "{reference}{ending} missing. Please make sure the reference was properly indexed by bwa.".format(
                 reference=reference, ending=ending
             )
         )
```
fgvieira, line 34: Can't we delete all this (same for other wrappers)?
```suggestion
```

bio/gridss/assemble/wrapper.py
```diff
@@ -23,28 +24,23 @@
...
-dictionary = path.splitext(reference)[0] + ".dict"
-if not path.exists(dictionary):
-    raise ValueError(
-        "{dictionary}.dict missing. Please make sure the reference dictionary was properly created. This can be accomplished for example by CreateSequenceDictionary.jar from Picard".format(
-            dictionary=dictionary
-        )
+with tempfile.TemporaryDirectory() as tmpdir:
+    shell(
+        "(gridss -s assemble "  # Tool
+        "--reference {reference} "  # Reference
```
fgvieira:
```suggestion
        "--reference {snakemake.input.reference} "  # Reference
```

fgvieira:
@RitzelC it looks quite good, but I am only a bit concerned about this `working_dir`. I see it as temp, so not sure why you need to specify it.

@dlaehnemann do you have experience with GRIDSS? What do you think about the use of a `params.working_dir`?

bio/gridss/assemble/test/Snakefile
```diff
@@ -1,3 +1,5 @@
+import os
+
 WORKING_DIR = "working_dir"
```
dlaehnemann, line 3: I know this was already there, previously, but this should just be part of the path in the rule. That way, the rule becomes more readable and easier to copy-paste.

bio/gridss/assemble/test/Snakefile
```diff
@@ -27,28 +29,45 @@ assembly_endings = (
...
         reference="reference/genome.fasta",
-        dictionary="reference/genome.dict",
-        indices=multiext("reference/genome.fasta", *reference_index_endings),
-        preprocess=expand("{working_dir}/{sample}.bam.gridss.working/{sample}.bam{ending}", working_dir=[WORKING_DIR], sample=samples, ending=preprocess_endings)
+        idx=multiext("reference/genome.fasta", *reference_index_endings),
+        preprocess=expand(
+            "{working_dir}/{sample}.bam.gridss.working/{sample}.bam{ending}",
+            working_dir=[WORKING_DIR],
+            sample=samples,
+            ending=preprocess_endings,
```
dlaehnemann, line 55: I would also use a `multiext()` statement here. Or is this somehow clashing with the other existing `multiext()` above?

bio/gridss/assemble/test/Snakefile
```diff
@@ -27,28 +29,45 @@ assembly_endings = (
...
     input:
-        bams=expand("mapped/{sample}.bam", sample=samples),
-        bais=expand("mapped/{sample}.bam.bai", sample=samples),
+        bam=expand("mapped/{sample}.bam", sample=samples),
+        bai=expand("mapped/{sample}.bam.bai", sample=samples),
         reference="reference/genome.fasta",
-        dictionary="reference/genome.dict",
-        indices=multiext("reference/genome.fasta", *reference_index_endings),
-        preprocess=expand("{working_dir}/{sample}.bam.gridss.working/{sample}.bam{ending}", working_dir=[WORKING_DIR], sample=samples, ending=preprocess_endings)
+        idx=multiext("reference/genome.fasta", *reference_index_endings),
```
dlaehnemann, line 50: I would list all of the endings / suffixes right here, so readers of the rule don't have to cross-reference with a different place in the code to see what is actually required.

bio/gridss/assemble/test/Snakefile
```diff
@@ -27,28 +29,45 @@ assembly_endings = (
...
     params:
         extra="--jvmheap 1g",
-        workingdir=WORKING_DIR
-    log:
-        "log/gridss/assemble/group.log"
-    threads:
-        100
+        workingdir=lambda wildcards, output: os.path.dirname(
+            os.path.dirname(output.assembly_others[0])
+        ),
```
dlaehnemann, line 71: I think the new `subpath()` helper function can make this more readable:
https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#sub-path-access

bio/gridss/assemble/test/Snakefile
dlaehnemann: If users want to make part of the path configurable upon module-import of a workflow, they can use `pathvar`s:
https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#path-variables

bio/gridss/assemble/test/Snakefile
dlaehnemann: Also, no need to `import os` then...

bio/gridss/preprocess/test/Snakefile
```diff
@@ -1,20 +1,37 @@
-WORKING_DIR="working_dir"
+import os
+
+WORKING_DIR = "working_dir"
```
dlaehnemann, line 3: Here, the `WORKING_DIR` is not even used, as all the occurrences below are snakemake wildcards, that don't use this global variable. So we should just get rid of this global variable in all of the wrappers.

dlaehnemann (commented):
I think getting the workflow_dir into the wrapper via a function call in params is OK. It's a hack than we use in many cases, where a tool just wants a folder name, but snakemake wants to track individual files. As mentioned in the respective comment, I would just opt for the recently added `subpath()` helper function, to make this as readable and straightforward as possible:
https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#sub-path-access

RitzelC:
> @RitzelC it looks quite good, but I am only a bit concerned about this `working_dir`. I see it as temp, so not sure why you need to specify it.
> 
> @dlaehnemann do you have experience with GRIDSS? What do you think about the use of a `params.working_dir`?

Yes, this is a temporary directory, but was not that easy to remove, because other steps of the workflow depend on it. I have to look into the subpath feature.

fgvieira:
If it's a temp folder, then we can use `tempfile` and it gets cleaned up when the run is over.
You also say that it is needed for other steps; if so, then it should be an `input` or `output`, no?

PR #4474, opened 2025-09-17
fix: Add fallback to FTP if HTTPS cache download fails for vep-cache

fxwiegand:
This pull request updates the cache download logic in the `bio/vep/cache/wrapper.py` script to improve reliability and compatibility with different network environments. The main change is to attempt downloading the cache tarball via HTTPS first, and if that fails, fall back to FTP.

## Summary by CodeRabbit

* **New Features**
  * Allow specifying download source as a host or full URL; defaults choose HTTPS for host input and FTP as fallback.
  * Add an "indexed" cache option that selects the indexed cache path and adjusts conversion behavior.

* **Bug Fixes**
  * More robust download: try HTTPS first, verify archive integrity, then fall back to FTP on failure.

* **Compatibility**
  * Tarball naming and downstream installation behavior remain unchanged.

### DIFF
--- a/bio/vep/cache/wrapper.py
+++ b/bio/vep/cache/wrapper.py
@@ -3,6 +3,7 @@ __copyright__ = "Copyright 2023, Johannes Köster"
 __email__ = "johannes.koester@uni-due.de"
 __license__ = "MIT"
 
+import subprocess as sp
 import tempfile
 from pathlib import Path
 from snakemake.shell import shell
--- a/bio/vep/cache/wrapper.py
+++ b/bio/vep/cache/wrapper.py
@@ -21,8 +21,16 @@ except ValueError:
 with tempfile.TemporaryDirectory() as tmpdir:
     # We download the cache tarball manually because vep_install does not consider proxy settings (in contrast to curl).
     # See https://github.com/bcbio/bcbio-nextgen/issues/1080
-    url_https = f"https://{snakemake.params.get('url', 'ftp.ensembl.org/pub')}/release-{release}/variation/{vep_dir}/{cache_tarball}"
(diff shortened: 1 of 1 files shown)


bio/vep/cache/wrapper.py
```diff
@@ -21,7 +21,8 @@
 with tempfile.TemporaryDirectory() as tmpdir:
     # We download the cache tarball manually because vep_install does not consider proxy settings (in contrast to curl).
     # See https://github.com/bcbio/bcbio-nextgen/issues/1080
-    cache_url = snakemake.params.get("url", "ftp://ftp.ensembl.org/pub")
+    url_https = f"https://{snakemake.params.get('url', 'ftp.ensembl.org/pub')}/release-{release}/variation/{vep_dir}/{cache_tarball}"
```
johanneskoester: Can we avoid the breaking change here, and still allow users to specify a full url with scheme?

johanneskoester (changes requested):

fgvieira:
Yes, that was just to show the general idea, but one could use `re.sub`:
```
try:
    shell(
            "curl -L {cache_url}/release-{release}/variation/{vep_dir}/{cache_tarball} -o {tmpdir}/{cache_tarball} {log}"
         )
except sp.CalledProcessError:
    cache_url = re.sub(r'^http', 'ftp', cache_url)
    shell(
            "curl -L {cache_url}/release-{release}/variation/{vep_dir}/{cache_tarball} -o {tmpdir}/{cache_tarball} {log}"
         )
```

if user, provides `http` and fails, try `ftp`.
If user provides `ftp`, try twice.

PR #4638, opened 2025-10-09
feat: allow setting launchDir in params

tedil:
Because nextflow does not allow specifying the `launchDir`, but stores its `.nextflow` folder in that directory, the wrapper now allows changing into the directory specified in the param `launch_dir` before invoking nextflow.
Not sure if this is the way to go / a robust way of doing so, though. Comments welcome.

## Summary by CodeRabbit

* **New Features**
  * Optional launch_dir support: runs can validate and switch to a specified working directory before execution to improve reproducibility.

* **Refactor**
  * Parameter forwarding adjusted to exclude internal keys (including launch_dir) from downstream command inputs for cleaner behavior.

* **Documentation**
  * Added configuration documentation describing the launch_dir option and its intended behavior.


fgvieira:
Is this `launch_dir` a temporary folder? Or are the results stored there?
Right now it is being set both in `output` and `params`, but could we just use `resources.tmpdir`?

PR #4656, opened 2025-10-10
fix: remove unecessary deps

fgvieira:
### QC

## Summary by CodeRabbit

* **Chores**
  * Removed explicit Python version pins from multiple GATK3 environments.
  * Removed samtools from Paladin index and prepare environments.
  * Removed numpy, pandas and polars from the datavzrd environment.
  * Updated Conda tooling metadata to 25.9.0 and reduced explicit third‑party package pins in several manifests.


utils/datavzrd/environment.yaml
```diff
@@ -4,6 +4,3 @@ channels:
 dependencies:
   - datavzrd =2.58.8
   - yte =1.8.1
-  - numpy
-  - pandas
-  - polars
```
johanneskoester, line 9: @fxwiegand I think these were added intentionally weren't they?

fgvieira: The tests do not fail and, if they are needed by `datavzrd`, shouldn't they be added directly on the conda recipe?

fxwiegand: They were added because of this template rendering issue with YTE that complained about something with numpy I think. @johanneskoester is this solved now?

johanneskoester: I think they have to be present whenever the datavzrd config that is passed to the wrapper is a YTE template that contains numpy, pandas, or polars elements. Before we remove them, we should definitely extend the test case in this repo to also do that, maybe in a followup PR instead of changing it here?

fgvieira: Ok, so I have reverted the changes on `datavzrd` so we can merge it.

fxwiegand: @fgvieira Thank you!

fxwiegand (approved):
I can only give the feedback that there shouldn't be an issue regarding the datavzrd wrapper. I would be happy to help out with testing in a separate PR where the deps in question would be removed.

Issue #4680, opened 2025-10-30
cnvkit batch wrapper fails if input file name contains extra delimiters


andrewgull:
**Snakemake version**
v9.6.2

wrapper version v5.2.1

**Describe the bug**

The CNVkit batch Snakemake wrapper currently fails to correctly locate output files when the input BAM filename contains extra delimiters (e.g., periods) or complex tags before the final extension. This occurs because CNVkit determines its output base name by stripping the extension from the input, but the wrapper's logic does not account for the possibility of a more complex base name.

**Logs**

[cnvkit_batch_wrapper_error.txt](https://github.com/user-attachments/files/23234986/cnvkit_batch_wrapper_error.txt)

**Minimal example**
```
# the wrapper will throw the error with this input file names
rule cnvkit_batch:
    input:
        bam=["mapped/a.haplotag.bam"],
        bai=["mapped/a.haplotag.bam.bai"],
        reference="ref/reference.cnn",
    output:
        antitarget_coverage="cnvkit/a.antitargetcoverage.cnn",
        bins="cnvkit/a.bintest.cns",
        regions="cnvkit/a.cnr",
        segments="cnvkit/a.cns",
        segments_called="cnvkit/a.call.cns",
        target_coverage="cnvkit/a.targetcoverage.cnn",
    log:
        "cnvkit/a.antitargetcoverage.cnn.log",
    params:
        method='hybrid',  # optional
        extra="",  # optional
    resources:
        mem_mb=1024,
    wrapper:
        "master/bio/cnvkit/batch"
```

**Additional context**
As far as I know, this bug doesn't affect `rule cnvkit_batch_build_reference`

PR #4682, opened 2025-10-30
fix: resolve 'Couldn't locate file' error with input BAM with extra delimiters

andrewgull:
Fixes issue #4680 

Updates the output file copying logic to correctly handle BAM files whose names contain intermediate words/tags (like `{sample}.haplotagged.bam`).

### QC

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logging: capture stdout alongside stderr and append to logs; safer skipping of ambiguous temporary outputs to avoid incorrect mappings or overwrites.

* **Refactor**
  * Unified discovery and movement of batch run outputs via a file-mapping and move-files workflow; removed legacy per-output copy logic while preserving special-case handling for reference outputs.

* **Chores**
  * Updated environment to add an auxiliary workflow utility and pinned packages to support the new mapping workflow.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


bio/cnvkit/batch/wrapper.py
```diff
@@ -65,15 +63,42 @@
...
+
+    for file in temp_files:
+        for suffix, attr in file_map.items():
+            if suffix in file:
+                # Skip ambiguous matches
+                if attr == "segments" and any(
+                    x in file for x in ["call.cns", "bintest.cns"]
+                ):
+                    continue
+                destinations.append(getattr(snakemake.output, attr, None))
```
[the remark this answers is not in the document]
andrewgull: I think this is not valid in Snakemake context, because if one of the input files is not provided `cnvkit batch` will throw an error before this check can be triggered

bio/cnvkit/batch/wrapper.py
```diff
@@ -65,15 +61,33 @@
...
+    # mapping of file suffixes to snakemake.output attributes
+    file_map = {
+        "antitargetcoverage.cnn": "antitarget_coverage",
+        "bintest.cns": "bins",
+        ".cnr": "regions",
+        "call.cns": "segments_called",
+        "targetcoverage.cnn": "target_coverage",
+        ".cns": "segments",
+        "reference.cnn": "reference",
+    }
```
fgvieira, line 76: How does cnvkit defined the output file names in the output dir? Can you define a prefix?

It would be cleaner and less error-prone then matching the suffix.

bio/cnvkit/batch/wrapper.py
```diff
@@ -65,15 +61,33 @@
...
+
+    # find matches btw generated files and snakemake output
+    for file in temp_files:
+        for suffix, attr in file_map.items():
+            if file.endswith(suffix):
+                # Skip ambiguous matches
+                if attr == "segments" and any(
+                    x in file for x in ["call.cns", "bintest.cns"]
+                ):
+                    continue
```
fgvieira, line 90: Why this extra testing?

bio/cnvkit/batch/wrapper.py
andrewgull: unfortunately, there is no way to define prefix (to the best of my knowledge). What `cnvkit batch` does is that it takes the input BAM filename, removes `.bam`, and uses this string as a base to construct the output filenames, hence this entire PR.

bio/cnvkit/batch/wrapper.py
andrewgull: it is done to distinguish a simple `.cns` output file from `.bintest.cns` and `.call.cns` output files

bio/cnvkit/batch/wrapper.py
```diff
@@ -65,15 +61,33 @@
...
+                mapping[attr] = join(tmpdirname, file)
+                break  # stop after first match
```
fgvieira: What if an output was not specified in the rule output files? As far as I can see, it would still be added to the mapping and it would fail on move_files, no?

Maybe something like:

```suggestion
    for suffix, attr in file_map.items():
        if not snakemake.output.get(attr):
            continue
        for file in temp_files:
            if file.endswith(suffix):
                # Skip ambiguous matches
                if attr == "segments" and any(
                    x in file for x in ["call.cns", "bintest.cns"]
                ):
                    continue
                mapping[attr] = join(tmpdirname, file)
                break  # stop after first match
```

PR #4779, opened 2025-12-05
feat: go-yq

tdayris:
This PR brings [`go-yq`](https://github.com/mikefarah/yq) to the list of available wrappers. I called this wrapper `go-yq` and not simply `yq` since it refers to the `conda-forge/go-yq` package which is not the same as `conda-forge/yq`. Yet, both have the same name though CLI, and a shared history.

### QC

* [X] I confirm that I have followed the [documentation for contributing to `snakemake-wrappers`](https://snakemake-wrappers.readthedocs.io/en/stable/contributing.html).

* [X] `test.py` was updated to call any added or updated example rules in a `Snakefile`
* [X] `input:` and `output:` file paths in the rules can be chosen arbitrarily
* [X] wherever possible, command line arguments are inferred and set automatically (e.g. based on file extensions in `input:` or `output:`)
* [X] temporary files are either written to a unique hidden folder in the working directory, or (better) stored where the Python function `tempfile.gettempdir()` points to 
* [X] the `meta.yaml` contains a link to the documentation of the respective tool or command under `url:`
* [X] conda environments use a minimal amount of channels and packages, in recommended ordering

## Summary by CodeRabbit

* **New Features**
  * Added a go-yq wrapper for YAML/CSV processing: concatenation, format conversion, content updates, expression evaluation, splitting, and automatic format detection/validation.

* **Tests**
  * Added end-to-end tests exercising concat, update, evaluate, split, and format-conversion scenarios.

* **Chores**
  * Added reproducible environment specifications and package metadata for the new wrapper.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


utils/go-yq/wrapper.py
```diff
@@ -0,0 +1,77 @@
...
+        raise ValueError(
+            "File splitting is not automatically handled. Please provide "
+            "adequate '-s/--split-exp' in 'params.extra'."
+        )
+    log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+    outfile = ""
+else:
+    log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+    outfile = f"> {snakemake.output[0]}"
+
```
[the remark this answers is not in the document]
tdayris: This should never happen, but a specific error message is now given.

utils/go-yq/wrapper.py
```diff
@@ -0,0 +1,77 @@
...
+        # Trying to detect output format from file extension
+        # since yq defaults to yaml outout and writes information
+        # in standard output.
+        fmt = accepted_formats.get(formats[0])
+        if fmt:
+            return f" {long_option} {fmt}"
+
+    return ""
+
```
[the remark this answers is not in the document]
tdayris: This case is now taken into account.

utils/go-yq/wrapper.py
```diff
@@ -0,0 +1,77 @@
...
+# Enhance extra parameters if needed
+extra = snakemake.params.get("extra", "")
+extra += detect_fmt(snakemake.input, "--input-format", "-p", extra)
+extra += detect_fmt(snakemake.output, "--output-format", "-o", extra)
+
+subcommand = snakemake.params.get("subcommand", "")
+# yq/jq expression should be quoted
+expression = shlex.quote(snakemake.params.get("expression", ""))
+
+# Handle the case use creates a file from command line
```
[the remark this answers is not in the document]
tdayris: Fixed.

test_wrappers.py
```diff
@@ -7165,3 +7165,20 @@ def test_orthanq(run):
...
+        [
+            "snakemake",
+            "--cores",
+            "1",
+            "--use-conda",
+            "-F",
+            "concat.yaml",
+            "updated.yaml",
+            "evaluated.yaml",
+            "table.json",
```
fgvieira, line 7182:
```suggestion
            "foo_bar.yaml",
            "table.json",
```

tdayris: Thanks!

utils/go-yq/wrapper.py
```diff
@@ -0,0 +1,79 @@
...
+        "tsv": "tsv",
+        "toml": "toml",
+        "sh": "shell",
+        "lua": "lua",
+        "ini": "ini",
+        "xml": "xml",
+    }
+
+    # Ensure user did not provide any format information
+    if not (is_arg(long_option, extra) or is_arg(short_option, extra)):
```
fgvieira, line 31: Shouldn't this check should give an error, no? The user should never specify format information, right? As it is right now, I think the function just returns `""`, no?

utils/go-yq/wrapper.py
```diff
@@ -0,0 +1,79 @@
...
+            "File splitting is not automatically handled. Please provide "
+            "adequate '-s/--split-exp' in 'params.extra'."
+        )
+    log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+    outfile = ""
+elif len(snakemake.output) == 1:
+    log = snakemake.log_fmt_shell(stdout=False, stderr=True)
+    outfile = f"> {snakemake.output[0]}"
+else:
+    raise ValueError("Inplace modifications are not handled by this wrapper.")
```
fgvieira, line 77: So this is triggered when there are no output files. Does the prog support this? Would it ever happen in `snakemake`?

tdayris: Hi,

Yes, `yq` supports inplace modifications with `-i`. However, I have not found how to detect it automatically. Is there a way to identify temporary flag files in snakemake (`temp(touch("..."))`) ?

utils/go-yq/wrapper.py
tdayris: Hi,

Yes, they would do it sometimes. Some format have no official extension name (e.g. `base64` is often saved in a `txt` file), and cannot be automatically inferred. Some formats are compatible with a too wide range of formats (e.g. shell formatting being compatible with bash, zsh, kzsh, xonsh, etc.).

utils/go-yq/wrapper.py
fgvieira: Not sure if it will work, but `snakemake` has a `is_temp`:
https://github.com/snakemake/snakemake/blob/555ab6a31def48feec9ce584d70022f9f683bf29/src/snakemake/io/__init__.py#L434-L435

You can give it a try.

tdayris: Doesn't it match all files marked with `temp()`, even if they are not `touch()` ?

fgvieira: I think you can do a specific file:
```
is_temp(snakemake.output[0])
```
or even maybe:
```
snakemake.output[0].is_temp()
```
but not sure.

fgvieira: On the other hand, since `snakemake` relies strongly on `input`/`output` files, maybe it actually makes sense to now allow for inplace modifications in the wrapper.

fgvieira (approved):

utils/go-yq/wrapper.py
tdayris: I have found how to do it.

Case 1: there is a `touch()`-ed output file that can be detected using `snakemake.io.is_flagged(snakemake.output[0], "touch")`. In that case I can add the `--inplace` argument.

Case 2: there is no output file, and I can add the `--inplace` argument.

However, you're right saying this is not a good practice. So... Should I add the changes or leave it in its actual state ?

fgvieira: Maybe lets leave it as-is for now and we'll see if there is need for it in the future.
thanks again!

PR #5190, opened 2026-03-27
test: Clean up tests

fgvieira:
Clean-up tests:
- merge tests using the same wrapper to speed up testing (conda env is only created once).
- make `--use-conda` and `-F` defaults in `run()`
- add `cores` default in `run()`
- general clean-up and reorganization

### QC


dlaehnemann (approved):
Very nice cleanup idea, looking great!

One suggestion, but maybe you already considered this:

As you did with `--use-conda` and `-F`, you could make `--cores 1` the default for `run()`. That way, you would only have to change this via a `run(cores=2)` argument whenever this value is different from 1. That would remove some more repetition....

PR #5230, opened 2026-04-16
fix: Change bbtools logging level

fgvieira:
### QC


dlaehnemann (approved):
This looks straightforward to me, although I don't fully understand the implications. I guess you just want to reduce the amount of logging output that is generated by default?

fgvieira:
Since bbtools log already prints the full command line, all this extra logging is not really necessary (I think). But I leave it there and change the logging level, in case someone in the future comes up with a case where it is nice to have the extra logging.

PR #5265, opened 2026-04-24
feat: add flye wrapper

arog-bioinfo:
This PR adds a wrapper for [Flye](https://github.com/mikolmogorov/Flye). Flye is a de novo assembler for single-molecule sequencing reads, such as those produced by PacBio and Oxford Nanopore Technologies.
### QC

## Summary by CodeRabbit

* **New Features**
  * Integrated Flye long-read de novo assembly tool for genomic sequence assembly.
  * Supports configurable parameters including sequencing type selection and intermediate file retention.
  * Produces assembled sequence, repeat graph visualization, and assembly metadata outputs.


bio/flye/wrapper.py
```diff
@@ -0,0 +1,113 @@
...
+    shell(
+        "flye"
+        " --threads {snakemake.threads}"
+        " {extra}"
+        " {input_arg}"
+        " -o {out_dir}"
+        " {log}"
+    )
+except Exception:
+    raise
```
fgvieira, line 100: No need for try, `snakemake` will take care of it.
```suggestion
shell(
    "flye"
    " --threads {snakemake.threads}"
    " {extra}"
    " {input_arg}"
    " -o {out_dir}"
    " {log}"
)
```

bio/flye/wrapper.py
```diff
@@ -0,0 +1,113 @@
...
+    for dirname in [
+        "00-assembly",
+        "10-consensus",
+        "20-repeat",
+        "30-contigger",
+        "40-polishing",
+    ]:
+        path = os.path.join(out_dir, dirname)
+        if os.path.isdir(path):
+            shutil.rmtree(path)
```
fgvieira, line 113: Maybe better to use a temp folder with `tempfile` to run all analyses, and just copy the requested files to their final destination. For example: like in:

https://github.com/snakemake/snakemake-wrappers/blob/e08e85fbb62bf3357d1b87c02bbc4351a267d4dc/bio/falco/wrapper.py#L17-L35

bio/flye/test/Snakefile
```diff
@@ -0,0 +1,23 @@
...
+rule run_flye:
+    input:
+        reads="test_reads/Loman_E.coli_MAP006-1_2D_50x.first500.fasta.gz",
+    output:
+        assembly="{dir_path}/assembly.fasta",
+        path=directory("{dir_path}/00-assembly"),
+    benchmark:
+        "logs/benchmarks/{dir_path}/flye.txt"
+    params:
+        out_dir=lambda wildcards: wildcards.dir_path,
```
fgvieira, line 13: this might be unnecessary, if we use `tempfile`.

bio/flye/test/Snakefile
```diff
@@ -0,0 +1,23 @@
...
+        seq_type="nano-raw",
+        keep_intermediates="yes",
```
fgvieira, line 15: Might be also be unnecessary, if we use `tempfile`.

bio/flye/test/Snakefile
```diff
@@ -0,0 +1,23 @@
+container: "docker://continuumio/miniconda3:4.4.10"
+
+
+rule run_flye:
+    input:
+        reads="test_reads/Loman_E.coli_MAP006-1_2D_50x.first500.fasta.gz",
+    output:
+        assembly="{dir_path}/assembly.fasta",
+        path=directory("{dir_path}/00-assembly"),
```
fgvieira, line 9: read the `tempfile` suggestion but, if it makes sense, here there should only be file names.

bio/flye/test/Snakefile
```diff
@@ -0,0 +1,23 @@
+container: "docker://continuumio/miniconda3:4.4.10"
+
+
```
fgvieira, line 3: Is this needed?
```suggestion
```

bio/flye/wrapper.py
arog-bioinfo: I thought about using a temporary folder (e.g. with tempfile), but I wanted to make use of the `--resume`parameter in Flye, which allows continuing a incomplete run from intermediate files (I think SPAdes has a similiar option in [it's wrapper](https://github.com/snakemake/snakemake-wrappers/blob/master/bio/spades/metaspades/wrapper.py)).

From my understanding, if I use a temporary directory and the process doesn’t finish, all the intermediate files would be deleted when the temp folder is cleaned up, which would prevent resuming. Is this correct?

Is there some workaround to still use a temp directory while keeping the ability to resume?

fgvieira: Under which circumstances would you'd like to resume a previous job? If the job failed, how can you be sure that the files are not corrupted?
If a job fails, snakemake will delete all existing output files in case they are corrupted.

Personally, I don't think resuming jobs is a good idea.
@dlaehnemann and @johanneskoester  what do you think?

arog-bioinfo: One situation I had in mind is external interruptions (e.g. power outage or system shutdown), since genome assembly jobs can take quite a long time.

Regarding corruption, my understanding is that Flye’s `--resume` restarts from the last completed stage and recomputes incomplete ones, rather than reusing partial outputs. But I’m still learning these tools, so I wouldn’t fully stand by this if there are edge cases I’m missing.

PR #5282, opened 2026-05-02
feat: add libarchive extraction and compression wrappers

irm-codebase:
This PR adds rather simple wrappers to extract / compress files from popular archiving formats (7zip, zip, tar, etc).
It relies on the python interface to the `libarchive` C library.

Reasoning: the conda-forge version of `zip` is not supported for windows machines, and `libarchive` is.

Also, it should save a bunch of boilerplate code / chores related to (un)zipping stuff.

### QC

## Summary by CodeRabbit

* **New Features**
  * Added archive extraction and compression utilities supporting 7z, zip, tar.gz with configurable internal-to-output path mapping and robust validation/error cleanup.

* **Tests**
  * Added end-to-end tests covering single- and multi-file extraction and compression scenarios with expected output assertions.

* **Chores**
  * Added reproducible, pinned Conda environment specifications for the extraction and compression tools.

[Review Change Stack](https://app.coderabbit.ai/change-stack/snakemake/snakemake-wrappers/pull/5282)


fgvieira:
Instead of `libarchive/extract`, maybe just `libarchive`? Or are you planning more sub-wrappers?

utils/libarchive/extract/environment.yaml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - conda-forge
+  - nodefaults
+dependencies:
+  - libarchive =3.8.7
```
fgvieira: This is already a requirement of `python-libarchive-c`, no?
```suggestion
```

irm-codebase: Ah, fair. I'll remove it.

irm-codebase:
@fgvieira I'm not planning more sub-wrappers for the time being, but:

Maybe in the future having a `libarchive/compress` wrapper could be useful... but that's for later.
In general, I think it's better to leave the door open to submodules, if the wrapped tool(s) has more use cases, and this one does.

fgvieira:
But, just like with `bgzip`, can't the compress/extract both use the same wrapper? 
One can check the output format and infer whether we are compressing or extracting.

irm-codebase:
@fgvieira I ended up adding the compress wrapper too... but as a separate subwrapper.

The reason is that combining both would bloat the wrapper, and make it confusing.
`libarchive` supports a lot of different formats, with combinations that are not always intuitive.
Combining everything into one wrapper will make maintenance a hassle due to too many conditionals, IMO.

Similarly, I do not think it's a good idea to "infer" stuff because there are simply too many combinations.
For example, tar files can be:
- pax + gzip -> *.tar.gz
- ustar + gzip -> *.tar.gz (again!)

I decided to force users to specify the format name as a parameter, always.

irm-codebase:
@fgvieira any modifications you'd like me to make?

utils/libarchive/compress/test/Snakefile
```diff
@@ -0,0 +1,33 @@
+rule compress_single_file_to_7zip:
+    input:
+        "a.txt",
+    output:
+        archive="results/compressed.7z",
```
fgvieira: No need to use a named output.
```suggestion
        "results/compressed.7z",
```

utils/libarchive/compress/test/Snakefile
```diff
@@ -0,0 +1,33 @@
...
+    log:
+        "logs/libarchive/compress/compress_single_file_to_7zip.log",
+    threads: 1
+    params:
+        format_name="7zip",
+    message:
+        "Compress single file into a 7zip archive."
```
fgvieira:
```suggestion
```

utils/libarchive/compress/test/Snakefile
```diff
@@ -0,0 +1,33 @@
...
+        archive="results/compressed.tar.gz",
+    log:
+        "logs/libarchive/compress/compress_multiple_files_to_tar_gz.log",
+    threads: 1
+    params:
+        internal_paths=["text/textfile_a.txt", "markdown/markdown_b.md"],
+        format_name="pax",
+        filter_name="gzip",
+    message:
+        "Compressing multiple files into a tar.gz archive."
```
fgvieira:
```suggestion
```

utils/libarchive/compress/test/Snakefile
```diff
@@ -0,0 +1,33 @@
...
+        "b.md",
+    output:
+        archive="results/compressed.tar.gz",
+    log:
+        "logs/libarchive/compress/compress_multiple_files_to_tar_gz.log",
+    threads: 1
+    params:
+        internal_paths=["text/textfile_a.txt", "markdown/markdown_b.md"],
+        format_name="pax",
+        filter_name="gzip",
```
fgvieira, line 27: Maybe simplify the names.
```suggestion
        format="pax",
        filter="gzip",
```

irm-codebase: I thought about this, but I went against it because of the following:
- `filter` is a python built-in function.
- These are the names used in `libarchive`. Keeping this match makes searching docs easier in case future fixes are needed.

irm-codebase:
Implemented the suggestions!
- `message` is no longer used in the tests.
- `archive` has been removed from both `compress` and `extract` (we now just fetch the first item in `output` / `input`, respectively).

I left `filter_name` and similar parameters. See my comments above.

utils/libarchive/compress/wrapper.py
```diff
@@ -0,0 +1,106 @@
...
+def _validate_internal_path(path: str) -> None:
+    """Reject archive paths that would be unsafe or ambiguous."""
+    parsed = PurePosixPath(path)
+    if (
+        not path
+        or path.endswith("/")
+        or parsed.is_absolute()
+        or any(part in {"", ".", ".."} for part in parsed.parts)
+    ):
+        raise ValueError(f"Internal path {path!r} is invalid.")
```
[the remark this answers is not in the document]
irm-codebase: I added a simpler version of this. The wrapper is going to be harsher on what paths it accepts, but that should be ok.

utils/libarchive/compress/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+params:
+  - format_name: >
+      Compression format (e.g., "7zip", "pax", "zip").
+      See the libarchive documentation for the full list.
+  - filter_name: |
+      Compression filter to use. Options vary per format.
+      Common combinations: 
+        * For zip and 7zip: none.
+        * For pax: gzip or xz (*.tar.gz, and *.tar.xz).
+  - extras: Dictionary with additional arguments for `libarchive.file_writer()`.
```
fgvieira: To be consistent with other wrappers, maybe make `params.extra` a string.

utils/libarchive/extract/meta.yaml
fgvieira, line 1: Are there no `params.extra` for `extract`?

utils/libarchive/compress/meta.yaml
irm-codebase: Done!

utils/libarchive/extract/meta.yaml
irm-codebase: Fair point! Added.

utils/libarchive/compress/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+params:
+  - format_name: >
+      Compression format (e.g., "7zip", "pax", "zip").
+      See the libarchive documentation for the full list.
+  - filter_name: |
+      Compression filter to use. Options vary per format.
+      Common combinations: 
+        * For zip and 7zip: none.
+        * For pax: gzip or xz (*.tar.gz, and *.tar.xz).
+  - extra: Dictionary with additional arguments for `libarchive.file_writer()`.
```
fgvieira: Are `internal_paths` also supported in `compress`? Are they mandatory? If not, are input paths used?
```suggestion
  - internal_paths: >
      Path of the output files within the archive.
      Order and length must match that of the output files.
  - extra: Dictionary with additional arguments for `libarchive.file_writer()`.
```

utils/libarchive/compress/test/Snakefile
```diff
@@ -0,0 +1,29 @@
...
+        "b.md",
+    output:
+        "results/compressed.tar.gz",
+    log:
+        "logs/libarchive/compress/compress_multiple_files_to_tar_gz.log",
+    threads: 1
+    params:
+        internal_paths=["text/textfile_a.txt", "markdown/markdown_b.md"],
+        format_name="pax",
+        filter_name="gzip",
```
fgvieira, line 27: Can you add an example for `params.extra`?

utils/libarchive/extract/test/Snakefile
```diff
@@ -0,0 +1,26 @@
...
+rule extract_multiple_files:
+    input:
+        "test.zip",
+    output:
+        ["results/zip/a.txt", "results/zip/b.txt"],
+    log:
+        "logs/libarchive/extract/extract_multiple_files.log",
+    threads: 1
+    params:
+        internal_paths=["textfile_a.txt", "subfolder/textfile_b.txt"],
```
fgvieira: Can you add an example for `params.extra`?

irm-codebase:
Added example for `extra`.
I also harmonised the `compress` / `extract` tests to reflect each other.

fgvieira:
Cool and thanks for your efforts!

One last thing; this wrapper calls a python library and so I understand it is easier to pass `params.extra` as a `dict`, but do you think it is possible (and makes sense) to have `params.extra` as a string rather than a `dict`?

It would just be, to be consistent with existing wrappers..

irm-codebase:
I'm a bit concerned that this could lead to either parsing issues in some cases (e.g., numeric values) and ultimately bloat these wrappers. `libarchive` is written in C, so it's very picky, and `python-libarchive-c` just wraps around it.
Most other wrappers I've seen with `extra` end up calling `shell`.

If there's functionality I could borrow from I'll do it. Otherwise, I'd prefer to avoid it if possible, to make the wrapper friendlier to data in `config.yaml`.

fgvieira (approved):

fgvieira:
Thanks again for your contribution!

irm-codebase:
Glad to help!

PR #5330, opened 2026-06-03
fix: improve sortmerna output file handling

Smeds:
The sortmerna wrapper decided how to write its output files based on
  whether the input reads were paired. That caused two problems:

  - Requesting only `aligned` produced no output, because the `--fastx`
    flag was only set when `other` was also requested.
  - Paired reads written to a single output file failed, because the
    wrapper always expected two separate files.

  Output handling now follows the following path: two output
  files give split reads, one output file gives interleaved reads.

### QC

## Summary by CodeRabbit

* **New Features**
  * Added interleaved paired-end output option alongside split paired outputs; output-format flags are now applied automatically based on requested outputs, and an index-directory can be supplied for reusing pre-built indexes.

* **Documentation**
  * Clarified inputs (reference FASTA, single vs paired query semantics), outputs (aligned/other/stats shapes and split vs interleaved behavior), params.extra meaning, and notes on automatic flags, temp-dir lifecycle, and idx_dir concurrency.

* **Tests**
  * Updated tests to exercise interleaved output and added index statistics snapshots.


bio/sortmerna/meta.yaml
```diff
@@ -4,12 +4,15 @@ url: https://sortmerna.readthedocs.io/
...
+  - other: Unaligned reads (single file, or a list of two files for split paired output)
+  - stats: Alignment statistics log (single file, optional)
 params:
-  - extra: aditional program arguments
+  - extra: additional program arguments
 notes: |
-  * The kvdb (key value database) will be created in a temporary directory. Then, you don't have to prune it before each run. If you want to re-use the index, you can specify the idx-dir parameter.
+  * `--fastx` is added automatically whenever `aligned` or `other` is requested.
+  * `--out2` is added automatically when an output is declared as a list of two files (split `_fwd`/`_rev` output); declaring single output files with paired input instead produces interleaved single files. `aligned` and `other`, when both requested, must use the same shape.
+  * The kvdb (key value database) is created in a temporary directory, so it does not need pruning before each run. To re-use a pre-built index, pass `--idx-dir <dir>` through `extra`.
```
fgvieira: Shouldn't `idx` be an `input:`?

bio/sortmerna/test/Snakefile
```diff
@@ -6,13 +6,34 @@ rule sortmerna_pe:
...
-    params:
-        extra="--idx-dir idx --paired_in --out2",
+    log:
+        "logs/sortmerna/reads_pe.log",
     threads: 16
     resources:
         mem_mb=3072,  # amount of memory for building the index
+    params:
+        # --out2 is injected automatically for list (split) outputs
+        extra="--idx-dir idx --paired_in",
```
fgvieira: Cant' this `--paired_in` parameter be inferred from the `input`?

bio/sortmerna/meta.yaml
Smeds: good suggestion! I moved it to input.

bio/sortmerna/test/Snakefile
Smeds: The paired_in (and paired_out) option is not related to the number of input or output files. Instead, it affects how paired-end reads are classified when one mate aligns to the reference and the other does not.

From the documentation:
...........................................................................................................................................................................................
The option paired_in is optimal for users who want all reads in the other.fasta file to be non-rRNA. However, there are small chances that reads which are non-rRNA will also be put into the aligned.fasta file.

The option paired_out is optimal for users who want only rRNA reads in the aligned.fasta file. However, there are small chances that reads which are rRNA will also be put into the other.fasta file.

In practice, paired_in causes the entire read pair to be classified as aligned if either mate aligns, whereas paired_out causes the entire pair to be classified as non-aligned if either mate is non-aligned.
...........................................................................................................................................................................................

fgvieira (approved):

PR #5341, opened 2026-06-11
feat: adds Porechop abi wrapper

arog-bioinfo:
Adds a standard wrapper for [Porechop_ABI](https://github.com/bonsai-team/Porechop_ABI), an ab initio adapter inference and trimming tool for Oxford Nanopore reads.

### QC

## Summary by CodeRabbit

* **New Features**
  * Added porechop_abi adapter-trimming with optional de novo adapter inference and consensus export.

* **Tests**
  * Added end-to-end test plus sample input and expected trimmed and consensus outputs to validate behavior.

* **Chores**
  * Added reproducible environment pinning for linux-64 (Python 3.12 and required runtime libraries) and updated environment configuration with explicit channels.


bio/porechop_abi/meta.yaml
```diff
@@ -0,0 +1,16 @@
...
+input:
+  - fq: one FASTA/FASTQ read file (can be gzipped)
+  - ab_initio_config: custom ab initio config file (optional)
+  - custom_adapters: custom adapters file (optional)
+output:
+  - fq: one trimmed FASTA/FASTQ read file
+  - consensus: consensus export file (only available with params.ab_initio=True)
+params:
+  - extra: additional program arguments
+  - ab_initio: "run adapter inference before trimming (bool, default: False)"
```
fgvieira, line 16: Do we need this param? Maybe we can just check if `input.ab_initio_config` is provided and, if so, run the ab_initio part.

arog-bioinfo: I think we should keep `params.ab_initio` separate from `input.ab_initio_config` because `porechop_abi` can run ab initio mode without a custom config file since the tool has a default config. So using the presence of `input.ab_initio_config` to decide whether to add `--ab_initio`  would make ab initio mode impossible unless a custom config is supplied. Additionally, I think some configuration can be also passed by command line options instead of the configuration file and would make that impossible too.

fgvieira (approved):

PR #5436, opened 2026-07-03
feat: add fastaguard wrapper

ehsanestaji:
## Summary

Adds a `bio/fastaguard` wrapper for [FastaGuard](https://github.com/ehsanestaji/FastaGuard), a FASTA preflight QC tool for assembly and submission-readiness workflows.

The wrapper uses FastaGuard 0.6 and writes HTML, JSON, TSV, and MultiQC custom-content JSON reports. QC verdicts remain in JSON and TSV, while process exit codes follow conventional command semantics.

### QC

* [x] I confirm that I have followed the documentation for contributing to `snakemake-wrappers`.
* [x] `test_wrappers.py` calls the example rules in `bio/fastaguard/test/Snakefile`.
* [x] `input:` and `output:` file paths in the rules can be chosen arbitrarily.
* [x] Command-line arguments are inferred from wrapper params where appropriate.
* [x] No temporary files are created by the wrapper.
* [x] `meta.yaml` links to the tool documentation under `url:`.
* [x] The Conda environment uses the recommended channel and package ordering.

### Local validation

- [x] `black --check bio/fastaguard/wrapper.py test_wrappers.py`
- [x] `snakefmt --check bio/fastaguard/test/Snakefile`
- [x] `snakemake --lint --snakefile bio/fastaguard/test/Snakefile`
- [x] `pytest -q test_wrappers.py::test_fastaguard`


bio/fastaguard/meta.yaml
```diff
@@ -0,0 +1,20 @@
...
+  fasta: Assembly FASTA file to validate before downstream QC.
+output:
+  html: Human-readable FastaGuard report.
+  json: Machine-readable FastaGuard report.
+  tsv: Tabular FastaGuard summary.
+  multiqc: MultiQC custom-content JSON.
+  exit_code: Captured FastaGuard process status for workflow routing.
+params:
+  profile: FastaGuard profile, normally assembly.
+  gate: FastaGuard gate mode, normally pipeline or submission.
```
fgvieira: Can't these params also be passed through `extra`?

bio/fastaguard/meta.yaml
```diff
@@ -0,0 +1,20 @@
+name: fastaguard
+description: FASTA preflight QC for assembly and submission-readiness workflows.
+url: https://github.com/ehsanestaji/FastaGuard
+authors:
+  - FastaGuard contributors
```
fgvieira: Here should go the author of the wrapper. 😄

bio/fastaguard/wrapper.py
```diff
@@ -0,0 +1,23 @@
...
+    "--out {snakemake.output.html} "
+    "--json {snakemake.output.json} "
+    "--tsv {snakemake.output.tsv} "
+    "--multiqc {snakemake.output.multiqc} "
+    "{extra} "
+    "{log}; "
+    'status="$?"; '
+    "set -e; "
+    'printf "%s\\n" "$status" > {snakemake.output.exit_code}; '
+    'if [ "$status" -eq 3 ] || [ "$status" -gt 3 ]; then exit "$status"; fi'
```
fgvieira:
```suggestion
    'if [ "$status" -ge 3 ]; then exit "$status"; fi'
```

fgvieira: Why only fail if exit status is greater or equal to three? What are those exit codes?

I understand your idea, but it is a bit unexpected to have a rule that keeps running even though the program returns a non-zero exit code. Do you think you could make it optional, so the user could define the exit status to fail? And default to 0, to be consistent with the other rules.

bio/fastaguard/wrapper.py
```diff
@@ -0,0 +1,23 @@
+from snakemake.shell import shell
+
+extra = snakemake.params.get("extra", "")
+allowed_exit_codes = snakemake.params.get("allowed_exit_codes", [0])
```
fgvieira:
```suggestion
allowed_exit_codes = snakemake.params.get("allowed_exit_codes", 0)
```

fgvieira:
I think it looks good but, since this exit code behavior is not standard, I just want to double-check.

@johanneskoester @dlaehnemann  what do you think about this wrappers handling of non-zero exit codes?

bio/fastaguard/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+input:
+  fasta: Assembly FASTA file to validate before downstream QC.
+output:
+  html: Human-readable FastaGuard report.
+  json: Machine-readable FastaGuard report.
+  tsv: Tabular FastaGuard summary.
+  multiqc: MultiQC custom-content JSON.
+  exit_code: Captured FastaGuard process status for workflow routing.
+params:
+  allowed_exit_codes: Exit codes treated as successful. Defaults to [0]; use [0, 1, 2] to preserve PASS, WARN, and QC-failure reports.
```
dlaehnemann: I would argue that `FastaGuard` itself should not use command line exit codes to signal these things. Because if the checked `fasta` file has issues, `FastaGuard` can still run successfully, providing the requested logging output. And having to hack around those exit codes, so that snakemake will preserve output files even for non-zero exit codes, is a symptom of this use of exit codes, which goes against the generally used convention for command line tools.

If you want workflows to fail hard if certain issues come up, this should be done downstream of `FastaGuard`, when parsing some of its outputs. This could be an initial check in expensive downstream rules, or a downstream rule dedicated to checking the `FastaGuard` results, which in turn returns some dummy file confirming passed QC.

As this is the current exit code logic of the tool, I guess we could include this wrapper as is. But I would strongly suggest to remove this logic from the tool itself and then update this wrapper. It would make it much simpler.

dlaehnemann (commented):
This sounds like a very useful tool with nice standardized outputs.

But this usage of exit codes goes against [the convention for shell commands, to indicate whether the command itself succeeded](https://en.wikipedia.org/wiki/Exit_status#Shell_and_scripts). Exit codes should not encode some quality control result.

So I would strongly suggest to change that logic in the tool itself. And then I can see two ways forward for this wrapper:

1. Hold back until a newly released version has a conventional exit code logic. Then create the much simpler wrapper from scratch (or by simplifying this pull request here drastically).
2. Merge this version of the wrapper with this version of the tool. Then update the wrapper accordingly with a newly released version that has a conventional exit code logic.

bio/fastaguard/meta.yaml
ehsanestaji: Thanks, this makes sense. Since the wrapper targets the released v0.5.0 contract, my preference is to keep the current exit-code handling for this PR, then simplify it in a follow-up after FastaGuard separates report generation from gate enforcement. Would that approach work for you?

dlaehnemann:
Regarding: https://github.com/snakemake/snakemake-wrappers/pull/5436#discussion_r3591137546

Yeah, of course. That's what I meant with option 2, so this is perfectly fine.

So @fgvieira, as you mostly reviewed this, do you want to merge this whenever you think it is ready?

fgvieira:
I feel the exit codes approach is a bit `hacky.
@ehsanestaji since you seem to be the main developer of [FastaGuard](https://github.com/ehsanestaji/FastaGuard), do you think you could make a small patch release to standardize the exit codes?

I'd say you could output a TSV summary file instead with the paths of the input files and the status:
```
file1    PASS
file2    WARN
file3    FAIL
```
this could be parsed downstream.

ehsanestaji:
Yes, that makes sense. FastaGuard already writes `verdict` and `gate_status` to TSV and JSON, with input paths included in compare output. I’ll prepare a small release that reserves non-zero exit codes for command or runtime errors, then update this wrapper to use the simplified behavior.

ehsanestaji:
Updated to FastaGuard 0.6.0 and removed the custom exit-code handling. QC verdicts remain in the JSON and TSV reports, while command failures now propagate normally. Local tests, formatting, and linting pass.

bio/fastaguard/wrapper.py
```diff
@@ -0,0 +1,14 @@
...
+
+extra = snakemake.params.get("extra", "")
+log = snakemake.log_fmt_shell(stdout=True, stderr=True)
+
+shell(
+    "fastaguard {snakemake.input.fasta} "
+    "--out {snakemake.output.html} "
+    "--json {snakemake.output.json} "
+    "--tsv {snakemake.output.tsv} "
+    "--multiqc {snakemake.output.multiqc} "
```
fgvieira, line 11: Are these outputs mandatory? If not, it could be nice to allow the user to select which outputs to generate.

ehsanestaji: At present, yes. FastaGuard 0.6.0 generates all four reports on every run, and those options only configure their paths. I kept them as explicit Snakemake outputs so every generated file remains tracked. Selective generation would require a tool-level option and a subsequent release. Would keeping the four outputs mandatory for this wrapper be acceptable?

fgvieira (approved):

PR #5539, opened 2026-07-10
feat: Add samtools import

fgvieira:
### QC

## Summary by CodeRabbit

- **New Features**
  - Added a Snakemake wrapper and tool metadata for `samtools import`, producing BAMs from FASTQ inputs with paired-end and interleaved options.
  - Added Conda environment definitions with pinned dependencies, including a platform-specific explicit environment lock.
- **Tests**
  - Updated the wrapper Snakemake workflow to use separate rules for interleaved, paired-end, and single-end imports.
  - Added an automated test to verify wrapper execution and expected BAM outputs (`a.bam`, `a.pe.bam`, `a.se.bam`).


bio/samtools/import/wrapper.py
```diff
@@ -0,0 +1,26 @@
...
+__copyright__ = "Copyright 2026, Johannes Köster"
+__license__ = "MIT"
+
+
+from snakemake.shell import shell
+from snakemake_wrapper_utils.samtools import get_samtools_opts
+
+samtools_opts = get_samtools_opts(snakemake)
+extra = snakemake.params.get("extra", "")
+log = snakemake.log_fmt_shell(stdout=True, stderr=True, append=True)
```
dlaehnemann: Are you sure you want to append here? If the wrapper gets executed for the same input files multiple times, you'll get a very long concatenated log file. But maybe I am missing something, and this is intentional?
```suggestion
log = snakemake.log_fmt_shell(stdout=True, stderr=True, append=False)
```

dlaehnemann (approved):
This looks very clean and straightforward, thanks!

Only one question about the `log` definition.
And maybe we could have tests for all three cases that the wrapper handles? So also a paired interleaved fastq and a single-end fastq, in addition? Shouldn't be too much hassle, right?

PR #5581, opened 2026-07-22
fix: swarm mem

fgvieira:
### QC

## Summary by CodeRabbit

* **New Features**
  * Automatically sets a memory-based `--ceiling {mem_mb}` for fastidious/fast swarm runs when no `-c/--ceiling` is provided, and only when bloom-bits options aren’t used.
* **Chores**
  * Refreshed pinned Conda environment specs and updated runtime/build pins; added `snakemake-wrapper-utils` pinned to version `0.9.0`.
* **Tests**
  * Updated swarm test rule options: `swarm_gz` now includes `--boundary 3`, and `swarm_bz2` removes `--fastidious` while keeping `--differences` and `--usearch-abundance`.


fgvieira:
Fails due to torognes/swarm#190. Will update to newest version when bioconda/bioconda-recipes#67435 is merged.

bio/swarm/wrapper.py
```diff
@@ -3,10 +3,24 @@
...
 
 from snakemake.shell import shell
+from snakemake_wrapper_utils.snakemake import get_mem, is_arg
 
 extra = snakemake.params.get("extra", "")
 log = snakemake.log_fmt_shell(stdout=True, stderr=True)
 
+# Define memory
+mem_mb = int(
+    get_mem(snakemake, "MiB", snakemake.params.get("mem_overhead_factor", 0.1))
```
dlaehnemann: I wouldn't think that a non-Java tool would need much of a memory overhead reserved. Or does your experience with `swarm` say otherwise?
```suggestion
    get_mem(snakemake, "MiB", snakemake.params.get("mem_overhead_factor", 0.01))
```

bio/swarm/wrapper.py
```diff
@@ -3,10 +3,24 @@
...
+)
+
+ceiling = ""
+if (
+    (is_arg("-f", extra) or is_arg("--fastidious", extra))
+    and not (is_arg("-y", extra) or is_arg("--bloom-bits", extra))
+    and not (is_arg("-c", extra) or is_arg("--ceiling", extra))
+):
+    ceiling = f"--ceiling {mem_mb}"
```
dlaehnemann, line 26: I would argue, that respecting `mem_mb` as specified for the workflow or the rule should not be conditional and should always happen. It should then in turn disallow the use of `--bloom-bits` or `--ceiling` in `params.extra`.

dlaehnemann: Ah, I see now. it should only be set explicitly, if the `--fastidious` option is set. Still, I would generally disallow the use of `--bloom-bits` and `--ceiling` via `params.extra` and also document the whole memory reservation behavior in the `meta.yaml`.

bio/swarm/test/Snakefile
```diff
@@ -33,7 +33,7 @@ rule swarm_gz:
         "logs/{sample}.gz.log",
     threads: 4
     params:
-        extra="--fastidious --differences 1 --usearch-abundance",
+        extra="--fastidious --boundary 3 --differences 1 --usearch-abundance",
```
dlaehnemann, line 44: I would add useful dynamic defaults of memory usage. I dug around a bit to find this one:
```suggestion
        extra="--fastidious --boundary 3 --differences 1 --usearch-abundance",
    resources:
        mem_mb=lambda wc, input: input.size_mb * 4 # with the fastidious option, the original swarm 2 paper suggests a factor of 4.75 * input_data (https://peerj.com/articles/1420/#p-22), a later pull request then suggests a 20% reduction in memory usage (https://github.com/torognes/swarm/issues/7#issuecomment-420290892)
```

bio/swarm/test/Snakefile
```diff
@@ -53,6 +53,6 @@ rule swarm_bz2:
         "logs/{sample}.bz2.log",
     threads: 4
     params:
-        extra="--fastidious --differences 1 --usearch-abundance",
+        extra="--differences 1 --usearch-abundance",
```
dlaehnemann, line 67:
```suggestion
        extra="--differences 1 --usearch-abundance",
    resources:
        mem_mb=lambda wc, input: input.size_mb * 0.7 # the swarm readme suggests a factor of 0.6 (https://github.com/torognes/swarm/blob/68e0dc095b2bf52ee054c806d72101ea9d2142a9/README.md?plain=1#L103), so this is a conservative memory reservation
```

dlaehnemann (commented):
As always, this generally looks good. Just some minor points:

1. I'd add useful suggestions of dynamic `mem_mb` specifications to the `Snakefile`.
2. I'd actively disallow the usage of `--ceiling` and `--bloom-bits` via `params: extra=`, and instead enforce the `--ceiling` whenever `--fastidious` is specified.
3. Document the memory reservation behavior in the `meta.yaml`.

bio/swarm/wrapper.py
fgvieira: Actually, no idea, but it is more because this memory setting is just for the Bloom filter part. So I assumed that some memory would still be needed for the rest of the program; but no idea how much.

bio/swarm/wrapper.py
fgvieira: Ok, so you suggest throwing an error if the user specifies them?

dlaehnemann: Yeah, I prefer being clear about this and letting the user clean up the arguments, instead of silently doing something with the specification to the best of our knowledge...

bio/swarm/wrapper.py
dlaehnemann: Ah, OK. I see the point. And yes, some overhead is probably a good idea. But seeing that the reservation for the Bloom filter will usually go into the GB and the overhead is probably more in the MB, a much smaller value probably makes sense, maybe `0.01`, which should set aside 10 MB (instead of 100 MB)  for every 1GB?

bio/swarm/meta.yaml
```diff
@@ -18,3 +18,4 @@ params:
   - extra: additional program arguments
 notes: |
   * Input files can be compressed with `gzip` or `bzip2`.
+  * Bloom filter size (if `--fastidious`) is controled solely by `--ceiling` (`--bloom-bits` option is not allowed in this wrapper) and infered from `resources.mem` and `params.mem_overhead_factor`.
```
dlaehnemann:
```suggestion
  * Bloom filter size (if `--fastidious`) is controlled solely by `--ceiling` (`--bloom-bits` option is not allowed in this wrapper) and inferred from `resources.mem` and `params.mem_overhead_factor`.
```

bio/swarm/wrapper.py
```diff
@@ -3,10 +3,28 @@
...
 from snakemake.shell import shell
+from snakemake_wrapper_utils.snakemake import get_mem, is_arg
 
 extra = snakemake.params.get("extra", "")
 log = snakemake.log_fmt_shell(stdout=True, stderr=True)
 
+# Check extra options
+assert not (
+    is_arg("-y", extra) or is_arg("--bloom-bits", extra)
+), "option '-y/--bloom-bits' not supported by wrapper"
```
dlaehnemann:
```suggestion
), "option '-y/--bloom-bits' not supported by wrapper, do not provide via 'params.extra'"
```

bio/swarm/wrapper.py
```diff
@@ -3,10 +3,28 @@
...
+assert not (
+    is_arg("-c", extra) or is_arg("--ceiling", extra)
+), "option '-c/--ceiling' is inferred from 'resources.mem'"
```
dlaehnemann:
```suggestion
), "option '-c/--ceiling' is inferred from 'resources.mem', do not provide via 'params.extra'"
```

dlaehnemann (approved):
This looks perfect, now. Just found two typos and have a slight addition suggestion in the new assertions. But do as you think is best with those.

